Issue #55, opened 2019-03-21
Deploy the extension to Chrome Web Store 🚀⚒


kmw101:
As a developer I want to deploy the chrome extension:

https://dev.to/gokatz/automate-your-chrome-extension-deployment-in-minutes-48gb

I need to link this to CI to automate the process.

albertandrejev:
@kmw101 who can create Google Chrome Web Store API keys and provide them to me like it was described at https://dev.to/gokatz/automate-your-chrome-extension-deployment-in-minutes-48gb and https://github.com/DrewML/chrome-webstore-upload/blob/master/How%20to%20generate%20Google%20API%20keys.md? In order to deploy extension to the WebStore I need this information.

PR #69, opened 2019-04-01
Setup browser extension exports

webmaster128:
First step for #55


webmaster128:
Updated and conflict resolved. Would be good to get this merged to get exporting tested and allow downloads of packages from Appveyor. Chrome Web Store will follow as an additional step.

Issue #20, opened 2019-02-12
Login Page


kmw101:
As a user I am asked to enter and my password. 

The continue button remains inactive until an email address (checked with a regular expression for a @ and a "." in the address).

If I enter an incorrect email or password I am shown the error toaster with a message "Sorry you have entered either an incorrect email address or password".

A correct email and password takes me to the account status page.

I have options to click on Restore Account (go to restore page https://projects.invisionapp.com/share/PRPRZCQ9MHY#/screens/341373581)
 and more options (Welcome page).

UX Guide: https://projects.invisionapp.com/share/PRPRZCQ9MHY#/screens/338674175

- [x] Create storybook with component view implementation (PR #71)
- [x] Toast component implementation (PR #77)
- [x] Add toast to Login view (PR #83)

albertandrejev:
But according to https://github.com/iov-one/ponferrada/issues/6 there is no need to check for email. Please correct me if I am wrong.

kmw101:
This is an old reference, email has been replaced by unique identifier that the user set up in the registration process.

albertandrejev:
- [ ] Create storybook with component view implementation
- [ ] Toast component implementation
- [ ] Logic to process login sequence

kmw101:
As as user if I enter the wrong password I am shown the error message along with my security hint.

webmaster128:
Closing since it is in the Done column

PR #71, opened 2019-04-02
Feature - Login page view

albertandrejev:
**Description**
This PR will add login view to the extension. It will close "Create storybook with component view implementation" part of the #20.

**Before**
No component was exists

**After**
Not ready yet


packages/sanes-chrome-extension/src/routes/login/index.tsx
```diff
@@ -0,0 +1,6 @@
+import * as React from 'react';
+import Layout from './components';
+
+const Login = (): JSX.Element => <Layout />;
```
apanizo: Remember now using only functional components, we need to balance the components I expect to have here, the validation function, onLogin function and the main layout.

```
function onValidate() {
    ...
}

function onLogin() {
   ...
}

const Login = () => {
    .... 
   return (
    <PageLayout id={LOGIN_ROUTE} primaryTitle="Log" title="In">
      <Form onSubmit={handleSubmit}>
         <LoginFormFields ... />
         </LoginFormControls ... />
     </Form>
   </PageLayout>

export default Login
```

packages/sanes-chrome-extension/src/routes/login/components/index.tsx
```diff
@@ -0,0 +1,106 @@
...
+        <Block marginTop={4} textAlign="center">
+          <Block marginBottom={1}>
+            <Link>
+              <Typography variant="subtitle2" color="primary" link inline>
+                Restore account
+              </Typography>
+            </Link>
+          </Block>
+          <Block>
+            <Link>
```
apanizo: If we use regular ```Link``` using just ```to={WELCOME_ROUTE}``` should be enough.

packages/sanes-chrome-extension/src/routes/login/index.dom.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    const LoginDom = await travelToLogin(store);
+
+    const inputs = TestUtils.scryRenderedDOMComponentsWithTag(
+      LoginDom,
+      'input'
+    );
+
+    expect(inputs.length).toBe(2);
+
+    TestUtils.findRenderedDOMComponentWithTag(LoginDom, 'button');
```
apanizo: why this line?

packages/sanes-chrome-extension/src/routes/login/index.dom.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    //Bottom links
+    const links = TestUtils.scryRenderedDOMComponentsWithTag(LoginDom, 'h6');
```
apanizo: `h6` is something really generic, so if you use it, you should do extra checks. Something like:
```
expect(links.length).toBe(3);
expect(links[0].innerHTML).toBe('whatever')
...
```

packages/sanes-chrome-extension/src/routes/welcome/index.dom.spec.ts
```diff
@@ -24,9 +24,25 @@ describe('DOM > Feature > Welcome', () => {
     await whenOnNavigatedToRoute(store, WELCOME_ROUTE);
   }, 55000);
 
-  it(`should create three buttons`, async () => {
+  it(`should create three buttons and redirect to appropriate route when clicking on it`, async () => {
     const buttons = TestUtils.scryRenderedDOMComponentsWithTag(dom, 'button');
 
     expect(buttons.length).toBe(3);
+
```
apanizo: 👌

packages/sanes-chrome-extension/src/routes/login/index.dom.spec.ts
albertandrejev: There should be only one button on this view.

albertandrejev: But I will update this line for better readability

packages/medulas-react-components/src/utils/storybook/index.tsx
```diff
@@ -1,10 +1,31 @@
...
+
+const store = createStore(createRootReducer(history));
+
 interface Props {
   readonly children: React.ReactNode;
 }
 
 export const Storybook = ({ children }: Props): JSX.Element => {
-  return <ThemeProvider injectFonts>{children}</ThemeProvider>;
+  return (
```
apanizo: The `medulas-react-components` package should not expose any redux or react router or history logic. Why do you need it?

albertandrejev: Because <Link> in storybook will not work otherwise.

albertandrejev:
`id={LOGIN_ROUTE}` from `login/index.tsx`

packages/sanes-chrome-extension/src/utils/test/LinkMock.tsx
```diff
@@ -0,0 +1,22 @@
...
+}
+
+const LinkMock = ({ to, children }: Props): JSX.Element => {
+  const onClick = (): boolean => {
+    history.push(to);
+    return false;
+  };
+
+  return (
+    <a href={to} onClick={onClick}>
```
apanizo: This is odd. Why in wallet-demo we did not need anything like this?

packages/sanes-chrome-extension/src/utils/storybook/index.tsx
```diff
@@ -0,0 +1,30 @@
+import { Storybook } from 'medulas-react-components/lib/utils/storybook';
+import { Provider } from 'react-redux';
+import { createMemoryHistory, History } from 'history';
+import { ConnectedRouter, connectRouter } from 'connected-react-router';
+import { createStore, combineReducers } from 'redux';
+import * as React from 'react';
+
+const history = createMemoryHistory();
```
apanizo: For avoiding confusion rename it to `storybookHistory`.

packages/sanes-chrome-extension/src/utils/storybook/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+//eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+const createRootReducer = (history: History) =>
```
apanizo: For avoiding confusion rename it to `storybookCreateRootReducer`

packages/sanes-chrome-extension/src/utils/storybook/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+  combineReducers({
+    router: connectRouter(history),
+  });
+
+const store = createStore(createRootReducer(history));
```
apanizo: `storybookStore`

packages/sanes-chrome-extension/src/routes/login/components/LoginForm.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  return (
+    <Form onSubmit={handleSubmit}>
+      <Block marginBottom={1}>
+        <TextFieldForm
+          placeholder="Enter your unique identifier"
+          form={form}
+          required
+          fullWidth
+          name={ACCOUNT_NAME_FIELD}
```
apanizo: Remove this Field for now. This field would imply to add extra logic for checking if an account exists after restoring the profile. And also send extra info to the account view status with the initial account to load.

We should leave that for the future. Let's cut the scope and move forward.

packages/sanes-chrome-extension/src/utils/test/LinkMock.tsx
albertandrejev: I do not know. But `Simulate.click` doesn't redirect to another route. Odd I know, but I can not understand why. Components looks same.

PR #75, opened 2019-04-03
Add a hack to shut down chain connections after query

ethanfrey:
After this `yarn test-unit src/logic` will work (when running all the chains from #72 and `export CHAINS_ENABLED=1`)

### DIFF
--- a/packages/sanes-chrome-extension/src/logic/index.ts
+++ b/packages/sanes-chrome-extension/src/logic/index.ts
@@ -22,9 +22,14 @@ const generateBlockchainAddressFrom = async (
   const nodes = chainSpec.bootstrapNodes;
   const codecType = codecFromString(chainSpec.codecType);
 
+  // this whole clause should be done elsewhere and cached...
+  // remember we need to close connections, so pass one in
   const connector: ChainConnector = chainConnector(codecType, nodes);
   const { connection } = await signer.addChain(connector);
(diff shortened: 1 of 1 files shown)


packages/sanes-chrome-extension/src/logic/index.ts
```diff
@@ -22,9 +22,14 @@ const generateBlockchainAddressFrom = async (
   const nodes = chainSpec.bootstrapNodes;
   const codecType = codecFromString(chainSpec.codecType);
 
+  // this whole clause should be done elsewhere and cached...
+  // remember we need to close connections, so pass one in
   const connector: ChainConnector = chainConnector(codecType, nodes);
   const { connection } = await signer.addChain(connector);
   const chainId = connection.chainId();
+  signer.shutdown();
```
webmaster128: The lifecycle of `signer` should be symmetric, i.e. `signer` should be created at the same abraction level as `signer.shutdown()` is called.

The whole `generateBlockchainAddressFrom` signature no sense to me. When you call it with different derivation, the same chains are connected multiple times.

apanizo: This method is removed in our version (the version we coded together on Friday).

Let's do one thing. I will add the `wontfix` tag, and close it. And we'll keep in mind the shutdown stuff.

PR #76, opened 2019-04-03
Create ProfileWithAccount encapsulation

ethanfrey:
This is an approach to handling all account info.

We can create a "full config" with a lot more logic information added to the config file, so we can pass that around (one place to handle this stuff), rather than call these switch statements from many places.

Create ProfileWithAccounts that wraps a profile with all logic for loading and creating accounts. This will check what already exists, can be used to extend existing derivations with new chains for example. Allows to list all created derivations. Also create new ones at will.

(This required a hack around iov-core API and doesn't store the name properly, but I think it is 80% there... really needs unit tests, but it is midnight already)

Take a look at this and let's discuss any questions. I think this will give you all the operators you will need to work with any account. Then time for cleanup....


packages/sanes-chrome-extension/src/logic/account/index.ts
```diff
@@ -1,24 +1,247 @@
...
+    walletId: WalletId,
+    chainId: ChainId,
+    path: ReadonlyArray<Slip10RawIndex>
+  ): Promise<PublicIdentity | undefined> {
+    // Breaking encapsulation... bad Ethan
+    const wallet: ReadonlyWallet = (this
+      .profile as any).findWalletInPrimaryKeyring(walletId); // eslint-disable-line @typescript-eslint/no-explicit-any
+    const ident = await wallet.previewIdentity(chainId, path);
+    const allIdentities = wallet.getIdentities();
+    return allIdentities.find(x => publicIdentityEquals(x, ident));
```
webmaster128: To replace this whole function, we would need a

```
  /**
   * Checks the existence of an identitiy in the wallet with the given ID in the primary keyring
   */
  public async identityExists(
    walletId: WalletId,
    chainId: ChainId,
    options: Ed25519Keypair | ReadonlyArray<Slip10RawIndex> | number,
  ): Promise<bool> {
    // ...
  }
```

i.e. the same interface as `UserProfile.createIdentity` but as an existence check, right?

packages/sanes-chrome-extension/src/logic/blockchain/wallet.ts
```diff
@@ -4,6 +4,36 @@ import { ValueAndUpdates } from '@iov/stream';
 import { WalletInfo, HdPaths } from '@iov/keycontrol';
 import { Slip10RawIndex } from '@iov/crypto';
 
+export type Algorithm = 'secp256k1' | 'ed25519';
```
webmaster128: This will soon bring a collision of the symbol `Algorithm`. Why not reuse the `Algorithm` enum from @iov/bcp?

PR #77, opened 2019-04-05
Feature - Toast component

albertandrejev:
**Description**
This PR will add Toast component. Part of the #20 - "Toast component implementation"


packages/medulas-react-components/src/theme/assets/toast/close.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" fill="none" viewBox="0 0 20 20">
```
apanizo: If these assets are only going to be used with Toast I prefer to move them to `src/context/ToastProvider/assets`.

packages/medulas-react-components/src/theme/utils/mui.ts
```diff
@@ -142,6 +142,29 @@ const themeObject: ThemeOptions = {
         fontSize: '1.4rem',
       },
     },
+    MuiSnackbarContent: {
+      root: {
+        [theme.breakpoints.up('xs')]: {
+          borderRadius: 2,
+          boxShadow: '0 0 6px 0 #f3f4fb',
+        },
+        width: 460,
```
apanizo: I believe the chrome extension only has... 350px as width, is this a problem? can you paste a couple of screenshots in the PR's description?

packages/medulas-react-components/src/context/ToastProvider/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+    });
+
+  const closeToast = (): void =>
+    setToast({
+      message: '',
+      open: false,
+      variant: ToastVariant.INFO,
+    });
+
+  const toastConfig = {
```
apanizo: The only information which needs to be passed down to tree is the `show` and `close` functionality, in this case, when a component does:
```
const { show, close} = useContext(ToastContext);
```
why do we need to expose the rest of messages?

packages/medulas-react-components/src/theme/assets/toast/close.svg
albertandrejev: But in this case you will need to change build script to copy those assets to the `lib`

apanizo: oh, you are right. Keep it as you did in first place.

packages/medulas-react-components/src/context/ToastProvider/index.tsx
albertandrejev: Yes, but SharedToast component (ToastProvider/index.tsx  line: 11) required some of this information.

PR #78, opened 2019-04-07
Feature - Refactor usage of react scripts

apanizo:
**Description**
This PR updates the packages for using a custom published version of react-scripts containing all logic for testing building and developing a chrome extension.

Basically we have reduced our technical debt because we do not use third party library `cra-build-watch` pointing to a commit in a PR done by me for supporting extension' scripts, and also, we do not need to maintain our package `oza-react-scripts` for allowing to build the compiled files of the extension.

Now everything is wrapped and published as a react-scripts version which can be used by anyone.

See:
- https://github.com/apanizo/create-react-app/pull/1
- https://www.npmjs.com/package/chrome-extension-react-scripts


packages/sanes-chrome-extension/package.json
```diff
@@ -2,22 +2,26 @@
...
+  "config": {
+    "cs": "src/extension/contentscript.ts",
+    "bs": "src/extension/backgroundscript.ts"
+  },
   "scripts": {
     "use-config-test": "cp ./config/test.json ./public/assets/config/conf.json",
     "use-config-development": "cp ./config/development.json ./public/assets/config/conf.json",
     "use-config-production": "cp ./config/production.json ./public/assets/config/conf.json",
-    "build": "oza-react-scripts build",
+    "build": "react-scripts build $npm_package_config_cs $npm_package_config_bs",
```
webmaster128: Looks like environment variables are not supported in Windows builds. `shx` does not support environment variables. Can we simplify this by inlining the file paths? Or do we need the ability to override the value?

Issue #66, opened 2019-03-31
Create web app example


apanizo:
For interacting with the extension we need an example web app.

This project should live inside ponferrada monorepo, and should be called `bierzo-wallet`.

Steps:
1. Create repo using create-react-app typescript version inside packages
2. Load medulas-react-components using `lerna add` medulas dependency
3. Find an inspiration BASIC webpage with BASIC functionality:
    - Requesting access to website
    - Send tx o IOV address

Keep in mind the web app should manage negativa answers from the extension (the extension rejects your petition, the extension fails when sending the tx)

kmw101:
Looking at the Demo Wallet, there is only the send transaction that needs signing. Please use the Demo Wallet as the web application to interact with the browser.

There are a number of flows in the [document ](https://docs.google.com/presentation/d/1a8QjPozeix_XUiufs2CCYWfYvdRSLH0G2PGbKOxFz8w/edit#slide=id.p)which show how the send token works with the signer.

kmw101:
Following a Slack conversion this is not a requirement.

webmaster128:
Closing as this is done

PR #79, opened 2019-04-08
Feature #66 - Create Web App Example

abefernan:
**Description**
This PR adds the initial commit for bierzo-wallet.

I have used the current implementation as a way to get familiar with React components & hooks, and material-ui. The following steps are to implement styles with classes and style the components.

**Screenshots**
The following screenshots showcase the implemented view and its boxing model:

imagen

imagen


packages/bierzo-wallet/README.md
```diff
@@ -0,0 +1,44 @@
+This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
```
apanizo: Change this and explain this project will be for demonstrating the basic interaction with the chrome extension.

packages/bierzo-wallet/src/components/wallet/Wallet.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
```
apanizo: this can be renamed to `index.tsx`. All default/main components should be named in this way.

packages/bierzo-wallet/src/components/wallet/Wallet.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const Wallet = () => {
+  const styleGrid: React.CSSProperties = {
+    backgroundColor: '#f9f9f9',
+    height: 'auto',
+    minHeight: '100vh',
+    display: 'grid',
+    gridTemplateColumns: '1fr minmax(375px, 450px) 1fr',
+    gridTemplateRows: 'repeat(4, auto)',
+    gridTemplateAreas: `
```
apanizo: I am curious to see this in action. Why are you following this approach? Can you update the PR and include an a couple of screenshots?

packages/bierzo-wallet/src/components/dummy/DummyRoot.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
```
apanizo: If those components are not used, please, remove them.

packages/bierzo-wallet/src/components/dummy/EditableLabel.tsx
```diff
@@ -0,0 +1,16 @@
+import TextField from '@material-ui/core/TextField';
```
apanizo: Same here.

packages/bierzo-wallet/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "react-scripts": "2.1.8",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6",
+    "typescript": "3.4.1"
+  },
+  "scripts": {
+    "start": "react-scripts start",
+    "build": "react-scripts build",
+    "test": "react-scripts test",
+    "eject": "react-scripts eject"
```
apanizo: The eject script should be removed. Never eject. Never!

packages/bierzo-wallet/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "dependencies": {
+    "@material-ui/core": "^3.9.3",
+    "@types/jest": "24.0.11",
+    "@types/node": "11.12.2",
+    "@types/react": "16.8.10",
+    "@types/react-dom": "16.8.3",
+    "medulas-react-components": "^0.0.0",
+    "normalize.css": "^8.0.1",
+    "react-scripts": "2.1.8",
+    "react": "^16.8.6",
```
apanizo: 💥 General libraries should be moved to root's lerna monorepo. At least the extension is using react, so please move up.

packages/bierzo-wallet/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "private": true,
+  "dependencies": {
+    "@material-ui/core": "^3.9.3",
+    "@types/jest": "24.0.11",
+    "@types/node": "11.12.2",
+    "@types/react": "16.8.10",
+    "@types/react-dom": "16.8.3",
+    "medulas-react-components": "^0.0.0",
+    "normalize.css": "^8.0.1",
+    "react-scripts": "2.1.8",
```
apanizo: Buuuut! react-scripts should be kept here! In the end, we are using another version in other packages of the same roots. So do not move it.

packages/bierzo-wallet/src/components/Wallet/ContinueButton/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+    },
+  },
+  label: {
+    '&&': {
+      textTransform: 'capitalize',
+    },
+  },
+})(Button);
+
+const ContinueButton = (props: { positioning: React.CSSProperties }) => {
```
apanizo: Using Typescript you should declare types in this way:
```
interface Props {
  readonly positioning: React.CSSProperties;
}

....

const continueButton = ({ positioning }: Props) => {
    ....
```

packages/bierzo-wallet/src/components/Wallet/CurrencyToSend/index.tsx
```diff
@@ -0,0 +1,65 @@
+import Avatar from '@material-ui/core/Avatar';
+import Paper from '@material-ui/core/Paper';
+import Select from '@material-ui/core/Select';
+import TextField from '@material-ui/core/TextField';
+import { withStyles } from '@material-ui/styles';
+import React from 'react';
+
+const StyledPaper = withStyles({
```
apanizo: ```withStyles``` usage AFAIK, is discouraged. If you want to use styled 💅 components, you should go with:
https://next.material-ui.com/css-in-js/basics/#styled-components-api

But I recommend the hook way, which is the one we are using in the extension:
https://next.material-ui.com/css-in-js/basics/#hook-api

packages/bierzo-wallet/src/components/Wallet/ContinueButton/index.tsx
```diff
@@ -0,0 +1,26 @@
+import { Button, withStyles } from '@material-ui/core';
+import React from 'react';
+
+const StyledButton = withStyles({
+  root: {
+    '&&': {
```
apanizo: What it does the double ampersand?

abefernan: A single ampersand is replaced by the compiled parent selector. The double ampersand is used to increase the specificity of the rules by having the compiled parent selector two times.

I found that without it I couldn't override the default rules of Material UI. For instance, I couldn't make the avatar's default color yellow.

packages/bierzo-wallet/src/components/Wallet/CurrencyToSend/index.tsx
abefernan: I tried both approaches but I couldn't find a way to override the inner classes. That's why I chose the "Shorthand" approach that can be found here:
[https://material-ui.com/customization/overrides/](https://material-ui.com/customization/overrides/)

packages/bierzo-wallet/src/components/wallet/Wallet.tsx
abefernan: Fixed by [14f6201](https://github.com/iov-one/ponferrada/pull/79/commits/14f6201854bd623d3fb74608509f7e5b7c587dc8)

packages/bierzo-wallet/src/components/dummy/DummyRoot.tsx
abefernan: Fixed by [14f6201](https://github.com/iov-one/ponferrada/pull/79/commits/14f6201854bd623d3fb74608509f7e5b7c587dc8)

packages/bierzo-wallet/src/components/dummy/EditableLabel.tsx
abefernan: Fixed by [14f6201](https://github.com/iov-one/ponferrada/pull/79/commits/14f6201854bd623d3fb74608509f7e5b7c587dc8)

packages/bierzo-wallet/package.json
abefernan: Fixed by [14f6201](https://github.com/iov-one/ponferrada/pull/79/commits/14f6201854bd623d3fb74608509f7e5b7c587dc8)

packages/bierzo-wallet/package.json
abefernan: Fixed by [14f6201](https://github.com/iov-one/ponferrada/pull/79/commits/14f6201854bd623d3fb74608509f7e5b7c587dc8)

packages/bierzo-wallet/package.json
abefernan: Ok, react-scripts was already in place then.

packages/bierzo-wallet/src/components/Wallet/ReceiverAddress/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+      height: '50px',
+      margin: '16px 0 8px 0',
+    },
+  }),
+};
+
+const ReceiverAddress = ({ positionedClass }: Props) => {
+  const classes = useStyles();
+
+  const overrideClasses = {
```
apanizo: This approach is absolutely different about what we are doing, also I have not seen it anywhere. I recommend to simplify it and take advantage of variable names:
```
const useStyles = makeStyles({
  foo: {
     ...
  },
  bar: {
     ...
  }
});

const ReceiverAddress = (...) => {
    const classes = useStyles();
    const paperClasses = {
       root: classes.foo,
   }
   const textFieldClasses = {
       root: classes.bar
   }

  ....
```

apanizo: Please, update all components and follow this approach. In this way, with the variable name,  you can get a direct hint about what is going on `textFieldClasses --> classes for textField` 😉.

apanizo:
@abefernan also disable the `test` in package.json for making Travis to work. Once you add tests you can enable it back again.

packages/bierzo-wallet/README.md
abefernan: Fixed by [14f6201](https://github.com/iov-one/ponferrada/pull/79/commits/14f6201854bd623d3fb74608509f7e5b7c587dc8)

packages/bierzo-wallet/src/components/wallet/Wallet.tsx
abefernan: Updated the PR with screenshots.

packages/bierzo-wallet/src/components/Wallet/ContinueButton/index.tsx
abefernan: Fixed by [3daf67f](https://github.com/iov-one/ponferrada/pull/79/commits/3daf67fea9627053c6c04109b97b5e0b2e8a2b46).

packages/bierzo-wallet/src/components/Wallet/CurrencyToSend/index.tsx
abefernan: Fixed by [3daf67f](https://github.com/iov-one/ponferrada/pull/79/commits/3daf67fea9627053c6c04109b97b5e0b2e8a2b46).

packages/bierzo-wallet/src/components/Wallet/ContinueButton/index.tsx
abefernan: Removed the double ampersand in [c294569](https://github.com/iov-one/ponferrada/pull/79/commits/c294569bcd0b16f865ce6527ed5af11185a3bcd5).

packages/bierzo-wallet/src/components/Wallet/ReceiverAddress/index.tsx
abefernan: Fixed in [7b42812](https://github.com/iov-one/ponferrada/pull/79/commits/7b4281279aca898a48cac504a7f98a8e7cbbf32b).

abefernan:
> @abefernan also disable the `test` in package.json for making Travis to work. Once you add tests you can enable it back again.

Fixed in [07c6ad0](https://github.com/iov-one/ponferrada/pull/79/commits/07c6ad040225cdc92ebc5b73e77d38163ba68437).

abefernan:
> Can you include a screenshot in the PR's description with the most important views?

Updated the PR with screenshots.

PR #82, opened 2019-04-09
Feature - Generate pdf with recovery phrase

### LINKED ISSUE #10: Create "show backup phrase" page
- [x] Create storybook with this screen (#81)
- [x] Add logic to show backup phrase of the current user (#81)
- [x] Add pdf generation and download (#82)

albertandrejev:
This PR will ability to generate PDF containing security phrase on client side. It will implement " Add pdf generation and download" part of the issue #10.

It will close #10.

**Before**
recovery-phrase

**After**
recovery-phrase2

**Developer notes**
In order to generate PDFs [jsPDF](https://github.com/MrRio/jsPDF) package was used


packages/medulas-react-components/src/components/Button/Download.tsx
```diff
@@ -0,0 +1,71 @@
...
+    },
+  },
+  sizeSmall: {
+    height: theme.spacing(4),
+  },
+  secondary: {
+    backgroundColor: theme.palette.background.default,
+    padding: 0,
+    boxShadow: 'none',
+    border: `1px solid #f3f3f3`,
```
apanizo: Why not `grey[300]` which is our default border?

packages/medulas-react-components/src/components/Button/Download.tsx
```diff
@@ -0,0 +1,71 @@
...
+      color="secondary"
+      aria-label="Export as CSV"
+      classes={fabClasses}
+      onClick={onDownload}
+    >
+      <CircleImage
+        icon={download}
+        circleColor={theme.palette.primary.main}
+        alt="Download"
+        dia={theme.spacing(4)}
```
apanizo: Instead of importing it, you can use `const theme = useTheme();`

albertandrejev:
No, those fonts should be used in PDF generator. Not from ThemeProvider.

package.json
```diff
@@ -41,6 +41,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.3.4",
+    "@emotion/core": "^10.0.10",
```
apanizo: What is this?

packages/medulas-react-components/src/components/Image/index.stories.tsx
```diff
@@ -1,14 +1,35 @@
 import { storiesOf } from '@storybook/react';
 import React from 'react';
 import Image from './index';
+import CircleImage from './CircleImage';
 import { Storybook } from '../../utils/storybook';
 import iovLogo from './assets/iov-logo.png';
+import download from '../../theme/assets/download.svg';
+import Grid from '../Grid';
+import GridItem from '../GridItem';
+import theme from '../../theme/utils/mui';
```
apanizo: Can we use `useTheme` here as well?

packages/sanes-chrome-extension/src/routes/signup/components/ShowPhraseForm.tsx
```diff
@@ -11,7 +11,6 @@ import { getGlobalPersona } from '../../../logic/persona';
 
 export const SECOND_STEP_SIGNUP_ROUTE = `${SIGNUP_ROUTE}2`;
 
-// Exported only for being used in test environments
```
apanizo: Do not remove this comment, right now is an escape solution for getting the mnemonic, but this will change soon. Info is going to be injected via the context API. For having a better visual alert, change it to:

// FIXME Refactor it once we inject data using Context API

package.json
albertandrejev: I do not know why it stop working without it but Storybook requires it to run. This is odd, this happens after branch was rebased with master.

albertandrejev: Fixed in latest alpha (https://github.com/storybooks/storybook/issues/5817) so I will upgrade it.

packages/medulas-react-components/src/components/Image/index.stories.tsx
albertandrejev: Not possible because of different versions of react (in storybook itself: 16.8.6 and system wide: 16.9.0-alpha.0) and this will it possible to use hooks (well known "Invalid hook call. Hooks can only be called inside of the body of a function component." issue)

PR #84, opened 2019-04-10
Refactor config, Persona; Add signer; Implement balances

webmaster128:
- Update TypeScript to 3.4 because some of our code required advances type inferance features
- Update IOV-Core for no reason
- Remove all generics in configuration and split into configuration file and runtime configuration
- Run all tests when running `yarn test`
- Remove all usage of database logic for now
- Extract `Persona` unit tests with no network
- Add and test `Persona.mnemonic`
- Add and test `Persona.addChain`

and similar stuff


packages/sanes-chrome-extension/src/logic/persona/persona.ts
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  public mnemonic(): string {
+    const firstWalletId = this._userProfile.wallets.value[0].id;
+    // all wallets use the same mnemonic, so we only need to check the first one
+    const mnemonic = this._userProfile.printableSecret(firstWalletId);
+    return mnemonic;
+  }
+
+  public async addChain(newChain: PersonaChainConfig): Promise<void> {
```
apanizo: This function is not used and we recreate everything when loading the persona. I suggest to remove it.

packages/sanes-chrome-extension/src/logic/network/personafromconfig.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  const persona = new Persona(profile, chains);
+  const derivation = 0;
+  await persona.generateAccount(derivation);
+
+  return persona;
+};
+
+/**
+ * Creates Persona if not yet created.
```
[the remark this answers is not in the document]
webmaster128: You are right! This is supposed to be a singleton but I copied from the wrong place.

packages/sanes-chrome-extension/src/logic/network/signerandprofile.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  return {
+    signer: signer,
+    profile: profile,
+  };
+}
+
+/** Creates MultiChainSigner and connects all chains if not yet created */
+export function getSignerAndProfile(): ReturnType<
```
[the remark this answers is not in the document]
webmaster128: well spotted, is supposed to be a singleton

packages/sanes-chrome-extension/src/logic/network/personafromconfig.ts
```diff
@@ -0,0 +1,26 @@
+import { getRuntimeConfiguration } from './runtimeconfiguration';
+import { Persona } from '../persona';
+import { getSignerAndProfile } from './signerandprofile';
+import { singleton } from '../../utils/singleton';
+
+const createPersonaFromConfig = async (): Promise<Persona> => {
```
apanizo: Unless I am wrong this is more or less what I did with PersonaFactory. This should be a kind of standard method to recreate persona each time the extension is open for:
A. Need to unlock it if the browser has just opened.
B. Just load it into memory for using it in the app (for example first usage should check it for navigating to  correct route).

Do not use it as a  singleton, using this approach, we need to have features to be able to clean it during tests sessions. Probably making Persona static, having a clear static method.

packages/sanes-chrome-extension/src/logic/network/personafromconfig.ts
```diff
@@ -0,0 +1,26 @@
...
+  return persona;
+};
+
+/**
+ * Creates Persona if not yet created.
+ *
+ * Uses the getSignerAndProfile() singleton internally, such that
+ * getSignerAndProfile() and getPersonaFromConfig() always use the same profile.
+ */
+export const getPersonaFromConfig = singleton<typeof createPersonaFromConfig>(
```
apanizo: I believe you can remove it.

packages/sanes-chrome-extension/src/logic/network/runtimeconfiguration.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+async function loadRuntimeConfiguration(): Promise<RuntimeConfiguration> {
+  const config = await getConfigurationFile();
+  const runtimeChains = await Promise.all(config.chains.map(loadChainSpec));
+  return { chains: runtimeChains };
+}
+
+/** Requires a network connection in order to get chain IDs */
+export const getRuntimeConfiguration = singleton<
```
apanizo: What is the point of having it singleton? Now persona should be recreated each time the extension is open, once the persona is recreated there is no need to have access to config, right? You can remove it.

packages/sanes-chrome-extension/src/logic/network/signerandprofile.ts
```diff
@@ -0,0 +1,33 @@
...
+import { chainConnector } from './runtimeconfiguration';
+import { getConfigurationFile, codecTypeFromString } from '../config';
+import { singleton } from '../../utils/singleton';
+
+export interface SignerAndProfile {
+  readonly signer: MultiChainSigner;
+  readonly profile: UserProfile;
+}
+
+async function createSignerAndProfile(): Promise<SignerAndProfile> {
```
apanizo: One class returning two things, well... I also do not like it. I believe a more correct way to do it, as I suggested before, is moving the signer creation into Persona constructor. We break less "rules"

packages/sanes-chrome-extension/src/logic/network/balances.ts
```diff
@@ -0,0 +1,28 @@
...
+  const identities = [...account.identities.values()];
+  const pendingAccountResults = identities.map(identity => {
+    const { chainId, pubkey } = identity;
+    return singletonSigner.connection(chainId).getAccount({ pubkey });
+  });
+  const accountResults = await Promise.all(pendingAccountResults);
+
+  const out: Amount[] = [];
+  for (const result of accountResults) {
+    if (result) {
```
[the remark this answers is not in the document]
webmaster128: Good idea, I just tried that. Unfortunatly after the filtering I still get the result type `Account | undefined`, but I need `Account` to access elements. So this version is better for type safety.

packages/sanes-chrome-extension/src/logic/network/balances.ts
```diff
@@ -0,0 +1,28 @@
+import { Amount } from '@iov/bcp';
+
+import { getPersonaFromConfig } from './personafromconfig';
+import { getSignerAndProfile } from './signerandprofile';
+
+export async function getBalances(
+  accountIndex: number
+): Promise<ReadonlyArray<Amount>> {
+  const singletonPersona = await getPersonaFromConfig();
```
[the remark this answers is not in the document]
webmaster128: updated this as discussed. Now there is `Persona.getBalances`

webmaster128:
I'll address the remaining comments together with some additional documentation tomorrow or on Monday.

packages/sanes-chrome-extension/src/logic/network/runtimeconfiguration.ts
webmaster128: I was able to refactor away the entire network connection in the configuration part. Now everthing happens in Persona.create(). So this is done 👍

packages/sanes-chrome-extension/src/logic/network/signerandprofile.ts
webmaster128: Now everthing is created together in Persona.create(), so createSignerAndProfile was removed

packages/sanes-chrome-extension/src/logic/network/personafromconfig.ts
webmaster128: A new Persona can now be created using Persona.create().

Now there is a `getGlobalPersona()` singleton as an intermediate step to bring this persona into the UI. If with #85 we can make sure that only one instance is created, great. Then `getGlobalPersona` can be removed.

packages/sanes-chrome-extension/src/logic/persona/persona.ts
```diff
@@ -0,0 +1,96 @@
...
+
+export class Persona {
+  /**
+   * Creates a new Persona instance.
+   *
+   * This function does everything that cannot be done in a constructor
+   * (because a constructor is synchonous): reading configs, connecting to the network,
+   * creating accounts.
+   */
+  public static async create(): Promise<Persona> {
```
apanizo: When I mentioned the Factory, the idea is having this constructor inside it. I like to have the factory because:
1. Is a standard way program against interfaces (no need to use new), and it is a pretty common way of solving this.
2. It is easier to put the logic of having multiple personas in the future.
3. You can put the logic of cleaning memory easily and solve one of the test problems at once.

```
class PersonaFactory {
  private static persona; // in the future personas: Map<id, Persona>

 public static create(): Persona { 
        .... 
       this.persona = persona;

      return this.persona;
  }

  public static get() {
     return this.persona
  }

  public static clear() { 
    this.persona = undefined;
  }

Comment (apanizo) [reply packages/sanes-chrome-extension/src/logic/persona/persona.ts]: Also, the fact of one entity know how to create itself seems to be odd, what is your reasoning behind this? Maybe I am missing something.

File: packages/sanes-chrome-extension/src/logic/persona/persona.ts [hunk 11]
```diff
@@ -0,0 +1,96 @@
...
+    accountIndex: number
+  ): Promise<ReadonlyArray<Amount>> {
+    const account = (await this.accountManager.accounts())[accountIndex];
+    const identities = [...account.identities.values()];
+    const pendingAccountResults = identities.map(identity => {
+      const { chainId, pubkey } = identity;
+      return this.signer.connection(chainId).getAccount({ pubkey });
+    });
+    const accountResults = await Promise.all(pendingAccountResults);
+
```
apanizo: Interesting, are you saying this:
```
const positiveResults = accountResults.filter(acc => !!acc);
```
does not work?

or even:
```
const balances = Array.from(accountResults.filter(acc => !!acc).map(acc => ...acc.balance));
```

packages/sanes-chrome-extension/src/logic/persona/persona.ts
```diff
@@ -0,0 +1,96 @@
...
+  public async getAccounts(): Promise<ReadonlyArray<AccountInfo>> {
+    return this.accountManager.accounts();
+  }
+
+  public async getBalances(
+    accountIndex: number
+  ): Promise<ReadonlyArray<Amount>> {
+    const account = (await this.accountManager.accounts())[accountIndex];
+    const identities = [...account.identities.values()];
```
apanizo: I like it.

packages/sanes-chrome-extension/src/logic/persona/persona.ts
```diff
@@ -0,0 +1,96 @@
...
+    const manager = new AccountManager(profile, managerChains);
+
+    // Setup accounts
+    await manager.generateAccount(0);
+
+    return new Persona(signer, manager);
+  }
+
+  private readonly signer: MultiChainSigner;
+  private readonly accountManager: AccountManager;
```
apanizo: Interesting, I will see what is going on inside AccountManager.

packages/sanes-chrome-extension/src/logic/persona/persona.ts
```diff
@@ -0,0 +1,96 @@
...
+    for (const result of accountResults) {
+      if (result) {
+        out.push(...result.balance);
+      }
+    }
+    return out;
+  }
+
+  public destroy(): void {
+    this.signer.shutdown();
```
apanizo: Why finally we need to shut it down?

packages/sanes-chrome-extension/src/logic/persona/personafactory.ts
```diff
@@ -0,0 +1,7 @@
+import { Persona } from './persona';
+import { singleton } from '../../utils/singleton';
+
+// TODO: Remove this when https://github.com/iov-one/ponferrada/pull/85 is done
```
apanizo: The PR 85 is blocked until I have something like:
```
const persona = await PersonaFactory.recreatePersona();
if (!persona) {
   history.push(WELCOME_VIEW);
}

if (persona.locked())  // I have seen there is a method on iov-core for knowing if the profile is locked.
   history.push(LOGIN_VIEW);

history.push(ACCOUNT_STATUS_VIEW);
```

tsconfig.json
```diff
@@ -8,8 +8,8 @@
     "noImplicitAny": true,
     "noImplicitReturns": true,
     "noImplicitThis": true,
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
+    "noUnusedLocals": false,
```
apanizo: :(

tsconfig.json
```diff
@@ -8,8 +8,8 @@
...
+    "noUnusedParameters": false,
```
apanizo: :(

packages/sanes-chrome-extension/src/logic/user/accountmanager.test.ts
```diff
@@ -0,0 +1,148 @@
+import { Algorithm, ChainId } from '@iov/bcp';
+import { HdPaths } from '@iov/core';
+
+import { AccountManager, AccountManagerChainConfig } from './accountmanager';
+import { createUserProfile } from './profile';
+
+describe('AccountManager', () => {
+  const chain1: AccountManagerChainConfig = {
```
apanizo: I understand what you are trying to do here. But I do not like in the sense it increases complexity. In the real app, are we going to handle (load - parse - use) chains in this way? I have the feeling, this is a conversion just to test something in a limited scope. This does not represent reality (in the sense of code executed by the app), right?

packages/sanes-chrome-extension/src/logic/user/accountmanager.ts
```diff
@@ -0,0 +1,160 @@
...
+import { UserProfile, WalletId } from '@iov/core';
+import { Slip10RawIndex } from '@iov/crypto';
+import { ReadonlyWallet } from '@iov/keycontrol/types/wallet';
+
+export interface AccountInfo {
+  readonly name: string;
+  readonly identities: ReadonlyMap<ChainId, PublicIdentity>;
+}
+
+export interface AccountManagerChainConfig {
```
apanizo: This is the part I would give another spin. Correct me if my assumptions are wrong.

You are trying to reduce complexity in Persona, exporting all the accounts related internal methods outside. Placing it inside the user folder. At this point, I see two weaknesses:
- User fights with persona and can mislead.
- mnemonic method is here because it needs the user profile, it is not related semantically with Account.

I do like extracting logic for making simpler the persona, but I suggest change folder into something like:
```
logic/index.ts // methods to be used by the app for recreating persona, also for having the method if it is locked?
logic/persona/persona.ts // persona entity
logic/persona/personaFactory.ts //static class used to keep identity in a singleton way having functionality to clear memory if needed, being ready to grow once we support multiple personas.
logic/persona/accountGenerator.ts // file containing logic for generating accounts. Keep in mind in the future this probably will change, in order to generate names in a deterministic way. And placing in one single file all "dense" methods we have created is great because it does not break the open-close principle.
```

Having this file as it is, once we need to change something on how we generate accounts we are going to modify modules which contain an absolutely not related logic.

packages/sanes-chrome-extension/src/logic/user/accountmanager.ts
```diff
@@ -0,0 +1,160 @@
...
+          wallet,
+          chainId,
+          path
+        );
+        await this._userProfile.setIdentityLabel(identity, `${derivation}`);
+      }
+    }
+  }
+
+  public async generateNextAccount(): Promise<void> {
```
apanizo: I would add some documentation here explaining the reasoning for generating naming in accounts.

webmaster128:
graph

This is the current overview of the dependencies. The old Persona with the account management logic was renamed to `AccountManager` and a new high level `Persona` was created. During yesterday's refactoring I was able to get rid of the two step configuration, which required connecting to the network.

tsconfig.json
webmaster128: This change converts compile errors into linter error. The good thing about having those as linter errors is that you can deactivate the linter on a case by case basis if you want to explicitly ignore it. This is especially useful during development when some parts of the code are already created but not yet connected. I changed this because I faced a case where with this compile setting, we could not make any progress:

```
  const onSignup = async (formValues: FormValues): Promise<void> => {
    // eslint-disable-next-line @typescript-eslint/no-unused-vars, no-unused-vars
    const password = formValues[PASSWORD_FIELD];
    accountName.current = formValues[ACCOUNT_NAME_FIELD];
```

But since the linter still complains, changing this TypeScript config should not lead to bad code or bugs.

packages/sanes-chrome-extension/src/logic/persona/persona.ts
webmaster128: When the application creates a single `Persona` only, you don't need to `destroy` the Persona. But in the Persona tests, I create a new persona for each test that run in the same process. Those persona instances need to be distroyed in order to close the open connections to the blockchains.

packages/sanes-chrome-extension/src/logic/persona/persona.ts
webmaster128: Right, because in `const positiveResults = accountResults.filter(acc1 => !!acc1).map(acc2 => ...acc2.balance)` the type of `acc2` is still `Account | undefined`. This can be avoided by writing an additional type checking method but unless we don't need this very often, I don't think it makes sense.

packages/sanes-chrome-extension/src/logic/persona/persona.ts
webmaster128: > Also, the fact of one entity know how to create itself seems to be odd, what is your reasoning behind this? Maybe I am missing something.

The reason for putting the creator function into the `Persona` class is: it allows me to make the constructor private. This is important to avoid misuse of the constructor. The constructor has certain requirements for the parameters we put into it (in this case a shared profile instance) and by having exactly one place where this constructor is called, we can avoid misuse.

So `Persona.create()` is the way to create a single persona instance. Every additional funtionality, we can build around it in a separate class as soon as we need it (like the PersonaFactory idea).

packages/sanes-chrome-extension/src/logic/user/accountmanager.test.ts
webmaster128: What this does is breaking down a big problem into smaller problems. Those smaller problems can be looked at, solved and tested one by one. As soon as the smaller problem is solves, we can look at the bigger problem while being sure, that our small problem is solved in a solid way.

The existance of AccountManager makes it possible to handle all the accounts+chains logic without bothering with a network connection.

packages/sanes-chrome-extension/src/logic/user/accountmanager.ts
webmaster128: Thanks for your feedback. There are some easy improvements I can make directly based on this:
- Move accountmanager.ts into the logic/persona folder since it is a direct dependency of the Persona. Not export `AccountManager` in `logic/persona/index.ts` because it should only be used inside the folder, not from outside.
- Remove `AccountManager.mnemonic`, which can go to the Persona entirely

packages/sanes-chrome-extension/src/logic/persona/persona.ts
webmaster128: Right, adding chains at runtime will not be needed. A new chain is always added after a config update and then a new `AccountManager` is created with different settings. I removed this now.

PR #92, opened 2019-04-16
Create PersonaManager

webmaster128:
Following Adolfos idea in https://github.com/iov-one/ponferrada/pull/84 with a one adjustment:
Call it "manager" as the primary functionality is to manage Persona instances and make them available in a static fashion. Also in the default case, the class does not create more than one object.

This allows us to be explicit about the lifetime of Persona instances in tests and destroy them all after the tests, closing all open socket connections.


packages/sanes-chrome-extension/src/logic/persona/personamanager.ts
```diff
@@ -0,0 +1,44 @@
...
+    if (!PersonaManager.instance) {
+      throw new Error('Persona instance does not exist');
+    }
+    return PersonaManager.instance;
+  }
+
+  /**
+   * Gets the global Persona instance when there is one
+   */
+  public static getWhenCreated(): Persona | undefined {
```
apanizo: React views & components won't access Persona directly, they will access the info in a different way. Not directly because we lose updates on the data, and react by design should:
1. Not have an idea where data comes from, meaning react only cares about received props.
2. Should be dynamic in order to recompose the react tree on data updates (we do not want mutations)

So this method should be removed. For now, if you want, wrap access to Persona in views using try/catch.

packages/sanes-chrome-extension/src/logic/persona/personamanager.ts
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  /**
+   * Gets the global Persona instance
+   *
+   * Throws if instance does not exist, which indicates a bug in the app life cycle.
+   */
+  public static get(): Persona {
+    if (!PersonaManager.instance) {
+      throw new Error('Persona instance does not exist');
```
apanizo: When I follow this pattern I am used to creating the Persona if it does not already exist. But I understand it has downsides:
1. If you recreate it (in the future maybe we can do that) we would need to manage the auto destroy of the previous instance.
2. Harder to trail errors
3. Seeing the only entry point will be a "react" PersonaProvider, is not that bad use there the  annoying try/catch

So if you want to keep as it is, despite is not the "recommended" way, SGTM.

packages/sanes-chrome-extension/src/routes/recovery-phrase/index.dom.spec.ts
```diff
@@ -1,23 +1,32 @@
...
-import { sleep } from '../../utils/timer';
+
 import { withChainsDescribe } from '../../utils/test/testExecutor';
+import { PersonaManager } from '../../logic/persona';
 
 withChainsDescribe('DOM > Feature > Recovery Phrase', () => {
   let store: Store<RootState>;
 
+  beforeAll(async () => {
+    // All tests in this file should use the same persona instance to improve test speed
```
apanizo: Remove this comment. We are doing this here wrongly:
- no persona needed & used in this test (for now)
- broken the feature rule of creating a persona and showing its recovery phrase using react, meaning we have not tested in a pure and real correct way the feature (which at the ends is what dom test do)

Because:
- We gain more making visual and clear the necessity of destroying personas due to the bugs we have had in the past (hangs builds due to open connections, mostly).

packages/sanes-chrome-extension/src/routes/recovery-phrase/index.tsx
```diff
@@ -13,7 +13,10 @@ const RecoveryPhrase = (): JSX.Element => {
   const [mnemonic, setMnemonic] = React.useState<string>('');
 
   React.useEffect((): void => {
-    getMnemonic().then((mnemonic: string): void => setMnemonic(mnemonic));
+    // when this screen is created independent of app (e.g. storybook tests),
+    // persona is undefined
+    const persona = PersonaManager.getWhenCreated();
```
apanizo: As I said before wrap it in a try/catch, the `getWhenCreated` method only adds confusion.

packages/sanes-chrome-extension/src/logic/persona/personamanager.ts
```diff
@@ -0,0 +1,44 @@
+import { Persona } from './persona';
```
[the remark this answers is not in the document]
webmaster128: I usually use all lower but there is no reason for that. It can be changed to Persona.ts, PersonaManager.ts etc., no problem.

packages/sanes-chrome-extension/src/logic/persona/personamanager.ts
webmaster128: Looking into the future, when Persona is either randomly generated, loaded from DB or restored from a mnemonic, I think we will run into all kind of trouble if we don't create the Persona explicitly. Thus I prefer the current solution.

packages/sanes-chrome-extension/src/logic/persona/personamanager.ts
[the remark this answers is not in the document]
apanizo: In React apps only components start with capital letters. In this case should be: `personaManager.ts, persona.ts...`

packages/sanes-chrome-extension/src/routes/recovery-phrase/index.dom.spec.ts
webmaster128: If we remove the creation of the Persona in this test, then we need to remove this check:

```
    const phraseParagraph = paragraph.innerHTML;
    expect(phraseParagraph).toBe(PersonaManager.get().mnemonic);
```

But maybe the test can inject a testing mnemonic instead. I adapt it now.

packages/sanes-chrome-extension/src/logic/persona/personamanager.ts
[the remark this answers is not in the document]
webmaster128: Thanks, updated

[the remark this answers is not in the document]
webmaster128: Thanks, updated

PR #104, opened 2019-04-17
Fix @material-ui/* dependencies

webmaster128:


apanizo:
I am not a real expert about how `lerna` handles dependencies. If it is possible I would like to have in `ponferrada/package.json` a `dendencies` property with those libraries.

Another topic in my head is there are no "real" differences using devDependencies and dependencies when you are not creating a library. But in this case, we have the `medulas-react-components` which can be perfectly a library in the future.

Said, that, summarising: If it is possible to have `dependencies` in root package.json I prefer to unify libraries there. If not feel free to merge.

webmaster128:
> If it is possible I would like to have in ponferrada/package.json a dendencies property with those libraries.

Unfortunatly not possible. Each package needs to define what it needs if it uses the stuff directly. I.e. bierzo-wallet does not need to add the dependencies of medulas-react-components but only the ones it imports in bierzo-wallet files. 

> Another topic in my head is there are no "real" differences using devDependencies and dependencies when you are not creating a library. 

I tend to agree in general. However, I was having compile errors in bierzo-wallet before this change. I don't understand in detail why but everything that needs to be available at runtime should be a dependency anyway.

webmaster128:
The same issue occurs right now for connected-react-router as part of `yarn start`

```
Failed to compile.

./src/index.tsx
Module not found: Can't resolve 'connected-react-router' in '/some_place/ponferrada/packages/bierzo-wallet/src'
```

PR #108, opened 2019-04-17
Create signing server in Persona and let example website send requests to it

webmaster128:
The current state of this is that the website calls the "getIdentities" RPC method every 2 seconds to query the Ethereum chains of the user. Unfortunatly at some point the data is serialized into JSON, which means the public key is not a Uint8Array anymore. This is a mayjor issue to be solved.


packages/bierzo-wallet/src/routes/welcome/components/ExtensionInteraction.tsx
```diff
@@ -0,0 +1,27 @@
+import { JsRpcRequest } from '@iov/core';
+import React from 'react';
+
+export class ExtensionInteraction extends React.Component {
+  componentDidMount(): void {
```
apanizo: This is old school, we do not follow this pattern anymore, but congratulations you figured out! Do not worry, we'll refactor soon (we use react hooks now).

packages/sanes-chrome-extension/src/logic/persona/persona.ts
```diff
@@ -96,7 +119,38 @@ export class Persona {
     return out;
   }
 
+  public startSigningServer(): UseOnlyJsRpcSigningServer {
```
apanizo: Move this, into a separate class, something like personaSignaling or something like that. Persona has enough keeping the props for interacting with chains and exposing methods for feeding react views. Don't you think?

packages/sanes-chrome-extension/src/logic/persona/persona.ts
```diff
@@ -12,6 +21,19 @@ import {
 } from '../config';
 import { AccountManager, AccountInfo, AccountManagerChainConfig } from './accountManager';
 
+/** Like JsRpcSigningServer but without functionality to create or shutdown */
+export interface UseOnlyJsRpcSigningServer {
```
apanizo: I prefer to move this away from Persona.

packages/bierzo-wallet/src/routes/welcome/components/ExtensionInteraction.tsx
```diff
@@ -0,0 +1,27 @@
...
+      const request: JsRpcRequest = {
+        id: 1,
+        method: 'getIdentities',
+        params: {
+          reason: 'I would like to know who you are on Ethereum',
+          chainIds: ['ethereum-eip155-5777'],
+        },
+      };
+
+      chrome.runtime.sendMessage(extensionId, request, response => {
```
apanizo: I think it is easier if the dummy web interacts directly with
```
// This in dummy web page
window.dispatchEvent(myCustomEvent)
const resolveCustomEvent = function (data) {
  window.removeEventListener('MY_CUSTOM_ANSWER_ID', resolveCustomEvent); // if needed
  console.log("Beautiful data received!!");
  console.log(data);
}
window.addEventListener('MY_CUSTOM_ANSWER_ID', resolveCustomEvent)
```

And then in the content script (which is the only place which is able to receive it), code what you have done here, using the `chrome` instance for talking to content script.

Reasons:
- Developers are more familiar with general window JS object
- We simplify the dummy webpage removing chrome instance from it.
- More natural??

packages/sanes-chrome-extension/src/extension/backgroundscript.ts
```diff
@@ -25,3 +27,26 @@ chrome.runtime.onMessage.addListener(
     }
   }
 );
+
+chrome.runtime.onMessageExternal.addListener((request, sender, sendResponse) => {
```
apanizo: Interesting, what is `onMessageExternal`? Does this allow to talk directly from websites?

apanizo: Interesting, according to [documentation](https://developer.chrome.com/apps/messaging#external), this is for talking with other "extensions" but you are receiving events from websites, right?

packages/sanes-chrome-extension/src/logic/persona/personaManager.ts
```diff
@@ -8,6 +8,23 @@ export class PersonaManager {
    */
   public static async create(fixedMnemonic?: string): Promise<Persona> {
     const newInstance = await Persona.create(fixedMnemonic);
+
+    const server = newInstance.startSigningServer();
```
apanizo: This should live in the chrome extension's background script, and data (if needed) should be passed to the extension via redux.

Find the `wrapStore(makeStore());` line before. Then doing store.dispatch(oneAction), you can have reducers & listeners in the extension for handling everything. This is the natural way of doing it.

Anyway, if you have found another way to send data from background script to extension, and being able to store it somehow in case of multiple extension open and closes let me know, we could remove redux at all (when I researched this, I did not find a proper way).

packages/sanes-chrome-extension/src/extension/backgroundscript.ts
webmaster128: See this mini-docs for website -> backrgound script messaging: https://github.com/iov-one/ponferrada/wiki/Communication-channels#website---background-script

packages/bierzo-wallet/src/routes/welcome/components/ExtensionInteraction.tsx
webmaster128: okay, thanks, done. Using `React.useEffect` now.

packages/sanes-chrome-extension/src/logic/persona/persona.ts
webmaster128: Since the `Persona` instance owns the `MultiChainSigner` instance without exposing it, the signing server also needs to live here. The server is just a wrapper around `MultiChainSigner` that allows processing commands from outside.

packages/sanes-chrome-extension/src/logic/persona/persona.ts
webmaster128: See discussion above

packages/bierzo-wallet/src/routes/welcome/components/ExtensionInteraction.tsx
webmaster128: See discussion about communication channels below

packages/sanes-chrome-extension/src/extension/backgroundscript.ts
webmaster128: Now that all communication interfaces are converted to JSON, we have all the options. I updated https://github.com/iov-one/ponferrada/wiki/Communication-channels to include _
Content script -> Background script -> extension_, which may be what we want, especially because we can do this without whitelisting websites.

packages/sanes-chrome-extension/src/logic/persona/personaManager.ts
webmaster128: The `MultiChainSigner` instance cannot be passed between UI process and background script as it is not copyable. We can either move the entire Persona class or nothing.

The advantage of keeping it in the UI process is that the connections to the blockchains are re-established as soon as the extension is re-opened. If we move it to the background script, a closed WebSocket connection will break the functionality of the extension as long as the browser is open. In https://github.com/iov-one/iov-core/issues/229 we want to build automatic retry and reconnect into IOV-Core. But we are quite far from getting this right. In the meantime we should use a new `MultiChainSigner` instance with new chain connections for every interaction.

webmaster128:
This is ready to merge from my point of view. It shows a proof of concept querying a list of accounts from the extension.

Using a different communcation channel (e.g. via content script) can be done as a next step.

PR #110, opened 2019-04-17
Feature - Txs in account status view

### LINKED ISSUE #25: Account Status - Layout
On the top section
---------------------
There is a back arrow at the top left that takes the user to the welcome page. 
In the middle a label "Account Status"

The Hamburger menu will be integrated here from #88 

The next section shows:
-------------------------
A drop-down listing the accounts and there is a link to create a new account which takes the user through the signup process (See issues #6 #7 #24 ) (for doing this we need to support multi-account from iov-core integration #105 )

There is a section showing the pending transactions, pending requests, rejected requests (for now just show the existing ones when we create the persona, no subscription should be implemented).

apanizo:
**Description**
This PR injects in the account status view the boxes for showing pending txs and transactions (errored and succeeded ones).

For doing that:
- Created functional components for listing transactions, included when they are empty
- Injected txs in PersonaProvider for feeding react views.
- Enhanced some existing components like Typography 
- Updated signup method redirecting to account status view after creating a Persona.
- Updated signup dom test 

It closes #25 

**Before and After**
There is no before and now, the view looks like:
Screenshot 2019-04-17 19 43 10

Screenshot 2019-04-17 19 43 17


packages/sanes-chrome-extension/src/logic/persona/persona.ts
```diff
@@ -75,6 +76,11 @@ export class Persona {
     return this.accountManager.accounts();
   }
 
+  // TODO implement
+  public async getTxs(): Promise<ReadonlyArray<ProcessedTx>> {
+    return [];
```
apanizo: @webmaster128 Added you as a reviewer just for letting you know I did not have time to implement this method. Is not a big deal, having the signer and the chainIds you can call `searchTx` on each chain. Feel free to implement/rearchitect in the way you can. I just need the `ReadonlyArray<ProcessedTx>`.

Thanks!

yarn.lock
```diff
@@ -7791,7 +7788,7 @@ file-loader@^3.0.1:
     loader-utils "^1.0.2"
     schema-utils "^1.0.0"
 
-file-saver@eligrey/FileSaver.js#1.3.8:
+"file-saver@github:eligrey/FileSaver.js#1.3.8":
```
webmaster128: I search half an hour for the root of this strange diff but I did not find anything. Could you please try a `yarn cache clean && yarn install`? Maybe we have luck with that

packages/sanes-chrome-extension/src/logic/persona/persona.ts
```diff
@@ -11,6 +11,7 @@ import {
   pathBuilderForCodec,
 } from '../config';
 import { AccountManager, AccountInfo, AccountManagerChainConfig } from './accountManager';
+import { ProcessedTx } from '../../routes/account/components/ListTxs';
```
webmaster128: A data class/file like Persona should not depend on a UI component to avoid circular dependencies. In this case this can easily be fixed by moving the data type `ProcessedTx` into `logic/persona/persona.ts`, where the `ProcessedTx` are created.

packages/sanes-chrome-extension/src/logic/persona/persona.ts
webmaster128: This can easily be done when #108 is in. The signing server exposes the necessary interface for the transactions

packages/sanes-chrome-extension/src/logic/persona/persona.ts
apanizo: Sounds good.

Issue #103, opened 2019-04-17
Clean bierzo-wallet welcome and payments view


apanizo:
**Description**
The `bierzo-wallet` project as it is has all parts to be a decent app but we need to polish it a bit.

Things to do:
- [x] Transform the welcome view into something minimalistic like [this](https://drive.google.com/file/d/1bNJP5tIXwn2FW10gIfUSCjSAzjpNCIMQ/view?usp=sharing) Basically the icon in the middle of the circle should be the IOV's logo. And the grey colour should be our palette's main.
- [x] Introduce HMR in reducers as well see [here](https://github.com/iov-one/ponferrada/blob/master/packages/sanes-chrome-extension/src/store/index.ts#L15) and [here](https://github.com/iov-one/ponferrada/blob/master/packages/sanes-chrome-extension/src/store/index.ts#L9)
- [x] The codification of the main view's layout should be done following the semantic approach. Which means not using htmls tags, just container components like (Block, Grid...) available in medulas-react-components. 
- [x] Create your first DOM test assuring when you click on the button in welcome view, you travel to send payments.
- [x] Stop using Muli Light and go back to Muli.
- [x] Introduce validation in the send payment form and hardcoded values on the dropdown.
- [x] Create one storybook story for each view

kmw101:
Issue is done

PR #112, opened 2019-04-22
Feature #103 - Changes the welcome view to a minimalistic neat style

abefernan:
**Description**
This PR changes the welcome view as suggested by Issue #103.

**Screenshots**
imagen


packages/bierzo-wallet/src/routes/welcome/index.tsx
```diff
@@ -1,30 +1,69 @@
...
...25uZWN0b3ItY3VydmF0dXJlPSIwIgogICAgICAgICAgIGQ9Im0gNy40MjEsMjkuOTcgaCA4LjM5IFYgMjAuODc0IEggNy40MiB2IDkuMDk2IHoiCiAgICAgICAgICAgaWQ9InBhdGgzODMyIiAvPgogICAgICA8L2c+CiAgICAgIDxtYXNrCiAgICAgICAgIGlkPSJtYXNrMzkwNCIKICAgICAgICAgd2lkdGg9IjkiCiAgICAgICAgIGhlaWdodD0iMTAiCiAgICAgICAgIHg9IjIyIgogICAgICAgICB5PSIyMCIKICAgICAgICAgbWFza1VuaXRzPSJ1c2VyU3BhY2VPblVzZSI+CiAgICAgICAgPHBhdGgKICAgICAgICAgICBzdHlsZT0iZmlsbDojZmZmZmZmIgogICAgICAgICAgIGlua3NjYXBlOmNvbm5lY3Rvci1jdXJ2YXR1cmU9IjAiCiAgICAgICAgICAgZD0ibSAyMi4yMDIsMjUuOTk3IDQuOTc3LC0wLjE0OCAtMi4xMjYsLTQuNzY4IGEgMy45MTMsMy45MTMgMCAwIDEgMy4yNTYsMC4yOCBjIDIuMDY0LDEuMTU1IDIuODcsMy45IDEuOCw2LjEyOSAtMS4wNywyLjIzIC0zLjYxMSwzLjEgLTUuNjc1LDEuOTQ0IC0xLjI2NiwtMC43MSAtMi4wNTYsLTIuMDE4IC0yLjIzMiwtMy40MzcgeiIKICAgICAgICAgICBpZD0icGF0aDM5MDIiIC8+CiAgICAgIDwvbWFzaz4KICAgICAgPGcKICAgICAgICAgbWFzaz0idXJsKCNlKSIKICAgICAgICAgaWQ9ImczODQxIj4KICAgICAgICA8cGF0aAogICAgICAgICAgIHN0eWxlPSJjbGlwLXJ1bGU6ZXZlbm9kZDtmaWxsOiM1ZGMwYjc7ZmlsbC1ydWxlOmV2ZW5vZGQiCiAgICAgICAgICAgaW5rc2NhcGU6Y29ubmVjdG9yLWN1cnZhdHVyZT0iMCIKICAgICAgICAgICBkPSJtIDIyLjIwMiwyOS45NDUgaCA4LjM4IHYgLTkuMDk2IGggLTguMzggeiIKICAgICAgICAgICBpZD0icGF0aDM4MzkiIC8+CiAgICAgIDwvZz4KICAgIDwvZz4KICA8L2c+Cjwvc3ZnPgo="
```
apanizo: You can create a file named `./assets/avatar.png`  and then just import it: `import avatar from `./assets/avatar.png`, in this way, when you compile the app, webpack will do some magic and optimize the build.

packages/bierzo-wallet/src/routes/welcome/index.tsx
```diff
@@ -1,30 +1,69 @@
...
+        />
+      </Avatar>
+      <h1 className={classes.heading}>IOV Wallet</h1>
```
apanizo: You can use `Typography` and specify h1 variant.

packages/bierzo-wallet/src/routes/welcome/index.tsx
```diff
@@ -1,30 +1,69 @@
...
+  avatarIcon: {
+    width: '50%',
+  },
+
+  heading: {
+    margin: '4rem 0 4rem 0',
+  },
+
+  buttonRoot: {
+    width: '15rem',
```
apanizo: Same here.

packages/bierzo-wallet/src/routes/welcome/index.tsx
```diff
@@ -1,30 +1,71 @@
...
+    width: '200px',
+    height: '200px',
+  },
+
+  avatarIcon: {
+    width: '50%',
+  },
+
+  heading: {
+    margin: '40px 0 40px 0',
```
apanizo: Wrap `Typography` with a `Block` and use the `marginTop` and `marginBottom` prop, then, remove the margin css style here.

The less css we write on top level components, more descriptive and more semantic we do the views. Then a developer just reading the render method will understand what is going on. CSS details should be on low level components.

packages/bierzo-wallet/src/routes/welcome/index.tsx
```diff
@@ -1,30 +1,71 @@
...
+    fontSize: '20px',
+  },
+
+  buttonRoot: {
```
apanizo: Be consistent. Remove the `Root` suffix 😉.

Also, specify the width of a button, does not seem to be natural, I suggest to wrap the button with a Block and use the width prop `<Block width="150px">...</Block>`, in this way you can remove this CSS rule.

abefernan: Button has inherent "width: auto". By wrapping it with `<Block width="150px">`, this happens:
imagen

abefernan: The same happens if I remove the width and height of the avatar class and wrap the avatar with a block of 200x200, it has a default size of 40x40 so this happens:
imagen

PR #115, opened 2019-04-22
Feature - Adding Signup route E2E tests

### LINKED ISSUE #93: Create e2e signup test
Signup is an important feature for the chrome extension because it starts the logic engine of it.  

An e2e test should be coded for assuring the functionality works having the extension installed in the browser.

albertandrejev:
**Description**
This PR adds E2E test for Signup route. 
It closes #93 

**Developer notes**
This PR also adds method to check for specific id existence. Method name is `findRenderedE2EComponentWithId`

### DIFF
--- a/packages/sanes-chrome-extension/src/routes/signup/index.e2e.spec.ts
+++ b/packages/sanes-chrome-extension/src/routes/signup/index.e2e.spec.ts
@@ -37,6 +37,7 @@ withChainsDescribe(
       await page.goto('chrome-extension://dafekhlcpidfaopcimocbcpciholgkkb/index.html', {
         waitUntil: 'networkidle2',
       });
+      page.on('console', msg => console.log('PAGE LOG:', msg.text()));
 
       await travelToSignupNewAccountStep(page);
       await submitAccountFormE2E(page, randomString(10), randomString(10));
--- a/packages/sanes-chrome-extension/src/routes/signup/index.e2e.spec.ts
+++ b/packages/sanes-chrome-extension/src/routes/signup/index.e2e.spec.ts
@@ -30,7 +30,9 @@ withChainsDescribe(
       }
     );
 
-    it('loads correctly', async (): Promise<void> => {
+    it('should redirect to signup route, fill required data, show recovery phrase and hint', async (): Promise<
+      void
+    > => {
       const page: Page = await browser.newPage();
       await page.goto('chrome-extension://dafekhlcpidfaopcimocbcpciholgkkb/index.html', {
         waitUntil: 'networkidle2',
--- a/packages/sanes-chrome-extension/src/routes/signup/index.e2e.spec.ts
+++ b/packages/sanes-chrome-extension/src/routes/signup/index.e2e.spec.ts
@@ -42,6 +42,6 @@ withChainsDescribe(
       await submitAccountFormE2E(page, randomString(10), randomString(10));
       await handlePassPhrase2E(page);
       await handleSecurityHintE2E(page, randomString(10));
-    }, 45000);
+    }, 60000);
   }
 );
--- a/packages/sanes-chrome-extension/src/routes/signup/index.e2e.spec.ts
+++ b/packages/sanes-chrome-extension/src/routes/signup/index.e2e.spec.ts
@@ -1,7 +1,7 @@
 import puppeteer, { Browser, Page } from 'puppeteer';
 import { EXTENSION_HEIGHT, EXTENSION_WIDTH } from '../../theme/constants';
 import { travelToSignupNewAccountStep } from './test/travelToSignup';
-import { submitAccountFormE2E, handlePassPhrase2E } from './test/fillSignupForm';
+import { submitAccountFormE2E, handlePassPhrase2E, handleSecurityHintE2E } from './test/fillSignupForm';
 import { randomString } from '../../utils/test/random';
 import { withChainsDescribe } from '../../utils/test/testExecutor';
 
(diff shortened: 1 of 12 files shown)


albertandrejev:
@apanizo I need your help on this. Those tests runs without any problems locally, but Travis did not complete it properly, I think something wrong inside Travis settings or docker. Because first step completed properly but data submission (to the blockchain) was not and thus redirect to next step was not succeeded.

packages/sanes-chrome-extension/src/routes/signup/index.e2e.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+import { randomString } from '../../utils/test/random';
+import { withChainsDescribe } from '../../utils/test/testExecutor';
+
+withChainsDescribe(
+  'DOM > Signup route',
+  (): void => {
+    let browser: Browser;
+
+    beforeEach(async (): Promise<void> => {
+      const CRX_PATH = require('path').join(__dirname, '../../../build');
```
apanizo: This is the second e2e test done, so this is the time to refactor it. In the same way, we have a general `createDom` method, we should create a `createBrowser` method and reuse it in both tests.

packages/sanes-chrome-extension/src/routes/signup/index.e2e.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+    afterEach(
+      async (): Promise<void> => {
+        await browser.close();
+      }
+    );
+
+    it('should redirect to signup route, fill required data, show recovery phrase and hint', async (): Promise<
+      void
+    > => {
+      const page: Page = await browser.newPage();
```
apanizo: The same applies from line 36 to 40.

packages/sanes-chrome-extension/src/routes/signup/test/fillSignupForm.ts
```diff
@@ -1,11 +1,29 @@
...
+export const submitAccountFormE2E = async (
+  page: Page,
+  accountName: string,
+  password: string
+): Promise<void> => {
+  await page.type(`input[name="${ACCOUNT_NAME_FIELD}"]`, accountName);
+  await page.type(`input[name="${PASSWORD_FIELD}`, password);
+  await page.type(`input[name="${PASSWORD_CONFIRM_FIELD}`, password);
+
+  await page.click('button[type="submit"]');
```
apanizo: Good job here, I wanted to do submitting the form but I prefer your solution. 👌

packages/sanes-chrome-extension/src/routes/signup/test/fillSignupForm.ts
```diff
@@ -56,6 +74,34 @@ export const submitAccountForm = async (
...
+export const handlePassPhrase2E = async (page: Page): Promise<void> => {
+  const checkbox = await page.$('input[type="checkbox"]');
+  expect(checkbox).not.toBeNull();
+
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  await checkbox!.click();
+
+  //await sleep(2000);
+  const mnemonic = await page.evaluate(
+    async (): Promise<string | null> => {
```
apanizo: I suggest:
```
const mnemonic = await page.evaluate(
    (): string | null => {
      const element = document.querySelector('p');
      if (!element) {
        return null;
      }

      return element.textContent;
    }
  );
```

apanizo: Also, have you considered the option to use: 
```
const mnemonicElem = page.$('p');
const mnemonic = mnemonicElem ? mnemonicElem.textContent || null
```

or something like that?

packages/sanes-chrome-extension/src/routes/signup/test/travelToSignup.ts
```diff
@@ -16,3 +19,9 @@ export const travelToSignup = async (store: Store): Promise<React.Component> =>
 
   return dom;
 };
+
+export const travelToSignupNewAccountStep = async (page: Page): Promise<void> => {
+  await page.click('button:nth-of-type(2)');
+
+  await findRenderedE2EComponentWithId(page, FIRST_STEP_SIGNUP_ROUTE);
```
apanizo: I had in mind to use `https://github.com/GoogleChrome/puppeteer/blob/v1.15.0/docs/api.md#pagewaitfornavigationoptions` for checking navigation, but seeing we have used ids in DOM tests I really like your solution.

packages/sanes-chrome-extension/src/routes/signup/test/fillSignupForm.ts
albertandrejev: I decided to use this approach because we are testing UI behavior, so we need to check submit button too.

packages/sanes-chrome-extension/src/routes/signup/test/fillSignupForm.ts
albertandrejev: Yes this solution is better, I think we should use `page` instance helper methods as far as possible instead  of `page.evaluate`.
But in this case we should call `page.evaluate` to get value of the element's text. So it is better to leave it as is. https://stackoverflow.com/questions/49388467/getting-property-from-elementhandle

apanizo: ok, sounds good to me. Just to let you know I was able to remove the `async` in my first proposed solution and it worked. Am I missing something?

PR #119, opened 2019-04-24
Feature #103 - Welcome page DOM test

abefernan:
**Description**
This PR is a WIP that has the code for a DOM test that checks if a user is correctly routed to "payment" when clicking on the button.

As of [cb0970b](https://github.com/iov-one/ponferrada/pull/119/commits/cb0970b32f7b487d5448f6a306357c910684bb4d), `yarn test` fails because I haven't managed to correctly integrate Jest with Typescript + Babel + Enzyme.


packages/bierzo-wallet/.env
```diff
@@ -0,0 +1 @@
+SKIP_PREFLIGHT_CHECK=true
```
apanizo: Why do you need this?

packages/bierzo-wallet/package.json
```diff
@@ -8,25 +8,63 @@
...
+    "start": "react-scripts start",
+    "test": "jest"
   },
   "browserslist": [
     ">0.2%",
     "not dead",
     "not ie <= 11",
     "not op_mini all"
-  ]
+  ],
```
apanizo: And this?

packages/bierzo-wallet/src/routes/welcome/index.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+describe('The welcome route', () => {
+  let fakeHistory, fakeReducer, fakeStore, wrapper: ReactWrapper, button: ReactWrapper;
+
+  beforeEach(() => {
+    fakeHistory = createMemoryHistory();
+    fakeReducer = createRootReducer(fakeHistory);
+    fakeStore = configureStore(fakeReducer, fakeHistory);
+
+    wrapper = mount(
```
apanizo: Maybe because you are using enzyme?

packages/bierzo-wallet/.env
abefernan: If the .env files are removed, `yarn start` fails. It's a temporary fix that I found so that I could continue implementing the test.

There is some kind of dependency problem that I don't fully understand. I believe it has to do with different dependencies needing different versions of Jest. I also had to upgrade react-scripts from 2.1.8 to 3.0.0 because of the issues that started to appear.

packages/bierzo-wallet/package.json
abefernan: Some solutions that I started to try in order to integrate the testing technologies used some kind of configuration for Babel and Jest. That config can be set in a separate file or in the package.json, like I did after this closing bracket.

packages/bierzo-wallet/src/routes/welcome/index.test.tsx
abefernan: Enzyme was the way that I found to mount the component and surround it with the necessary components to pass the Store, the History, and the ThemeProvider.

I believe the current problem is that Jest does not correctly parse the imported MedulasThemeProvider, so it fails. When I checked that the tests where correctly implemented was when I did not use the ThemeProvider component.

However, if I remove the ThemeProvider now, it also fails because the makeStyles function of the components expect a "theme" variable that before didn't use to expect (when I was not using theming in the CSS).

packages/bierzo-wallet/.env
apanizo: updating the scripts into the 3.0 version is dangerous, I will do that in the next Friday of code. Thanks for the explanation, tomorrow I will look into it.

packages/bierzo-wallet/package.json
apanizo: huh, using create-react-app should abstract you to all of this. I believe this is something related to enzyme.

packages/bierzo-wallet/src/routes/welcome/index.test.tsx
apanizo: I will look into it. Using enzyme is something which makes sense, of course, is a good library and well supported, but React updates like in the 18.3 version makes some problems into them. Let's wait for @albertandrejev, I want to hear his opinion as well, but my gut tells me to use the react test utilities done by react itself. We have used it in the past and is enough.

albertandrejev: To be honest we have three possible options to do testing:
1. Snapshot using storybook (I would strongly recommend to prepare stories for storybook for all used views in webapp);
2. We have DOM tests using React TestUtils;
3. We have E2E tests using puppeteer.
So I think adding additional option for testing will be too redundantly and not required. We should be consistent in all parts of application. Also we have examples for all cases in chrome extensions package.

apanizo: @abefernan left a request changes comment with steps to follow. Once they are done let me know and I will specify the next ones.

Issue #99, opened 2019-04-16
Restore account


kmw101:
As a user I want to restore my "persona":
I am asked to enter my 12 recovery words. The label is "Restore your account with your recovery words" - the help text "Enter your recovery words here".

There is a continue button which remains inactive until the recovery words have been entered (note there are 12 words expected).

If I click continue and I have entered the wrong recovery words I am shown an error toaster with the message "The words you entered are not correct".

If I correctly type in my recovery words I am taken to the account status page

- [x] Create restore account view and add story to storybook (PR #120)
- [x] Add to route (PR #121)
- [x] Add logic to restore account and show toast in case of error (PR #122)
- [x] Prepare tests (PR #122)

apanizo:
As it is done in the route' signup logic, we should follow the same procedure:
1. Import PersonaProvider context
2. Create persona using mnemonic
3. Load info into PersonaProvider

Something like:
```
const mnemonic = getMnemonicFromForm();
try {
  const persona = await PersonaManager.create(mnemonic); <--- Note here we can pass a specific mnemonic
  const accounts = await persona.getAccounts();

  personaProvider.update(accounts, persona.mnemonic);
} catch (err) {
  console.log('Error raised when creating persona');
  console.log(err);
}
```
A DOM test should be provided, also a e2e is a very specific feature (you can create initially the first persona manually, no need to travel to signup fill everythig and then come back. If you can do it easily, perfect, if not, just take the shortcut)

PR #120, opened 2019-04-25
Feature - Restore account view

albertandrejev:
**Description**
This PR will add Restore Account view to the extension, it will finish "Create restore account view and add story to storybook" part of the issue #99.

**Before**
Nothing

**After**
restore-account

### DIFF
--- a/packages/sanes-chrome-extension/src/routes/restore-account/index.tsx
+++ b/packages/sanes-chrome-extension/src/routes/restore-account/index.tsx
@@ -9,7 +9,7 @@ const onBack = (): void => {
   history.goBack();
 };
 
-const Signup = (): JSX.Element => {
+const RestoreAccount = (): JSX.Element => {
   const personaProvider = React.useContext(PersonaContext);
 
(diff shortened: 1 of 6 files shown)


packages/sanes-chrome-extension/src/routes/restore-account/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+const onBack = (): void => {
+  history.goBack();
+};
+
+const RestoreAccount = (): JSX.Element => {
+  const personaProvider = React.useContext(PersonaContext);
+
+  const onRestore = async (formValues: FormValues): Promise<void> => {
+    try {
+      const persona = await PersonaManager.create();
```
apanizo: Left a comment in #121 about how persona should be created. Also, I would like to see the dom & e2e test. This feature is important.

apanizo: I get it. Do not worry you will fix it in #122

PR #123, opened 2019-04-28
Feature #103 - Semantic components for payment view

abefernan:
**Description**
This PR removes the usage of HTML tags and replaces them with medulas/mui components as suggested by Issue #103.


packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
```diff
@@ -91,13 +93,13 @@ export const CurrencyToSend = ({ positionedClass }: Props) => {
       <Avatar classes={avatarClasses}>
         <FontAwesomeIcon icon={faUser} className={classes.avatarIcon} />
       </Avatar>
-      <label className={classes.sendLabel}>You send</label>
-      <div className={classes.currencyContainer}>
+      <InputLabel className={classes.sendLabel}>You send</InputLabel>
```
apanizo: This `InputLabel` can be replaced by `Typography` from `medulas-react-components`. If for any reason you need a label as final dom element, you can use Typography's component prop [see here](https://next.material-ui.com/api/typography/) and specify label there.

packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
```diff
@@ -33,7 +35,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     color: '#ffffff',
   },
 
-  sendLabel: {
+  sendLabelRoot: {
```
apanizo: I do not like the `Root` suffix, when you add the class to mui's component like:
```
const sendLabelClasses = {
    root: classes.sendLabelRoot,
  };
```
You directly connect the intentions. No need to add extra suffix  😉

packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
```diff
@@ -33,7 +35,7 @@ const useStyles = makeStyles((theme: Theme) => ({
...
     color: theme.palette.text.primary,
     fontFamily: '"Muli SemiBold", sans-serif',
```
apanizo: Also, you can remove the Muli stuff here, should be injected automatically from the MedulasThemeProvider.

packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
```diff
@@ -53,13 +55,13 @@ const useStyles = makeStyles((theme: Theme) => ({
     height: '3rem',
```
apanizo: Use PX, please.

packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
```diff
@@ -53,13 +55,13 @@ const useStyles = makeStyles((theme: Theme) => ({
...
   },
 
-  validityLabel: {
+  validityLabelRoot: {
     color: theme.palette.error.main,
     fontFamily: '"Muli SemiBold", sans-serif',
```
apanizo: Same applies here.

packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
```diff
@@ -53,13 +55,13 @@ const useStyles = makeStyles((theme: Theme) => ({
...
     fontSize: '1.4rem',
   },
 
-  balanceLabel: {
+  balanceLabelRoot: {
     color: '#a2a6a8',
     fontFamily: '"Muli SemiBold", sans-serif',
```
apanizo: Well, all Muli references in the project.

packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
```diff
@@ -53,13 +55,13 @@ const useStyles = makeStyles((theme: Theme) => ({
     height: '3rem',
   },
 
-  validityLabel: {
+  validityLabelRoot: {
     color: theme.palette.error.main,
```
apanizo: Also for the color you can use the `MUI Typography color prop` avoiding the class creation! Have a look:
`<Typography color="primary | secondary | textPrimary ... or even error">Foo bar</Typography>`

packages/bierzo-wallet/src/routes/payment/components/ReceiverAddress/index.tsx
```diff
@@ -18,7 +20,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     padding: '4rem',
   },
 
-  toLabel: {
+  toLabelRoot: {
     color: theme.palette.text.primary,
```
apanizo: Same applies here.

packages/bierzo-wallet/src/routes/payment/components/ReceiverAddress/index.tsx
```diff
@@ -4,52 +4,46 @@ import { Theme } from '@material-ui/core';
...
 }
 
 const useStyles = makeStyles((theme: Theme) => ({
-  paperRoot: {
+  paper: {
     display: 'flex',
     flexDirection: 'column',
     width: '100%',
-    padding: '4rem',
+    padding: '40px',
```
apanizo: Again, this should be solved with `Block` component which is based on:
https://next.material-ui.com/system/spacing/

abefernan: I'm not sure if I should delete the Paper component completely and apply its styles to Block, or surround Paper with Block. What do you think?

packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
```diff
@@ -6,98 +6,89 @@ import Paper from '@material-ui/core/Paper';
...
   },
 
-  sendLabel: {
-    color: theme.palette.text.primary,
-    fontFamily: '"Muli SemiBold", sans-serif',
-    fontSize: '1.4rem',
-  },
-
   currencyContainer: {
     display: 'flex',
```
apanizo: Where this prop is used, you should be able to use directly in the component: https://next.material-ui.com/system/flexbox/

packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
```diff
@@ -6,98 +6,89 @@ import Paper from '@material-ui/core/Paper';
...
-  },
-
-  textFieldRoot: {
-    height: '5rem',
-    marginRight: '1rem',
-  },
-
-  selectRoot: {
-    height: '3rem',
+    margin: '40px 0 8px 0',
```
apanizo: Same with margin, you can use:
```
<Block marginTop={5} marginBottom={1}> <-- Remember the spacing unit.
```

packages/bierzo-wallet/src/routes/payment/components/ReceiverAddress/index.tsx
apanizo: I would say to apply styles directly to Block because you are simplifying things and also you avoid the usage of an extra component.

packages/bierzo-wallet/src/routes/payment/components/ReceiverAddress/index.tsx
```diff
@@ -4,52 +4,38 @@ import { Theme } from '@material-ui/core';
...
-    height: '5rem',
-    margin: '1.6rem 0 .8rem 0',
-  },
-
-  validityLabel: {
-    color: theme.palette.error.main,
-    fontFamily: '"Muli SemiBold", sans-serif',
-    fontSize: '1.4rem',
+    height: '50px',
+    margin: '16px 0 8px 0',
```
apanizo: What about wrap the `textField` with a Block and use the margin attributes?

packages/bierzo-wallet/src/routes/payment/components/ReceiverAddress/index.tsx
```diff
@@ -4,52 +4,38 @@ import { Theme } from '@material-ui/core';
...
   tooltipContainer: {
     display: 'flex',
     alignSelf: 'flex-end',
-    fontSize: '1.4rem',
-    marginTop: '2.4rem',
-  },
-
-  tooltipLabel: {
-    color: theme.palette.text.primary,
+    fontSize: '14px',
```
apanizo: What about remove the fontSize here and include it in the FontAwesomeIcon (if needed) and also in the Typography component using the variant prop.

Do the same with the marginTop prop, is better to wrap it using a block and specify the margin via props.

packages/bierzo-wallet/src/routes/payment/components/TextNote/index.tsx
```diff
@@ -10,34 +10,34 @@ interface Props {
 }
 
 const useStyles = makeStyles(() => ({
-  paperRoot: {
+  paper: {
```
apanizo: Can you remind me the usage of `positionedClass`?

Remember the rule for styles, fontSizes should be declared using Typography and its variant prop, not directly the size, in this way, if we change the Typography component, everything will be updated accordingly.

And the margin and padding should be specified in Block component via props, not css.

PR #124, opened 2019-05-02
Feature #103 - Payment Form validation

abefernan:
**Description**

This PR introduces validation in the send payment form and hardcoded values on the dropdown as suggested by Issue #103.

### DIFF
--- a/packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
+++ b/packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
@@ -47,9 +47,9 @@ const currencyItems = Object.keys(currencies).map(currency => {
   return item;
 });
 
-const onSubmit = () => {};
+const onSubmit = (): void => {};
 
-const CurrencyToSend = () => {
+const CurrencyToSend = (): JSX.Element => {
   const classes = useStyles();
   //NOTE hardcoded initial state for "currency"
   const [currency, setCurrency] = useState(currencyItems[1].name);
@@ -59,7 +59,7 @@ const CurrencyToSend = () => {
     root: classes.avatar,
   };
 
-  const handleChange = (item: Item) => {
+  const handleChange = (item: Item): void => {
     setCurrency(item.name);
     setBalance(currencies[item.name]);
   };
--- a/packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
+++ b/packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
@@ -1,6 +1,6 @@
 import { faUser } from '@fortawesome/free-regular-svg-icons';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import { Theme } from '@material-ui/core';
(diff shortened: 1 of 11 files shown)


packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
```diff
@@ -38,16 +96,26 @@ export const CurrencyToSend = () => {
...
+            <TextField
+              placeholder="0,00"
+              fullWidth
+              onBlur={textFieldHandleBlur}
+              onChange={textFieldHandleChange}
+            />
           </Block>
-          <Select />
+          {/*TODO properly style Select component*/}
+          <Select value={currency} onChange={selectHandleChange} autoWidth>
```
apanizo: There is a SelectField in `medulas-react-components` you can use.

packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
```diff
@@ -38,16 +96,26 @@ export const CurrencyToSend = () => {
         </Typography>
         <Block display="flex" marginTop={5} marginBottom={1}>
           <Block marginRight={1}>
-            <TextField placeholder="0,00" />
+            <TextField
```
apanizo: Use the TextField from medulas here as well, and delete the onBlur stuff.

packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
```diff
@@ -38,16 +96,26 @@ export const CurrencyToSend = () => {
         </Typography>
         <Block display="flex" marginTop={5} marginBottom={1}>
```
apanizo: You should wrap all fields elements in a Form, which is also available in medulas, doing this, you should implement validation in the same way as we are doing, in the same package as in medulas you can see a story, or directly find forms in the extension.

packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
```diff
@@ -20,13 +24,86 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+
+const CurrencyToSend = () => {
   const classes = useStyles();
 
   const avatarClasses = {
     root: classes.avatar,
   };
 
+  //NOTE hardcoded initial state for "currency"
+  const [currency, setCurrency] = useState(currencyItems[1].name);
```
apanizo: Please put all hooks once after other, meaning move down the avatarClasses. In this way is easier for the eye and harder to introduce bugs.

apanizo:
Do not forget to rebase! 😉

packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
abefernan: Fixed by [99da2af](https://github.com/iov-one/ponferrada/pull/124/commits/99da2af71c58070eed2dffe202fc5b44feb007da).

PR #125, opened 2019-05-02
Improve blockchain start scripts

webmaster128:


webmaster128:
> How can the scripts work without those lines? 

`docker run` automatically pulls the image from the internet if the image is not available locally. But if the image is already available locally (see `docker images`), this avoids connecting to the internet and checking for the tag. So this also saves a lot of time when starrting the chains.

PR #126, opened 2019-05-02
Move Persona to background script

webmaster128:
~Based on #125~


packages/sanes-chrome-extension/src/extension/backgroundscript.ts
```diff
@@ -22,30 +26,53 @@ chrome.runtime.onMessage.addListener(
         // do an operation in the extension
         // replying back to content script
         break;
+      case 'create_persona':
```
apanizo: can you create and use a constant for this 'create_persona'?

webmaster128: The unit type `readonly action: 'ctb_HELLO' | 'create_persona';` in `MessageToBackground` serves as a constant. Any other string value than those are rejected by the type system.

apanizo: I know that but... I am using to use the finder to find by references in order to allocate things faster. I think it is good for doing a brain connection of parts involved in the app.

I can do that by searching by string, of course, but is less... elegant, isn't it?

packages/sanes-chrome-extension/config/development.json
```diff
@@ -1,10 +1,9 @@
 {
-  "chains":
-  [
+  "chains": [
     {
       "chainSpec": {
         "codecType": "bns",
-        "bootstrapNodes": ["wss://bns.hugnet.iov.one/"]
```
apanizo: This change makes unit tests to fail because jsdom raises CORS errors. Options:
- ~Do not use react-scripts and call JEST directly after tweaking CORS stuff~
- ~Tweak and dig --env=node in unit testing. Need to fix react-scripts errors because it expects jsdom~
- Create another file called `config/unit.json` using websockets and update package.json' unit-test script.

My preferred option is the third one, otherwise, the effort done for keeping minimal the number of libraries needed would be for nothing.

packages/sanes-chrome-extension/src/extension/backgroundscript.ts
```diff
@@ -22,30 +25,57 @@ chrome.runtime.onMessage.addListener(
         // do an operation in the extension
         // replying back to content script
         break;
+      case MessageToBackgroundAction.CreatePersona:
```
apanizo: Thanks!

packages/sanes-chrome-extension/config/development.json
webmaster128: The CORS issue is basically a misconfuguration of the local BNS/BCP blockchains. It was never detected before because we always connected via WebSockets to those chains. The last commit fixes the blockchain configuration to allow all clients.

Some other test fails are probably left for next week but the CORS issue should be solved.

apanizo:
Related to travis errors.... maybe this is useful: https://stackoverflow.com/questions/44657140/jest-test-referenceerror-google-is-not-defined-reactjs-and-google-charts ?

webmaster128:
🎉

Issue #16, opened 2019-02-07
Share identity view [UI]


ethanfrey:
As a user shown a share identity request on the account status page.

I am shown the information about which site is requesting (through the signaling server) and for which token.

There are two buttons:
Accept
Reject

UX Guide: https://projects.invisionapp.com/share/PRPRZCQ9MHY#/screens/338674193

- [x] Create storybook and view of the screen (PR #129)
- [x] Create storybook and view of the Reject screen (PR #138)
- [x] Add route and implement DOM tests (PR #150)

kmw101:
Please ignore the confirmation page after accepting. It is not needed.

prince272:
I've rewritten the Asp Core Identity for Mvc. Feel free to take a look at https://github.com/prince272/AspCoreIdentityMvc

Simply change the url for the area 'Identity' to 'IdentityMvc' and you'll be automatically directed to the Mvc version for Identity.

kmw101:
Issue done

PR #130, opened 2019-05-03
Feature #16 - Share identity accept view

albertandrejev:
**Description**
This PR will add Share Identity Accept view and simple logic to the extension. It will implement "Create storybook and view of the Accept screen" part of the #16.

**Before**
N/A

**After**
accept-request


packages/sanes-chrome-extension/src/routes/share-identity/components/AcceptRequest.tsx
```diff
@@ -0,0 +1,43 @@
...
+      <Typography variant="body1" inline color="primary">
+        will be able to see
+      </Typography>
+      <Typography variant="body1" inline>
+        {' '}
+        your identity on
+      </Typography>
+      <Typography variant="body1" inline color="primary">
+        {' '}
+        BTC
```
apanizo: Change BTC because we do not support it yet. Just to avoid confusion. Use ' ETH'

Issue #133, opened 2019-05-06
Test error: Not implemented: HTMLCanvasElement.prototype.getContext


webmaster128:
We currently have those errors when executing DOM tests in package `sanes-chrome-extension`:

```
PASS src/routes/welcome/index.dom.spec.ts (8.829s)
  ● Console

    console.error ../../node_modules/jest-environment-jsdom/node_modules/jsdom/lib/jsdom/virtual-console.js:29
      Error: Not implemented: HTMLCanvasElement.prototype.getContext (without installing the canvas npm package)
          at module.exports (/some/place/ponferrada/node_modules/jest-environment-jsdom/node_modules/jsdom/lib/jsdom/browser/not-implemented.js:9:17)
          at HTMLCanvasElementImpl.getContext (/some/place/ponferrada/node_modules/jest-environment-jsdom/node_modules/jsdom/lib/jsdom/living/nodes/HTMLCanvasElement-impl.js:42:5)
          at HTMLCanvasElement.getContext (/some/place/ponferrada/node_modules/jest-environment-jsdom/node_modules/jsdom/lib/jsdom/living/generated/HTMLCanvasElement.js:50:45)
          at /some/place/ponferrada/node_modules/jspdf/dist/jspdf.min.js:270:5484
          at /some/place/ponferrada/node_modules/jspdf/dist/jspdf.min.js:270:7062
          at /some/place/ponferrada/node_modules/jspdf/dist/jspdf.min.js:1:89
          at Object.<anonymous> (/some/place/ponferrada/node_modules/jspdf/dist/jspdf.min.js:1:93)
          at Runtime._execModule (/some/place/ponferrada/node_modules/jest-runtime/build/index.js:858:13)
          at Runtime._loadModule (/some/place/ponferrada/node_modules/jest-runtime/build/index.js:551:12)
          at Runtime.requireModule (/some/place/ponferrada/node_modules/jest-runtime/build/index.js:415:10) undefined
```

There are discussions in
* [jest-canvas-mock](https://github.com/hustcc/jest-canvas-mock/issues/2)
* [jsom](https://github.com/jsdom/jsdom/pull/1964)

So this issue is probably fixed in jsdom 13.0.0+.

Unfortunately, our `jest-environment-jsdom` uses an old version of `jsdom`

```
jest-environment-jsdom@^24.7.1:
  version "24.7.1"
[...]

PR #134, opened 2019-05-06
Fix – Load jspdf only when PDFGenerator is constructed

webmaster128:
Loading the module 'jspdf' in an environment with no canvas available leads to the error message
> Error: Not implemented: HTMLCanvasElement.prototype.getContext

See also https://github.com/iov-one/ponferrada/issues/133

To avoid this error message to pollute test outputs, we load jspdf only when PDFGenerator is constructed, not when it is imported.

### DIFF
--- a/packages/sanes-chrome-extension/src/routes/recovery-phrase/utils/pdfGenerator.ts
+++ b/packages/sanes-chrome-extension/src/routes/recovery-phrase/utils/pdfGenerator.ts
@@ -1,15 +1,6 @@
-/**
- * This package cause jest-canvas-mock to print message to console.error (Error: Not implemented: HTMLCanvasElement.prototype.getContext (without installing the canvas npm package))
- *
- * But this message doesn't make test to fail so can be ignored.
- * Accroding to: https://github.com/hustcc/jest-canvas-mock/issues/2
- */
-import jsPDF from 'jspdf';
-import '../assets/Muli-normal';
-import '../assets/Muli-bold';
-
 class PDFGenerator {
-  private doc: jsPDF;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private doc: any;
   private readonly countWords: string[] = [
     'One',
     'Two',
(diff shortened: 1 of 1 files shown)


packages/sanes-chrome-extension/src/routes/recovery-phrase/utils/pdfGenerator.ts
```diff
@@ -26,6 +17,17 @@ class PDFGenerator {
...
 
   public constructor() {
+    // Loading the module 'jspdf' in an environment with no canvas available leads to the error message
+    // > Error: Not implemented: HTMLCanvasElement.prototype.getContext
+    // See also https://github.com/iov-one/ponferrada/issues/133
+    // To avoid this error message to pollute test outputs, we load jspdf only when PDFGenerator is
+    // constructed, not when it is imported.
+
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const { jsPDF } = require('jspdf');
```
webmaster128: this import is probably wrong (`const jsPDF = require('jspdf');`) but I could not read CI error message before it was restared

albertandrejev: No, there is an issue with Travis from time to time, it can not pass one E2E test, it failed by timeout (maybe because of lack of resources, I do not know). But restart helps it to finish. So please do not worry. At least I hope this will help.

albertandrejev: Hmm, it doesn't helps.

webmaster128: Let me try to reproduce locally. The test did not run locally before since I did not start the blockchains

apanizo: After researching a bit, I believe this should be:
`const jsPDF = require('jspdf');`

But it would not surprise me if using `cont { jsPDF } =` works as well.

albertandrejev: In my case (I got this fail some time ago) locally tests was running properly, but it was failed in Travis, and as I said restart was helpws to fix this.

albertandrejev:
Maybe you can add some description to this PR? Just comment you put top the sources.

packages/sanes-chrome-extension/src/routes/recovery-phrase/utils/pdfGenerator.ts
webmaster128: `const jsPDF = require('jspdf');` is correct, tested locally. PR was updated

PR #135, opened 2019-05-06
Feature #103 - Welcome page DOM test

abefernan:
**Description**
This PR adds code for a DOM test that checks if a user is correctly routed to "/payment" when clicking on the button of the "/welcome" route. It is part of the Issue #103.

### DIFF
--- a/packages/bierzo-wallet/src/routes/welcome/index.dom.spec.ts
+++ b/packages/bierzo-wallet/src/routes/welcome/index.dom.spec.ts
@@ -26,10 +26,6 @@ describe('The /welcome route', () => {
       }
     );
 
-    it('exists', () => {
-      expect(TestUtils.isDOMComponent(button)).toBeTruthy();
-    });
-
     it('redirects to the /payment route when clicked', () => {
       TestUtils.Simulate.click(button);
       expectRoute(store, PAYMENT_ROUTE);
--- a/packages/bierzo-wallet/src/routes/welcome/index.dom.spec.ts
+++ b/packages/bierzo-wallet/src/routes/welcome/index.dom.spec.ts
@@ -1,0 +1,38 @@
+import TestUtils from 'react-dom/test-utils';
+import { Store } from 'redux';
+import { configureStore } from '../../store';
+import { RootState } from '../../store/reducers';
+import { expectRoute } from '../../utils/test/dom';
+import { PAYMENT_ROUTE } from '../paths';
+import { travelToWelcome } from './test/travelToWelcome';
+
+describe('The /welcome route', () => {
+  let store: Store<RootState>;
+  let welcomeDom: React.Component;
+
+  beforeEach(
+    async (): Promise<void> => {
+      store = configureStore();
+      welcomeDom = await travelToWelcome(store);
+    }
+  );
+
+  describe('has a button that', () => {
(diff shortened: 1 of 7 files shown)


yarn.lock
```diff
@@ -7778,7 +7810,7 @@ file-loader@3.0.1, file-loader@^3.0.1:
     loader-utils "^1.0.2"
     schema-utils "^1.0.0"
 
-file-saver@eligrey/FileSaver.js#1.3.8:
+"file-saver@github:eligrey/FileSaver.js#1.3.8":
```
webmaster128: @abefernan the CI [just got a new check](https://github.com/iov-one/ponferrada/pull/137) to reject this diff. The issue is: after `yarn upgrade`, this is `"file-saver@github:eligrey/FileSaver.js#1.3.8":` but after `yarn install`, this is changed to `file-saver@eligrey/FileSaver.js#1.3.8:`. At the end of the day, this is an annoying yarn bug and nobody's fault.

To avoid this CI fail, use the following strategy:
1. If not necessary, do not `yarn upgrade`
2. If you really need to upgrade something, bump the version of the dependency in the package.json and then run `yarn install`

abefernan: Ok, thank you. Not sure how that ended up there if I never performed a `yarn upgrade`.

packages/bierzo-wallet/src/routes/welcome/index.dom.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+    let button: React.ReactInstance;
+
+    beforeEach(
+      async (): Promise<void> => {
+        button = await TestUtils.findRenderedDOMComponentWithTag(welcomeDom, 'button');
+      }
+    );
+
+    it('exists', () => {
+      expect(TestUtils.isDOMComponent(button)).toBeTruthy();
```
albertandrejev: No need for this test, in case if `findRenderedDOMComponentWithTag` did not find any element it will throw an exception https://reactjs.org/docs/test-utils.html#findrendereddomcomponentwithtag

abefernan: Fixed by [2f36278](https://github.com/iov-one/ponferrada/pull/135/commits/2f36278ffdd8bd5dd484b0aa572b64e5a1670308).

packages/bierzo-wallet/src/routes/welcome/index.dom.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+import { expectRoute } from '../../utils/test/dom';
+import { PAYMENT_ROUTE } from '../paths';
+import { travelToWelcome } from './test/travelToWelcome';
+
+describe('The /welcome route', () => {
+  let store: Store<RootState>;
+  let welcomeDom: React.Component;
+
+  beforeEach(
+    async (): Promise<void> => {
```
webmaster128: Small side note: we changed the linter settings, such that an explicit return type is not necessary, **if the function is unnamed**. This is especially useful in test code:

```ts
beforeEach(async () => {
  // lalalalala
});
```

abefernan: Thanks! I will take it into account.

PR #141, opened 2019-05-07
Feature – Navigate to show account when Persona exists

webmaster128:
Now that the Persona lives in the background script, we can check the existence when the UI is created

### DIFF
--- a/packages/sanes-chrome-extension/src/extension/messages.ts
+++ b/packages/sanes-chrome-extension/src/extension/messages.ts
@@ -10,6 +10,7 @@ export const WTC_MSG_HELLO = 'wtc_HELLO';
  */
 export enum MessageToBackgroundAction {
   CreatePersona = 'create_persona',
+  GetPersona = 'get_persona',
 }
 
 /**
@@ -51,6 +52,32 @@ export async function sendCreatePersonaMessage(mnemonic?: string): Promise<Creat
         mnemonic: response.mnemonic,
         txs: response.txs,
       });
+    });
+  });
+}
+
+export type GetPersonaResponse = PersonaData | null;
+
+export async function sendGetPersonaMessage(): Promise<GetPersonaResponse> {
+  return new Promise((resolve, reject) => {
+    const message: MessageToBackground = { action: MessageToBackgroundAction.GetPersona };
+    chrome.runtime.sendMessage(message, response => {
+      const lastError = chrome.runtime.lastError;
+      if (lastError) {
+        const errorMessage = lastError.message || 'Unknown error in chrome.runtime.lastError';
+        reject(errorMessage);
+        return;
+      }
+
+      if (!response) {
+        resolve(null);
+      } else {
+        resolve({
+          accounts: response.accounts,
+          mnemonic: response.mnemonic,
+          txs: response.txs,
+        });
+      }
     });
   });
 }
--- a/packages/sanes-chrome-extension/src/extension/messages.ts
+++ b/packages/sanes-chrome-extension/src/extension/messages.ts
@@ -22,11 +22,14 @@ export interface MessageToBackground {
   readonly data?: any;
 }
 
-export interface CreatePersonaResponse {
+export interface PersonaData {
   readonly accounts: ReadonlyArray<AccountInfo>;
   readonly mnemonic: string;
   readonly txs: ReadonlyArray<ProcessedTx>;
 }
+
+/* eslint-disable-next-line @typescript-eslint/no-empty-interface */
+export interface CreatePersonaResponse extends PersonaData {}
 
 export async function sendCreatePersonaMessage(mnemonic?: string): Promise<CreatePersonaResponse> {
   return new Promise((resolve, reject) => {
--- a/packages/sanes-chrome-extension/src/routes/welcome/index.tsx
+++ b/packages/sanes-chrome-extension/src/routes/welcome/index.tsx
@@ -10,20 +10,20 @@ const goToAccountStatus = (): void => {
 };
 
 const Welcome = (): JSX.Element => {
+  const isExtensionContext = typeof chrome !== 'undefined';
   const personaProvider = React.useContext(PersonaContext);
 
-  async function onLoad(): Promise<void> {
-    if (typeof chrome !== 'undefined') {
-      const response = await sendGetPersonaMessage();
-      if (response) {
-        personaProvider.update(response.accounts, response.mnemonic, response.txs);
-        goToAccountStatus();
-      }
-    }
-  }
-
   React.useEffect(() => {
-    onLoad();
+    (async () => {
+      if (isExtensionContext) {
+        const response = await sendGetPersonaMessage();
+        if (response) {
+          personaProvider.update(response.accounts, response.mnemonic, response.txs);
+          goToAccountStatus();
+        }
+      }
+    })();
+
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
--- a/packages/sanes-chrome-extension/src/routes/welcome/index.tsx
+++ b/packages/sanes-chrome-extension/src/routes/welcome/index.tsx
@@ -1,6 +1,33 @@
 import * as React from 'react';
 import Layout from './components';
+import { ACCOUNT_STATUS_ROUTE } from '../paths';
+import { history } from '../../store/reducers/';
+import { PersonaContext } from '../../context/PersonaProvider';
+import { sendGetPersonaMessage } from '../../extension/messages';
 
-const Welcome = (): JSX.Element => <Layout />;
+const goToAccountStatus = (): void => {
+  history.push(ACCOUNT_STATUS_ROUTE);
+};
+
+const Welcome = (): JSX.Element => {
+  const personaProvider = React.useContext(PersonaContext);
+
+  async function onLoad(): Promise<void> {
+    if (typeof chrome !== 'undefined') {
+      const response = await sendGetPersonaMessage();
+      if (response) {
+        personaProvider.update(response.accounts, response.mnemonic, response.txs);
+        goToAccountStatus();
+      }
+    }
+  }
+
+  React.useEffect(() => {
+    onLoad();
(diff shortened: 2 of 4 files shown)


packages/sanes-chrome-extension/src/routes/welcome/index.tsx
```diff
@@ -1,6 +1,33 @@
...
+import { history } from '../../store/reducers/';
+import { PersonaContext } from '../../context/PersonaProvider';
+import { sendGetPersonaMessage } from '../../extension/messages';
 
-const Welcome = (): JSX.Element => <Layout />;
+const goToAccountStatus = (): void => {
+  history.push(ACCOUNT_STATUS_ROUTE);
+};
+
+const Welcome = (): JSX.Element => {
```
apanizo: You are using an external dependency `personaProvider` in a function (`onLoad`) which is declared outside of the effect, and this could lead into bugs. I suggest:
```
const personaProvider = React.useContext(PersonaContext);

React.useEffect(() => {
  const fetchPersona = async () => {
      const extensionContext = chrome !== ' undefined'
      if (!extensionContext) {
          return
      }
     const response = await sendGetPersonaMessage();
      if (response) {
        personaProvider.update(response.accounts, response.mnemonic, response.txs);
        goToAccountStatus();
      }
  }

  fetchPersona();
}, []);
````

File: packages/sanes-chrome-extension/src/extension/messages.ts [hunk 2]
```diff
@@ -52,3 +56,29 @@ export async function sendCreatePersonaMessage(mnemonic?: string): Promise<Creat
...
+    const message: MessageToBackground = { action: MessageToBackgroundAction.GetPersona };
+    chrome.runtime.sendMessage(message, response => {
+      const lastError = chrome.runtime.lastError;
+      if (lastError) {
+        const errorMessage = lastError.message || 'Unknown error in chrome.runtime.lastError';
+        reject(errorMessage);
+        return;
+      }
+
+      if (!response) {
```
apanizo: I would do:
```
const persona = response ? { ...response } : null
resolve(persona);
```

packages/sanes-chrome-extension/src/extension/backgroundscript.ts
```diff
@@ -22,6 +22,26 @@ async function handleMessageToBackground(
         console.log('Received message from: ' + sender.tab.id);
       }
       return { msg: 'Hello mate' };
+    case MessageToBackgroundAction.GetPersona: {
+      if (sender.id !== chrome.runtime.id) {
+        return 'Sender is not allowed to perform this action';
+      }
+      let persona: Persona | undefined;
```
apanizo: I would do something like: 
```
try {
   const persona = PersonaManager.get(); <--- remember if persona does not exist it throws an exception (this functionality should be tested for assuring the exception is raised)
   return {
       mnemonic: persona.mnemonic,
       txs: await persona.getTxs(),
       accounts: await persona.getAccounts(),
    }
} catch () {
    return null;
}
```

packages/sanes-chrome-extension/src/extension/backgroundscript.ts
```diff
@@ -22,6 +22,26 @@ async function handleMessageToBackground(
         console.log('Received message from: ' + sender.tab.id);
       }
       return { msg: 'Hello mate' };
+    case MessageToBackgroundAction.GetPersona: {
+      if (sender.id !== chrome.runtime.id) {
```
apanizo: This condition should also be tested in an e2e test. IF this is out of the scope of this PR, create a ticket and assign it to me.

packages/sanes-chrome-extension/src/extension/messages.ts
```diff
@@ -23,12 +24,15 @@ export interface MessageToBackground {
...
 
-export interface CreatePersonaResponse {
+export interface PersonaData {
   readonly accounts: ReadonlyArray<AccountInfo>;
   readonly mnemonic: string;
   readonly txs: ReadonlyArray<ProcessedTx>;
 }
 
+/* eslint-disable-next-line @typescript-eslint/no-empty-interface */
+export interface CreatePersonaResponse extends PersonaData {}
```
apanizo: Why this?

packages/sanes-chrome-extension/src/routes/welcome/index.tsx
webmaster128: Your version does exactly the same thing as mine, just with the function declared in the `React.useEffect(` body.

The `onLoad` is captured in the `React.useEffect`, so no bug happens there.

packages/sanes-chrome-extension/src/extension/messages.ts
webmaster128: Since `response` is of type any, we should compose the `GetPersonaResponse` explicitly to make sure to copy the data we want. I tried to use only one resolve but the code does not become nicer than the `if`

```
      resolve(
        response
          ? {
              accounts: response.accounts,
              mnemonic: response.mnemonic,
              txs: response.txs,
            }
          : null
      );
```

packages/sanes-chrome-extension/src/extension/backgroundscript.ts
webmaster128: The important difference here is that I do not want to catch exceptions from `await persona.getTxs()` and `await persona.getAccounts()`. If I catch them, errors are interpreted as _no persona_ and not detected.

packages/sanes-chrome-extension/src/extension/backgroundscript.ts
webmaster128: https://github.com/iov-one/ponferrada/issues/145

packages/sanes-chrome-extension/src/extension/messages.ts
webmaster128: To be able to reuse `PersonaData` but have an explicit return type for `sendCreatePersonaMessage` analogue to `sendGetPersonaMessage`, which can change independently of `PersonaData`.

packages/sanes-chrome-extension/src/extension/backgroundscript.ts
apanizo: What happens if `getAccounts`  throws an Error? where is it caught?

What about...
```
try {
   const persona = PersonaManager.get(); <--- remember if persona does not exist it throws an exception (this functionality should be tested for assuring the exception is raised)
   return {
       mnemonic: persona.mnemonic,
       txs: await persona.getTxs(),
       accounts: await persona.getAccounts(),
    }
} catch (err) {
    console.log(err); <--- maybe here you can see where the error comes from
    return null;
}
```

Or what do you want to do with the error if it comes from `getAccounts`? I am curious.

packages/sanes-chrome-extension/src/routes/welcome/index.tsx
apanizo: This is just part of the react hooks guidelines for avoiding potential bugs due to possible future changes. I believe I saw some examples where the personaProvider could not be the latest one (in case it has changed) I think lint should complain. 

I also think there is no problem because you use `[]` which means, do that on load. But I also think this does not harm anyone and we make react happy. What do you think?

packages/sanes-chrome-extension/src/extension/backgroundscript.ts
apanizo: Just for giving more context, the main reason because I wrote this comment, was the difficulty to read the code.

packages/sanes-chrome-extension/src/extension/messages.ts
apanizo: And what about:
```
const persona = response ? { ...(response as Whatever) } : null
resolve(persona);
```

Again, I do not see anything wrong in your solution, just wanted to improve readability. Your decision!

packages/sanes-chrome-extension/src/routes/welcome/index.tsx
webmaster128: > I also think there is no problem because you use [] which means, do that on load. But I also think this does not harm anyone and we make react happy. What do you think?

The dependency `[]` means: no dependencies, run one on load. This is exactly what I want to do (and why the method is called `onLoad`: it should be run exactly once when the component is loaded.

If you want, I can move `async function onLoad() { ...}` into the `React.useEffect(` body like you did with `fetchPersona`. But that should not change anything.

packages/sanes-chrome-extension/src/extension/messages.ts
webmaster128: Casting with `as` only changes type information at compile time. At runtime, any garbage from `response` will be copied. So in this case I don't see a nicer way of writing it than the current version.

packages/sanes-chrome-extension/src/extension/backgroundscript.ts
webmaster128: > What happens if `getAccounts` throws an Error? where is it caught?

A primose rejection in handleMessageToBackground.

> the main reason because I wrote this comment, was the difficulty to read the code.

Yeah, there is a much simpler way to improve readability: Remove the obsolete `PersonaManager` and store the persona instance in the BS directly. I can make this as part of this PR if you like.

packages/sanes-chrome-extension/src/routes/welcome/index.tsx
apanizo: Just for following the react guidelines, please do.

packages/sanes-chrome-extension/src/extension/messages.ts
apanizo: I see. Sounds good then.

packages/sanes-chrome-extension/src/extension/backgroundscript.ts
apanizo: > A primose rejection in handleMessageToBackground.

Sounds good.

> store the persona instance in the BS directly

No, please, the BS will change in the incoming PRs, once it is stabilised we can think about it, but leave it as it is now.



👌

webmaster128:
Will do the final polishing as soon as #147 is merged

PR #142, opened 2019-05-07
Fix – Let CI fail on lint errors and warnings

webmaster128:


packages/medulas-react-components/src/components/forms/SelectFieldForm/index.tsx
```diff
@@ -83,6 +83,7 @@ const SelectFieldForm = ({
         onChange(initial);
       }
     } catch (err) {}
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
apanizo: This is something that I need to debug and refactor, do not silent it.

webmaster128: Could you instead create a ticket for that such that you have it in the tasks list? Without silencing it, the CI would fail.

apanizo: SGTM #143

PR #150, opened 2019-05-08
Feature #16 - Share identity DOM tests

albertandrejev:
**Description**
This PR will add DOM tests and route for the Share Identity view. It will implement "Add route and implement DOM tests" part of the #16.


packages/medulas-react-components/src/components/forms/CheckboxField/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const onCheckBoxChange = (_: React.ChangeEvent<HTMLInputElement>, checked: boolean): void => {
+    onChange(checked);
+    if (onChangeCallback) {
+      onChangeCallback(checked);
+    }
+  };
+
+  const control = (
+    <Checkbox
+      checked={value === '' ? false : value}
```
apanizo: What is value? a string? Maybe you just want to do `checked={!!value}` ?

packages/sanes-chrome-extension/src/routes/share-identity/test/operateShareIdentity.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export const clickOnAcceptButton = async (ShareIdentityDom: React.Component): Promise<void> => {
+  const inputs = TestUtils.scryRenderedDOMComponentsWithTag(ShareIdentityDom, 'button');
+
+  expect(inputs.length).toBe(2);
+
+  const acceptButton = inputs[0];
+
+  TestUtils.act(
+    (): void => {
```
webmaster128: We changed the linter settings, such that the return type only needs to be set when the function has a name. Especially in test code, you don't need to set it. This saves you 2 lines and one indent level:

```ts
TestUtils.act(() => {
  TestUtils.Simulate.click(acceptButton);
});
```

packages/sanes-chrome-extension/src/routes/share-identity/index.dom.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+  let identityDOM: React.Component;
+
+  beforeEach(async () => {
+    store = aNewStore();
+    identityDOM = await travelToShareIdentity(store);
+  });
+
+  it('should accept incoming request and come back', async (): Promise<void> => {
+    await clickOnAcceptButton(identityDOM);
+    await confirmSelectionButton(identityDOM);
```
apanizo: Let's follow the conversation in the front-end channel, but I do believe this screen is not necessary.

packages/sanes-chrome-extension/src/routes/share-identity/index.dom.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    //TODO: Check here that share request rejection has been reject successfuly
+
+    /**
+     * Remove this code if not required in case if there is another redirection
+     * in confirmAcceptButton method. And apply this method in separate test method.
+     */
+    await clickOnBackButton(identityDOM);
```
apanizo: What are you trying to do here? as far as I see in the invision mock, the final redirection is to the account status view. 🤔

albertandrejev: Ok, will check for account status view

albertandrejev: Hmmm.. where to redirect in case of requests queue?

albertandrejev: I got the answer

> Once the request is dispatched, the view should be redirected to: `request view` if there are more pending requests, or to `account-status view` if nothing is left.

PR #152, opened 2019-05-09
Change payment form validation approach

abefernan:
The goal of this PR is to make use of React Final Form Hooks' new [useField](https://github.com/final-form/react-final-form-hooks#usefield) implementation so that the validation of Bierzo Wallet's payment form can be made in a more modular and decoupled fashion.

Medulas has been updated to provide validators that can be used independently or combined in its TextFieldForm component. This way, instead of having to wrap each component with a dummy form, each field has a `validate`attribute that can be set like this to use independently:

```typescript
<TextFieldForm
    name={TEXTNOTE_FIELD}
    form={props.form}
    validate={notLongerThan(150)}
    placeholder="Add a note"
    multiline
    rows="2"
    fullWidth
    margin="none"
/>
```

Or that can be composed in order to use several validators:

```typescript
<TextFieldForm
    name={QUANTITY_FIELD}
    form={props.form}
    validate={composeValidators(
        required,
        number,
        lowerOrEqualThan(balance),
        lowerOrEqualThan(QUANTITY_MAX),
        greaterOrEqualThan(QUANTITY_MIN)
     )}
     placeholder="0,00"
     fullWidth
     margin="none"
/>
```

### DIFF
--- a/packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
+++ b/packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
@@ -103,7 +103,6 @@ const CurrencyToSend = (props: Props): JSX.Element => {
                 margin="none"
               />
             </Block>
-            {/*NOTE hardcoded initial value*/}
             <Block height="32px">
               <SelectFieldForm
                 fieldName={CURRENCY_FIELD}
--- a/packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
+++ b/packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
@@ -3,15 +3,18 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import Avatar from '@material-ui/core/Avatar';
 import Paper from '@material-ui/core/Paper';
 import { makeStyles } from '@material-ui/styles';
+import { FormApi } from 'final-form';
 import Block from 'medulas-react-components/lib/components/Block';
-import Form, {
-  FormValues,
-  useForm,
-  ValidationError,
-} from 'medulas-react-components/lib/components/forms/Form';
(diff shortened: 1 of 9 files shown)


packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
```diff
@@ -120,32 +86,36 @@ const CurrencyToSend = (): JSX.Element => {
...
+                  lowerOrEqualThan(balance),
+                  lowerOrEqualThan(QUANTITY_MAX),
+                  greaterOrEqualThan(QUANTITY_MIN)
+                )}
+                placeholder="0,00"
+                fullWidth
+                margin="none"
+              />
+            </Block>
+            {/*NOTE hardcoded initial value*/}
```
apanizo: You can remove this comment 😉

packages/bierzo-wallet/src/routes/payment/index.tsx
```diff
@@ -44,38 +44,37 @@ const onSubmit = (): void => {};
 const Payment = (): JSX.Element => {
   const classes = useStyles();
 
-  const { handleSubmit, invalid, pristine, submitting } = useForm({
+  const { form, handleSubmit, invalid, pristine, submitting } = useForm({
     onSubmit,
```
apanizo: Just FYI, here if you need to validate something at form level, let's say the user can send this payment, or whatever which does not fit in local field validation, it can be specified here. And this function will be validated if the local validation of fields passes.

Good job.

packages/medulas-react-components/package.json
```diff
@@ -23,7 +23,7 @@
     "@types/node": "11.9.4",
     "add": "^2.0.6",
     "final-form": "^4.12.0",
-    "react-final-form-hooks": "^1.0.1",
+    "react-final-form-hooks": "^2.0.0",
```
apanizo: 😍

PR #157, opened 2019-05-09
Feature – Show transactions in UI and create sendAndPost example

### LINKED ISSUE #107: Subscribe to transactions
As a user I want to subscribe to messages (transactions, requests etc) so that the display on the Account Status is updated as new messages are published.

webmaster128:
Closes #107

### DIFF
--- a/packages/bierzo-wallet/src/routes/welcome/index.tsx
+++ b/packages/bierzo-wallet/src/routes/welcome/index.tsx
@@ -34,7 +34,7 @@ const Welcome = (): JSX.Element => {
   const toast = React.useContext(ToastContext);
   const classes = useStyles();
 
-  const onGetIdentities = async (): Promise<void> => {
+  const onSendRequestToBeSigned = async (): Promise<void> => {
     toast.show('Interaction with extension, fetching identities. Check console, please.', ToastVariant.INFO);
     try {
       const identities = await sendGetIdentitiesRequest();
@@ -65,8 +65,8 @@ const Welcome = (): JSX.Element => {
         <Button className={classes.button} onClick={onPayment}>
           SEND PAYMENT
         </Button>
-        <Button className={classes.button} onClick={onGetIdentities}>
-          GET IDENTITIES
+        <Button className={classes.button} onClick={onSendRequestToBeSigned}>
+          SEND REQUEST TO BE SIGNED
         </Button>
       </Block>
     </Block>
--- a/packages/bierzo-wallet/src/routes/welcome/index.tsx
+++ b/packages/bierzo-wallet/src/routes/welcome/index.tsx
@@ -11,6 +11,7 @@ import { history } from '../../store/reducers';
 import { PAYMENT_ROUTE } from '../paths';
 import { ToastVariant } from 'medulas-react-components/lib/context/ToastProvider/Toast';
 import { sendGetIdentitiesRequest } from '../../communication/identities';
+import { sendSignAndPostRequest } from '../../communication/signAndPost';
 
 const useStyles = makeStyles((theme: Theme) => ({
   welcome: {
@@ -33,9 +34,16 @@ const Welcome = (): JSX.Element => {
   const toast = React.useContext(ToastContext);
   const classes = useStyles();
 
-  const onGetIdentities = (): void => {
+  const onGetIdentities = async (): Promise<void> => {
     toast.show('Interaction with extension, fetching identities. Check console, please.', ToastVariant.INFO);
-    sendGetIdentitiesRequest(toast);
+    try {
+      const identities = await sendGetIdentitiesRequest();
+      const transactionId = await sendSignAndPostRequest(identities[0]);
+      console.log('Transaction ID:', transactionId);
+    } catch (error) {
+      console.error(error);
+      toast.show('Error processing the request. Have you created your account?', ToastVariant.WARNING);
+    }
   };
 
   return (
--- a/packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
+++ b/packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
@@ -30,6 +30,29 @@ export const PersonaProvider = ({ children }: Props): JSX.Element => {
   const [txs, setTxs] = React.useState<ReadonlyArray<ProcessedTx>>([]);
   const isExtensionContext = typeof chrome !== 'undefined';
 
+  React.useEffect(() => {
+    if (isExtensionContext) {
+      chrome.runtime.onMessage.addListener((message, sender, _sendResponse) => {
+        if (sender.id !== chrome.runtime.id || !isMessageToForeground(message)) {
+          // Only handle messages from background script
+          return;
+        }
+
+        switch (message.action) {
+          case MessageToForegroundAction.TransactionsChanges:
+            if (!Array.isArray(message.data)) {
+              throw new Error('Data must be an array');
+            }
+            setTxs(message.data);
+            break;
+          default:
+            throw new Error('Unknown action');
+        }
+      });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
   const loadPersonaInReact = (
     accountNames: Accounts,
     mnemonic: string,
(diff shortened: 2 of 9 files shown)


packages/bierzo-wallet/src/routes/welcome/index.tsx
```diff
@@ -33,9 +34,16 @@ const Welcome = (): JSX.Element => {
   const toast = React.useContext(ToastContext);
   const classes = useStyles();
 
-  const onGetIdentities = (): void => {
+  const onGetIdentities = async (): Promise<void> => {
```
apanizo: rename `onGetIdentities` because is misleading. I suggest using `onSendRequestToBeSigned`. Also, update the text on line 69.

packages/sanes-chrome-extension/src/logic/persona/persona.ts
```diff
@@ -114,9 +118,14 @@ export class Persona {
...
 
-  // TODO implement
   public async getTxs(): Promise<ReadonlyArray<ProcessedTx>> {
-    return [];
+    if (!this.core) {
+      return [];
+    }
+
+    const processed = await Promise.all(this.core.signedAndPosted.value.map(s => this.processTransaction(s)));
+    const filtered = processed.filter(isNonNull);
```
apanizo: Why we can have null transactions?

packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
```diff
@@ -23,6 +25,7 @@ interface Props {
 }
 
 export const PersonaProvider = ({ children }: Props): JSX.Element => {
+  const isExtensionContext = typeof chrome !== 'undefined';
```
apanizo: Move this to line 32, hooks first.

packages/sanes-chrome-extension/src/extension/bsMessageHandler/internalHandler.ts
```diff
@@ -63,7 +70,20 @@ export async function handleInternalMessage(
           return true;
         };
 
-        signingServer = persona.startSigningServer(getIdentitiesCallback, signAndPostCallback);
+        async function transactionsChangedHandler(transactions: ReadonlyArray<ProcessedTx>): Promise<void> {
```
apanizo: Why this is `async`?

packages/sanes-chrome-extension/src/extension/messages.ts
```diff
@@ -81,3 +81,27 @@ export async function sendGetPersonaMessage(): Promise<GetPersonaResponse> {
...
+ * A message either from background script to foreground
+ */
+export interface MessageToForeground {
+  readonly type: 'message_to_foreground';
+  readonly action: MessageToForegroundAction;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  readonly data?: any;
+}
+
+export function isMessageToForeground(data: unknown): data is MessageToForeground {
```
apanizo: This function should be unit tested for understanding when this could be a problem. If this is trivial, then just a /** doc is enough.

packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
```diff
@@ -44,5 +47,25 @@ export const PersonaProvider = ({ children }: Props): JSX.Element => {
     update: loadPersonaInReact,
   };
 
+  if (isExtensionContext) {
+    chrome.runtime.onMessage.addListener((message, sender, _sendResponse) => {
+      if (sender.id !== chrome.runtime.id || !isMessageToForeground(message)) {
+        // Only handle messages from background script
```
apanizo: in this case, please do:
```
if (!msgFromBackgroundScript) {
    return;
}
```

packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
```diff
@@ -44,5 +47,25 @@ export const PersonaProvider = ({ children }: Props): JSX.Element => {
     update: loadPersonaInReact,
   };
 
+  if (isExtensionContext) {
+    chrome.runtime.onMessage.addListener((message, sender, _sendResponse) => {
```
apanizo: I believe this is the only penalty we have to pay because we do not use redux.

Do you see another cleaner mechanism to update txs besides this back and forth of messaged between BS - persona - PersonaProvider?

packages/sanes-chrome-extension/src/logic/persona/persona.ts
```diff
@@ -142,7 +151,8 @@ export class Persona {
 
   public startSigningServer(
     authorizeGetIdentities?: GetIdentitiesAuthorization,
-    authorizeSignAndPost?: SignAndPostAuthorization
+    authorizeSignAndPost?: SignAndPostAuthorization,
```
apanizo: Why those attrs are optional? I do not see another use case for the signing service without those callbacks, right?

packages/sanes-chrome-extension/src/logic/persona/persona.ts
webmaster128: Because we cannot process every transaction. Only `SendTransacrtion` can be converted into `ProcessedTx` at the moment.

packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
webmaster128: Passing messages is the only possible communication method between BS and UI, which are two independent JavaScript processes.

redux can possibly wrap this my hiding the messaging, but I have no idea if and how that would work.

packages/sanes-chrome-extension/src/logic/persona/persona.ts
webmaster128: The only reason is unit testing. If you want we can make them non-optional and move the default implementations into the test file.

packages/sanes-chrome-extension/src/extension/messages.ts
webmaster128: This is a trivial type checker. You can put anything into `data` and it telly you if that is data of type `MessageToForeground`. We need this since messages come as type  `any`.

packages/sanes-chrome-extension/src/extension/bsMessageHandler/internalHandler.ts
webmaster128: Good point. `async` is not necessary anymore after a refactoring

packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
webmaster128: I just tried

```ts
      const msgFromBackgroundScript = sender.id !== chrome.runtime.id || !isMessageToForeground(message);

      if (!msgFromBackgroundScript) {
        return;
      }
```

The problem is TypeScript needs the type checker `isMessageToForeground` directly to do the static analysis necessary to give `message` type `MessageToForeground`

packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
apanizo: At the end we are coding a chrome extension, so after thinking a bit, I think it is ok to have this listener in a react component. We have a great benefit of not using redux much less boilerplate code to maintain.

Once we have finished with everything we can think about other solutions.

One comment, seeing you have placed this in a react component, can you try to wrap this method in a React.useEffect? I do not want to subscribe more than one listener for this in case the component is re-rendered. Also, can you unsubscribe returning a function? Something like:
```
React.useEffect(() => { 
  if (isExtensionContext) {
    chrome.runtime.onMess....
    return chrome.runtime.onMessage.removeListener() <-- not sure if this does exist or is necessary.
 }
```

packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
apanizo: Not sure if I understand you, at the end `msgFromBackgroundScript` is a boolean, right? Why TS has complained? 🤔

webmaster128: TS does not understand that in the code I posted in the comment, `message` is always of type `MessageToForeground` after the `return;`. That means we loose type safety later in the listener if we do that.

webmaster128:
👍 to test this with the local chains, restore account with mnemonic "oxygen fall sure lava energy veteran enroll frown question detail include maximum"

PR #158, opened 2019-05-09
Feature #109 - Create left arrow in PageLayout Menu

### LINKED ISSUE #109: Create left arrow in PageLayout Menu
In some views, we need to integrate an arrow in the top left of the container. We need to do that in ```PageLayout``` component.

If the user clicks on the arrow it should go back, calling: `history.back()`

This option, should appear in account status view (related to #25)

albertandrejev:
**Description**
This PR will add left arrow button on the PageLayout menu and it will close #109.

**Before**
page-layout

**After**
page-layout-arrow

**Developer notes**
In order to use left arrow feature you should provide `onBack` prop to the PageLayout component.

### DIFF
--- a/packages/medulas-react-components/src/components/PageLayout/index.tsx
+++ b/packages/medulas-react-components/src/components/PageLayout/index.tsx
@@ -24,23 +24,21 @@ const PageLayout = ({ id, children, title, primaryTitle, onBack }: Props): JSX.E
       paddingTop={2}
       height="100%"
     >
-      <Block display="flex" flexDirection="row" alignItems="center">
-        {onBack && (
-          <Block marginLeft={-2}>
-            <Typography link>
-              <ArrowBackIcon fontSize="large" onClick={onBack} />
-            </Typography>
-          </Block>
-        )}
-        <Block>
-          <Typography color="primary" variant="h4" inline>
-            {primaryTitle}
-          </Typography>
-          <Typography variant="h4" inline>
-            {' '}
-            {title}
+      {onBack && (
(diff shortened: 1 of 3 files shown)


packages/medulas-react-components/src/components/PageLayout/index.tsx
```diff
@@ -3,15 +3,17 @@ import Block from '../Block';
...
   readonly id?: string;
   readonly children: React.ReactNode;
   readonly primaryTitle: string;
   readonly title: string;
+  readonly onBack?: () => void;
 }
 
-const PageLayout = ({ id, children, title, primaryTitle }: Props): JSX.Element => {
+const PageLayout = ({ id, children, title, primaryTitle, onBack }: Props): JSX.Element => {
   return (
```
apanizo: for adding more readability I would do:
```
  const showBackArrow = !!onBack;
  ...
  { showBackArrow && ( ....
```

packages/medulas-react-components/src/components/PageLayout/index.tsx
```diff
@@ -22,6 +26,13 @@ const PageLayout = ({ id, children, title, primaryTitle }: Props): JSX.Element =
       paddingTop={2}
       height="100%"
     >
+      {showBackArrow && (
+        <Block marginLeft={-2}>
+          <Typography link>
+            <ArrowBackIcon fontSize="large" onClick={onBack} />
```
apanizo: I do believe for having the same appearance as the hamburger menu you need to wrap it using `IconButton`

PR #160, opened 2019-05-10
Fix - Avoid exceeding maximum update depth

abefernan:
The goal of this PR is to fix a console warning noticed by @albertandrejev that said "Maximum update depth exceeded"

### DIFF
--- a/packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
+++ b/packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
@@ -15,7 +15,7 @@ import {
   number,
   required,
 } from 'medulas-react-components/lib/utils/forms/validators';
-import React, { useState } from 'react';
+import React, { useMemo, useState } from 'react';
 
 const useStyles = makeStyles(() => ({
   avatar: {
@@ -60,6 +60,16 @@ const CurrencyToSend = (props: Props): JSX.Element => {
   const [currency, setCurrency] = useState(currencyItems[1].name);
   const [balance, setBalance] = useState(currencies[currency]);
 
+  const validator = useMemo(() => {
+    return composeValidators(
+      required,
+      number,
+      lowerOrEqualThan(balance),
+      lowerOrEqualThan(QUANTITY_MAX),
+      greaterOrEqualThan(QUANTITY_MIN),
+    );
+  }, [balance]);
+
   const avatarClasses = {
     root: classes.avatar,
   };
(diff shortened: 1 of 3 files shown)


abefernan:
This PR fixes what it aims to, but introduces the following usability issue.

**Background**
The hardcoded values for each currency's balance are 15 for ALT and 24 for IOV. Because of this, when you use the *SelectFieldForm* to select *ALT*, the validator will render a warning if *TextFieldForm*'s input exceeds 15. In like manner, a warning will be rendered if the input exceeds 24 when *IOV* is selected:
imagen

**Current and desired behaviour**
If there is a warning currently displaying because the current amount exceeds the selected currency's balance, if a new currency is selected, the warning should display the new balance requirement or not display it at all if it's not being exceeded:
imagen


**Faulty behaviour introduced by this PR**
The warning is not updated until the input changes, that is to say, in the example on the right of the following screenshot you'd have to input or delete something and write the "20 "again.
imagen

apanizo:
🤔 can you guys give me more info? I do not like this error. Probably is a bug in the library.

abefernan:
> thinking can you guys give me more info? I do not like this error. Probably is a bug in the library.

The simplest occurrence of this warning takes place in the TextNote component, since it only uses one validator that does not depend on any variable.

The warning:
```
Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
in TextFieldForm (at TextNote/index.tsx:26)
```

If I extract the `notLongerThan(150)` function call to a variable **inside** the component like shown in the following snippet, the problem persists:
```typescript
const TextNote = (props: Props): JSX.Element => {
  const validator = notLongerThan(150);
  return (
...
              validate={validator}
...
};
```
However, if extracted outside of the component, the warning disappears:
```typescript
const validator = notLongerThan(150);
const TextNote = (props: Props): JSX.Element => {
  return (
...
              validate={validator}
...
};
```

apanizo:
@abefernan @albertandrejev do not merge this. Let me investigate.

apanizo:
@abefernan @albertandrejev today seems I will continue working on the enqueue request logic, let me know if this PR blocks you in some way.

abefernan:
> @abefernan @albertandrejev today seems I will continue working on the enqueue request logic, let me know if this PR blocks you in some way.

It does not, it just aims to remove a warning.

apanizo:
@albertandrejev can you rebase this PR?

albertandrejev:
@apanizo was rebased

apanizo:
@webmaster128 @abefernan @albertandrejev please, do not remove this branch.

PR #162, opened 2019-05-10
Feature - Improve extension load when persona has already created

apanizo:
**Description**
This PR solves the painful blinking when a persona is found in memory and the extension is redirected automatically to welcome route.


apanizo:
@webmaster128 I have not added you as a reviewer because this is mostly frontend code, just to let you know we do not have any more unpleasant blinking. When a persona is in memory and the view is redirected.

packages/sanes-chrome-extension/src/routes/account/index.tsx
```diff
@@ -5,60 +5,85 @@ import PageLayout from 'medulas-react-components/lib/components/PageLayout';
...
+  React.useEffect(() => {
+    async function fetchPersonaData(): Promise<void> {
+      const isExtensionContext = typeof chrome !== 'undefined';
+      if (!isExtensionContext) {
+        return;
+      }
+
+      const response = await sendGetPersonaMessage();
+      if (response) {
+        personaProvider.update(response.accounts, response.mnemonic, response.txs);
```
webmaster128: Could we do this in `sanes-chrome-extension/src/index.tsx` already? Then we do not need 2 round trips of `sendGetPersonaMessage` and we don't need to handle the impossible case that `response` is falsy here.

packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
```diff
@@ -22,35 +22,37 @@ export const PersonaContext = React.createContext<PersonaContextInterface>({
...
+  readonly persona: GetPersonaResponse;
 }
 
-export const PersonaProvider = ({ children }: Props): JSX.Element => {
-  const [accountNames, setAccountNames] = React.useState<Accounts>([]);
-  const [mnemonic, setMnemonic] = React.useState<string>('');
-  const [txs, setTxs] = React.useState<ReadonlyArray<ProcessedTx>>([]);
-  const isExtensionContext = typeof chrome !== 'undefined';
-
+export const PersonaProvider = ({ children, persona }: Props): JSX.Element => {
```
apanizo: @webmaster128 @abefernan @albertandrejev now passing `persona: GetPersonaResponse` here for hydrating data, will solve much time doing DOM testing because we can mock it 🎉

packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
```diff
@@ -22,35 +22,37 @@ export const PersonaContext = React.createContext<PersonaContextInterface>({
...
-            if (!Array.isArray(message.data)) {
-              throw new Error('Data must be an array');
-            }
-            setTxs(message.data);
-            break;
-          default:
-            throw new Error('Unknown action');
-        }
-      });
+    console.log('PersonaProvider registering listener');
```
webmaster128: Could you move this a few line down right above `chrome.runtime.onMessage.addListener(`? Otherwise the log is misleading in non-extension contexts.

PR #163, opened 2019-05-10
Feature - remove redux wrapper

apanizo:
**Description**
In this PR I remove the unused redux wrapper. Communication between background script and the extension is done using the recommended protocol communication.

@webmaster128 @albertandrejev do you guys think we can have a race condition here? I have assumed the background script is loaded before the extension is opened.


webmaster128:
> do you guys think we can have a race condition here? I have assumed the background script is loaded before the extension is opened.

I think this is safe to assume

PR #171, opened 2019-05-14
Feature – Update Ethereum scripts

webmaster128:
* Use latest Ethereum scripts from IOV-Core, including the `scripts/ethereum/README.md`
* Add empty Ethereum init script, which we will need soon
* move all blockchain scripts into repo roo `./scripts` for consistency with other projects and because all the packages can use them (bierzo-wallet will need to connect to the chains independently of sanes-chrome-extension)
* Update documentation for starting and stopping


webmaster128:
> Have you seen it? Did we change something on LSK scripts lately?

Yes I am experiencing the issue too (especially in a slow environment like Travis). The issue came since the update if the Lisk node from 1.4 to 1.6. Now the HTTP API respons before the node is ready, either with HTTP 500 (Server error) or with genesis block not yet applied (height=0). I fixed most of that in IOV-Core already. Right now I am working on some small edge case bug in the start script. When that is done and tested in IOV-Core, I'll copy over the updated scripts as well.

PR #172, opened 2019-05-14
Feature – Set all trailing commas

webmaster128:
This reduces the diff when adding or removing arguments (i.e. simplifies review and makes merging much easier)


webmaster128:
After this, I had to restart VSCode, such that the eslint plugin read the prettier config again

webmaster128:
Thanks @albertandrejev!

PR #179, opened 2019-05-14
Feature - Add Back Button to Login View

abefernan:
**Description**
This PR adds a Back button to the login view of the Chrome extension, and removes the redundant "More options" link.

**Current view:**
imagen

**Proposed view:**
imagen


albertandrejev:
There is two different ways to implement "back" feature. One is like it was done in this PR and second is to use `onBack` prop of the `PageLayout` component. @apanizo what do you think which way is better to use in this view?

abefernan:
I must update the DOM tests, I'm on it.

packages/sanes-chrome-extension/src/routes/login/components/LoginForm.tsx
```diff
@@ -1,17 +1,19 @@
-import * as React from 'react';
-import Button from 'medulas-react-components/lib/components/Button';
 import Block from 'medulas-react-components/lib/components/Block';
+import Button from 'medulas-react-components/lib/components/Button';
+import Back from 'medulas-react-components/lib/components/Button/Back';
 import Form, { useForm } from 'medulas-react-components/lib/components/forms/Form';
 import TextFieldForm from 'medulas-react-components/lib/components/forms/TextFieldForm';
+import * as React from 'react';
```
webmaster128: What's the reason for moving this imports (just to understand the idea)? In IOV-Core we use separate import blocks separated by an empty line from global to local (1. all external imports, 2. all imports from "friend" packages @iov/*, 3. relative file path imports). In ponferrada I did not see any particular order yet. What would be your preference?

abefernan: I wasn't aware there was a policy regarding import order and always performed Visual Studio Code's "Organize Imports" command before saving.

webmaster128: As far as I know, there is no policy in ponferrada yet (correct me if I'm wrong @apanizo @albertandrejev). But it would be good to have one. I like the idea of using Visual Studio Code's "Organize Imports", which is much faster than manually sorting imports.

abefernan: The goal of that command is to just run it without ever having to think about the imports, and it has served me really well.

Unfortunately, it seems to lack a customization layer that allows us to put together an in-house policy.

The only issue I found with it is that is not reliable to configure it to run onSave, so I usually do the following steps:

1. Edit a file
2. Ctrl+S (save + formatting on save)
3. Shift + Alt + O (Organize imports. Removes unused imports besides sorting them)
4. Ctrl+S again

webmaster128: I like that way @abefernan describes. The in-house policy at IOV-Core costs a lot of time manually sorting the imports for making them only slightly nicer.

abefernan: This friday I may take a look at projects such as this: https://marketplace.visualstudio.com/items?itemName=mike-co.import-sorter

albertandrejev: In `wallet-demo` project we used simple alphabetical import sorting. So I think this is suitable approach to use in `ponferrada` project.

PR #181, opened 2019-05-14
Feature #156 enqueue requests in chrome extension

### LINKED ISSUE #156: Queue request logic
There may be a number of requests by websites and to manage the multiple request the following logic needs implementing:

- [x] Enqueue in order txs
- [x] Show number of pending transactions in extension's icon

apanizo:
**Description**
This PR introduces the logic of enqueueing requests to the extension.

It closes #156 

You can see it in action [here](https://drive.google.com/open?id=1k1EB2ZTlR0Ra7afhcOV12Wy9ny2oVVr7)


packages/sanes-chrome-extension/src/extension/background/externalHandler/index.ts
```diff
@@ -0,0 +1,92 @@
...
+    parseMethod(request.method);
+  } catch (err) {
+    return generateErrorResponse(responseId, err.message);
+  }
+
+  if (SenderWhitelist.isBlocked(sender)) {
+    return generateErrorResponse(responseId, 'Sender has been blocked by user');
+  }
+
+  return new Promise(resolve => {
```
webmaster128: All the queueing should happen at the level of autorization requests. That is the point where the signing server asks the UI/user: _please handle the request and tell me what I should do with it_. Since the UI is not always available, those requests can be queued.

To queue pending callbacks, you can do:

```ts
interface QueuedElement {
  reason: string;
  senderUrl: string;
  accept(): void;
  reject(): void;
}

const queue = new Array<QueuedElement>();

const getIdentitiesCallback: GetIdentitiesAuthorization = async (
  _reason,
  matchingIdentities,
): Promise<ReadonlyArray<PublicIdentity>> => {
  // sender will be available here after upgrading to IOV-Core 0.13.7
  // https://github.com/iov-one/iov-core/pull/993
  if (isNewSender('http://finnex.com')) {
    chrome.browserAction.setIcon({ path: 'assets/icons/request128.png' });
    chrome.browserAction.setBadgeBackgroundColor({ color: [0, 0, 0, 0] });
    chrome.browserAction.setBadgeText({ text: '*' });
  }

  const pendingGetIdentititesRequest = new Promise<ReadonlyArray<PublicIdentity>>(resolve => {
    function acceptThisRequest(): void {
      resolve(matchingIdentities); // select all
    }

    function rejectThisRequest(): void {
      resolve([]); // select none
    }

    queue.push({
      reason: _reason,
      senderUrl: 'http://finnex.com',
      accept: acceptThisRequest,
      reject: rejectThisRequest,
    });
  });

  return pendingGetIdentititesRequest;
};
```

apanizo: I can create it in the internalHandler when I create the persona and therefore restart the RequestHandler easily. I do not see a blocker on this. So far so good.

What about signAndPost? That was my main problem, I receive the callback once the tx is signed, how can I reject it?

packages/sanes-chrome-extension/src/extension/backgroundscript.ts
```diff
@@ -28,7 +33,7 @@ chrome.runtime.onMessage.addListener((message: MessageToBackground, sender, send
  */
 chrome.runtime.onMessageExternal.addListener((request, sender, sendResponse) => {
   const signingServer = getSigningServer();
-  handleExternalMessage(signingServer, request)
+  handleExternalMessage(signingServer, request, sender.url)
```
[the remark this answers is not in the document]
apanizo: 👍

packages/sanes-chrome-extension/src/extension/background/externalHandler/index.ts
webmaster128: > What about signAndPost? That was my main problem, I receive the callback once the tx is signed, how can I reject it?

You receive the callback before the transaction is signed. You can click on the types to get a bunch of docs:

```ts
export interface SignAndPostAuthorization {
    /**
     * Called by the signing server and lets user decide if they want to
     * authorize a sign and post request.
     *
     * Must return true if the user authorized the signing and false if the
     * user rejects it.
     *
     * Any error (thrown or returned as a rejected promise) from this
     * callback is sent to the RPC client as an unspecified
     * "Internal server error" and the callback author should
     * ensure this does not happen.
     *
     * @param reason An explanation why the autorization is requested. This is created by the website and shown to the user.
     * @param transaction The transaction to be signed.
     * @param meta An object that is passed by reference from request handlers into the callback.
     */
    (reason: string, transaction: UnsignedTransaction, meta?: any): Promise<boolean>;
}
```

packages/sanes-chrome-extension/src/extension/background/createPersona/requestCallback.ts
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  return typeof (data as RequestMeta).senderUrl === 'string';
+}
+
+type CallbackResponse = ReadonlyArray<PublicIdentity> | boolean;
+
+async function requestCallback(
+  reason: string,
+  meta: RequestMeta,
```
webmaster128: this type is wrong: since you did not yet check what kind of data you got here, this is `any`. A couple of lines below, you have the check that ensures `meta` is the type you want.

packages/sanes-chrome-extension/src/extension/background/createPersona/requestCallback.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    RequestHandler.add({ reason, sender: senderUrl, accept, reject });
+    updateExtensionBadge();
+  });
+}
+
+export async function getIdentitiesCallback(
+  reason: string,
+  matchingIdentities: ReadonlyArray<PublicIdentity>,
+  meta: RequestMeta,
```
webmaster128: `meta: any`

packages/sanes-chrome-extension/src/extension/background/createPersona/requestCallback.ts
```diff
@@ -0,0 +1,69 @@
...
+): Promise<CallbackResponse> {
+  return requestCallback(reason, meta, matchingIdentities, []);
+}
+
+export async function signAndPostCallback(
+  reason: string,
+  _transaction: UnsignedTransaction,
+  meta: RequestMeta,
```
webmaster128: `meta: any`

packages/sanes-chrome-extension/src/extension/background/createPersona/requestHandler.ts
```diff
@@ -0,0 +1,62 @@
+interface Request {
+  readonly reason: string;
+  readonly sender: string;
+  readonly accept: () => void;
+  readonly reject: (permanently: boolean) => void;
+}
+
+export function parseMethod(method: string): void {
```
webmaster128: Method obsolete now

packages/sanes-chrome-extension/src/extension/background/createPersona/requestHandler.ts
```diff
@@ -0,0 +1,62 @@
...
+  const allowed = method === 'getIdentities' || method === 'signAndPost';
+
+  if (!allowed) {
+    throw new Error('Request method not allowed, use getIdentities or signAndPost');
+  }
+}
+
+export class RequestHandler {
+  private static instance: Request[];
```
webmaster128: Why not just `private static readonly instance: Request[] = [];`? Then you don't need the create() method at all. Additionally, you can drop all those checks: `if (!RequestHandler.instance) {`

packages/sanes-chrome-extension/src/extension/background/createPersona/requestAppUpdater.ts
```diff
@@ -0,0 +1,12 @@
+/*global chrome*/
+import { ProcessedTx } from '../../../logic/persona';
+import { MessageToForeground, MessageToForegroundAction } from '../messages';
+
+export function transactionsUpdater(transactions: ReadonlyArray<ProcessedTx>): void {
```
webmaster128: 👍

packages/sanes-chrome-extension/src/extension/background/createPersona/requestExtensionBadge.ts
```diff
@@ -0,0 +1,17 @@
+import { RequestHandler } from './requestHandler';
+
+/*global chrome*/
+export function updateExtensionBadge(): void {
```
webmaster128: I think it would be nice if this just takes a `count: number` argument instead of letting `updateExtensionBadge` depend on `RequestHandler`.

packages/sanes-chrome-extension/src/extension/background/createPersona/requestExtensionBadge.ts
```diff
@@ -0,0 +1,17 @@
+import { RequestHandler } from './requestHandler';
+
+/*global chrome*/
```
webmaster128: Move comment to top of file?

packages/sanes-chrome-extension/src/extension/backgroundscript.ts
```diff
@@ -1,38 +1,71 @@
...
+        .catch(console.error);
+      break;
+    case MessageToBackgroundAction.CreatePersona:
+      createPersona()
+        .then(sendResponse)
+        .catch(console.error);
+      break;
+    default:
+      return 'Unknown action';
+  }
```
webmaster128: Why was handleInternalMessage and handleExternalMessage inlined? I mean both ways are fine but I just want to understand the reason for the change

apanizo: What do you mean by inlined? The main reason to have moved the listeners to `/background/handlers/externalHandler.ts` and `/background/handlers/internalHandler.ts` is because you are not able to import a concrete function (which does not have any chrome reference) if in the same file you have chrome references. And, therefore, JEST was complaining.

In the last commit the `backgroundscript.ts` looks like:
```
/*global chrome*/
import { handleExternalMessage } from './background/handlers/externalHandler';
import { internalHandler } from './background/handlers/internalHandler';

/**
 * Listener for dispatching extension requests
 */
chrome.runtime.onMessage.addListener(internalHandler);

/**
 * Listener for dispatching website requests towards the extension
 */
chrome.runtime.onMessageExternal.addListener(handleExternalMessage);
```

I know maybe is not ideal, but was the best solution I found for unit testing without too much mocking the external messages.

packages/sanes-chrome-extension/src/extension/background/createPersona/requestHandler.ts
apanizo: The main reason to do not do that is because I can control in a natural way when the list is cleared, and reset it when we support switch of multiple personas.

I am not a big fan of singleton static classes to be initialised automatically. 

Do not worry about possible side effect problems because we have a test checking what happens when the RequestHandler has not `loaded`.

packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestSenderWhitelist.ts
```diff
@@ -0,0 +1,64 @@
+const ALLOWED_URLS_KEY = 'auk';
+const BLOCKED_URLS_KEY = 'buk';
+const SEPARATOR = ',';
+
+export class SenderWhitelist {
```
albertandrejev: We have `ShareIdentity` block list and `Transaction Request` block list. Should we use one list for all purposes?

apanizo: As we have the code today we can get rid of the allowed list. We do not process any request automatically. Coded in this way because I had in mind the recurring transactions which at the end we are not going to implement.

Good catch.

packages/sanes-chrome-extension/src/extension/backgroundscript.ts
webmaster128: In software development, "inline" usually means to take the code of a function's body in put it in the caller code instead of calling the function. One example of _inlining add_ is

Old code

```js
function add(a, b) { return a+b }

const x = add(1, 2)
const y = add(3, 4)
```

New code

```js
const x = 1 + 2
const y = 3 + 4
```

Anyways, in the latest code version `internalHandler` and `handleExternalMessage` are not inlined anymore. They were at the point when I commented, so this is resolved.

packages/sanes-chrome-extension/src/extension/background/createPersona/requestHandler.ts
webmaster128: Hmm, this means we combine the non-explicitness of global state with singletons and the hardness of manual initialization. Worst of both worlds ;) Anyway, fine for me as it will do the job.

> when we support switch of multiple personas.

As long as we don't support multiple Persona (which will be many months from now), there is no point at even thinking about that case. Our code will look entirely different at that point.

Closing this for now

packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestSenderWhitelist.ts
```diff
@@ -0,0 +1,34 @@
+const BLOCKED_URLS_KEY = 'buk';
+const SEPARATOR = ',';
```
webmaster128: "," is a valid character in URLs. But I think the blacklisting/whitelisting should happen based on domains anyway, so the separator is good. Postponed: https://github.com/iov-one/ponferrada/issues/187

PR #183, opened 2019-05-15
Fix – Update Lisk start scripts

webmaster128:
The important line here is

```
$gnutimeout" 50 bash -o pipefail -c 'until [ "$(HEIGHT=$(curl  -sS --fail http://localhost:4000/api/node/status | jq -e .data.height) && echo $HEIGHT || echo 0)" -ge 1 ]; do sleep 2; done'
```

Now the Lisk node is only considered ready when
* it responds
* status code is not HTTP 5xx (Server errors)
* height is >= 1, i.e. genesis is applied

Additionally load_faucet.sh is merged into init.sh, to allow using the same scripts in IOV-Core and ponferrada


apanizo:
@webmaster128 good job.

Remember, branch name: `feature/lisk-setup` 😉 
Also, remember to assign the PR to browser extension project for generating a ticket automatically and keep good track of what we do (necessary for generating reports).

webmaster128:
@apanizo yes, I noticed only when the PR was already created. Will be more careefull. Thanks for the reminder!

PR #184, opened 2019-05-15
Feature – Create new type PersonaAccount, which is JSON compatible

webmaster128:
Before this change we passed (unused) `PublicIdentity`s over JSON, which does not work

### DIFF
--- a/packages/sanes-chrome-extension/src/logic/persona/accountManager.ts
+++ b/packages/sanes-chrome-extension/src/logic/persona/accountManager.ts
@@ -4,7 +4,7 @@ import { Slip10RawIndex } from '@iov/crypto';
 import { ReadonlyWallet } from '@iov/keycontrol/types/wallet';
 
 export interface AccountInfo {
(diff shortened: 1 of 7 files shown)


packages/sanes-chrome-extension/src/logic/persona/accountManager.ts
```diff
@@ -1,10 +1,10 @@
-import { Algorithm, ChainId, publicIdentityEquals, PublicIdentity } from '@iov/bcp';
+import { Algorithm, ChainId, PublicIdentity, publicIdentityEquals } from '@iov/bcp';
 import { UserProfile, WalletId } from '@iov/core';
 import { Slip10RawIndex } from '@iov/crypto';
 import { ReadonlyWallet } from '@iov/keycontrol/types/wallet';
 
 export interface AccountInfo {
```
apanizo: As far as I see this interface is only used here, so you can remove the `export`, right?

PR #186, opened 2019-05-16
Feature – Modernize Travis

webmaster128:
let's see if that passes ...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,10 @@ node_js:
   - "10"
 sudo: required
 
+services:
+  - docker
+  - xvfb
+
 cache:
   directories:
     - node_modules
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ before_install:
   # Enable user namespace cloning
   - "sudo sysctl -w kernel.unprivileged_userns_clone=1"
 
+# Override default install step (travis.sh takes care of that)
+install: true
+
 before_script:
   # Make variables and function from Travis available in our script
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,10 +1,5 @@
 language: node_js
-dist: trusty
-addons:
-  apt:
-    packages:
-      # This is required to run new chrome on old trusty
-      - libnss3
```
apanizo: Why do you remove libnss3? It is needed for puppeteer... 
https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-on-travis-ci

webmaster128: I just want to try if installing the package is still necessary on Ubuntu xenial, which is worth trying when needed "on old trusty", which we don't use anymore. When CI greed -> good, when CI red -> revert the commit.

webmaster128: All e2e test running without the manuall installation of libnss3 :)

scripts/travis.sh
```diff
@@ -64,7 +64,7 @@ fold_end
 # Test
 #
 fold_start "yarn-tests"
-yarn test
+xvfb-run --auto-servernum yarn test
```
apanizo: What is this?

webmaster128: It is an replacement for the previous

```
  # Launch XVFB
  - "export DISPLAY=:99.0"
  - "sh -e /etc/init.d/xvfb start"
```

So instead of telling all processes that _there use the DISPLAY=:99.0_, you now prepend `xvfb-run` to the processes that need to use a display.

PR #191, opened 2019-05-16
Feature – Add additional accounts

### LINKED ISSUE #105: Integration iov-core to allow multi account creation
**Description**
For finishing completely the account status view, we need to give support for multi-account creation inside persona.

webmaster128:
Closes #105

### DIFF
--- a/packages/sanes-chrome-extension/src/extension/background/actions/createAccount/index.ts
+++ b/packages/sanes-chrome-extension/src/extension/background/actions/createAccount/index.ts
@@ -1,0 +1,11 @@
+import { CreateAccountResponse } from '../../messages';
+import { getCreatedPersona } from '../createPersona';
+
+export async function createAccount(): Promise<CreateAccountResponse> {
+  const persona = getCreatedPersona();
+  await persona.createAccount();
+  const response: CreateAccountResponse = {
+    accounts: await persona.getAccounts(),
+  };
+  return response;
+}
--- a/packages/sanes-chrome-extension/src/logic/persona/persona.unit.spec.ts
+++ b/packages/sanes-chrome-extension/src/logic/persona/persona.unit.spec.ts
@@ -58,6 +58,28 @@ withChainsDescribe('Persona', () => {
     });
   });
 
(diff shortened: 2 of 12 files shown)


packages/sanes-chrome-extension/src/extension/background/actions/createAccount/index.ts
```diff
@@ -0,0 +1,11 @@
+import { CreateAccountResponse } from '../../messages';
+import { getCreatedPersona } from '../createPersona';
+
+export async function createAccount(): Promise<CreateAccountResponse> {
```
apanizo: I am going to propose, for reducing complexity migrate all communication between the extension -> bacgkround script using the same approach I implemented for getting the requests' queue. As you have already identified, `getBackgroundPage`. 

In this way, we save some back and forth of messages. But we can do it easily in a separate PR. Not urgent now.

packages/sanes-chrome-extension/src/logic/persona/persona.unit.spec.ts
```diff
@@ -58,6 +58,28 @@ withChainsDescribe('Persona', () => {
     });
   });
 
+  describe('createAccount', () => {
+    it('can create an account', async () => {
+      const persona = await Persona.create();
+
+      {
```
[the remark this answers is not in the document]
webmaster128: It creates a new scope, such that scoped variables (`let`, `const`) are only available in this scope. E.g.

```js
const a = 1;

// a exists here

{
  const b = 2;
  // a exists here, b exists here
}

// a axists here, but no b anymore, since b went out of scope.
```

This is useful in tests when you want to reuse the same variable names multiple times. Here I have 2 `const accounts`.

packages/sanes-chrome-extension/src/extension/background/actions/createAccount/index.ts
webmaster128: I agree, let's hope this works nicely. This also removes the limitation of JSON compatible data

PR #192, opened 2019-05-17
Feature #190 route requests

### LINKED ISSUE #190: Finish routes/requests view connecting it with BS request queue
**Description**
This ticket is for creating the UI view for showing enqueued requests and send them to the proper route.

- [x] Created context' RequestProvider for updating the extension's view automatically
- [x] Created logic for exposing easily the queue to the app without a big back-forth message exchanging.
- [x] RequestHandler (which is the static class handling queue) created to an empty array when loading.
- [ ] Finish the route's request view

apanizo:
**Description**
This PR creates the UI for showing the enqueued requests which need to be accepted or rejected. For being able to create this view, this PR also has the logic which enables the extension to get enqueued requests from the BS. Also, it has the logic just in case the extension is opened and it receives another request the view is updated automatically.

It closes #190, you can see it in action [here](https://drive.google.com/open?id=1_iZET76IT8wwoLWgTjrPRsqmbaP1n1pX)

**SCREENSHOT**
Screenshot 2019-05-20 14 18 24

**DONE**
- Created context' RequestProvider for updating the extension's view automatically
- Created logic for exposing easily the queue to the app without a big back-forth message exchanging.
- RequestHandler (which is the static class handling queue) created to an empty array when loading.
- Finish the route's request view

### DIFF
--- a/packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
+++ b/packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
@@ -48,9 +48,9 @@ export const PersonaProvider = ({ children, persona }: Props): JSX.Element => {
     console.log('PersonaProvider registering listener');
     chrome.runtime.onMessage.addListener((msg, sender, _sendResponse) => {
       const sameTarget = sender.id === chrome.runtime.id;
-      const messageToForeground = isMessageToForeground(msg, MessageToForegroundAction.TransactionsChanges);
-
-      if (!sameTarget || !messageToForeground) {
+      const msgToForeground = isMessageToForeground(msg);
+      const msgToPersonaProvider = msg.action === MessageToForegroundAction.TransactionsChanges;
+      if (!sameTarget || !msgToForeground || !msgToPersonaProvider) {
         // Only handle messages from background script
         return;
       }
--- a/packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
+++ b/packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
@@ -46,18 +46,21 @@ export const PersonaProvider = ({ children, persona }: Props): JSX.Element => {
     }
 
     console.log('PersonaProvider registering listener');
-    chrome.runtime.onMessage.addListener((message, sender, _sendResponse) => {
-      if (sender.id !== chrome.runtime.id || !isMessageToForeground(message)) {
+    chrome.runtime.onMessage.addListener((msg, sender, _sendResponse) => {
+      const sameTarget = sender.id === chrome.runtime.id;
+      const messageToForeground = isMessageToForeground(msg, MessageToForegroundAction.TransactionsChanges);
+
+      if (!sameTarget || !messageToForeground) {
         // Only handle messages from background script
         return;
       }
 
-      switch (message.action) {
+      switch (msg.action) {
         case MessageToForegroundAction.TransactionsChanges:
-          if (!Array.isArray(message.data)) {
+          if (!Array.isArray(msg.data)) {
             throw new Error('Data must be an array');
           }
-          setTxs(message.data);
+          setTxs(msg.data);
           break;
         default:
           throw new Error('Unknown action');
--- a/packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
+++ b/packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
@@ -6,6 +6,7 @@ import {
   MessageToForegroundAction,
 } from '../extension/background/messages';
 import { PersonaAcccount, ProcessedTx } from '../logic/persona';
+import { extensionContext } from '../utils/chrome';
 
 /** Only the fields that are set will be updated */
 export interface PersonaContextUpdateData {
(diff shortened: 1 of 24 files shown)


packages/sanes-chrome-extension/src/index.tsx
```diff
@@ -33,7 +48,9 @@ const render = (Component: React.ComponentType, persona: GetPersonaResponse): vo
 };
 
 sendGetPersonaMessage().then(persona => {
-  render(Route, persona);
+  const extensionWindow = chrome.extension.getBackgroundPage() as IovWindowExtension;
```
webmaster128: 👍 Very interesting that you can access the BS from the UI, did not know that

packages/sanes-chrome-extension/src/extension/background/messages/index.ts
```diff
@@ -98,9 +101,12 @@ export interface MessageToForeground {
   readonly data?: any;
 }
 
-export function isMessageToForeground(data: unknown): data is MessageToForeground {
+export function isMessageToForeground(data: unknown, expectedAction: string): data is MessageToForeground {
```
webmaster128: Please don't extend this type guard function. This should only check the type at runtime to allow type narrowing. This is the reason for the special return type `data is MessageToForeground`. See "User Defined Type Guards" in https://basarat.gitbooks.io/typescript/docs/types/typeGuard.html#user-defined-type-guards

What you could do instead is

```
const transactionsChangesMessage = isMessageToForeground(msg) && msg.action === MessageToForegroundAction.TransactionsChanges);
                                                                 ^^^ here TS knows what type msg is
```

packages/sanes-chrome-extension/src/extension/background/messages/index.ts
```diff
@@ -81,11 +81,14 @@ export async function sendGetPersonaMessage(): Promise<GetPersonaResponse> {
...
 }
 
+export type GetRequestResponse = ReadonlyArray<Request>;
+
 /**
  * String values will be used in the JSON message objects
  */
 export enum MessageToForegroundAction {
   TransactionsChanges = 'transaction_changed',
+  RequestChanges = 'request_changed',
```
webmaster128: Ahh, sorry for my stupid naming above. It should be "transactions changed" for both the symbol and the string, to say: _here is a signal telling you that something happened in the past, feel free to react to that_. I'll update the existing code.

packages/sanes-chrome-extension/src/extension/background/messages/index.ts
```diff
@@ -81,11 +81,14 @@ export async function sendGetPersonaMessage(): Promise<GetPersonaResponse> {
   });
 }
 
+export type GetRequestResponse = ReadonlyArray<Request>;
```
webmaster128: I think this symbol can be removed. It belongs to

```
getQueuedRequests(): ReadonlyArray<Request>
```

not here and for `ReadonlyArray<Request>` or `readonly Request[]` (new notation), we probably don't need an alias

packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
```diff
@@ -40,24 +41,26 @@ export const PersonaProvider = ({ children, persona }: Props): JSX.Element => {
...
     }
 
     console.log('PersonaProvider registering listener');
-    chrome.runtime.onMessage.addListener((message, sender, _sendResponse) => {
-      if (sender.id !== chrome.runtime.id || !isMessageToForeground(message)) {
+    chrome.runtime.onMessage.addListener((msg, sender, _sendResponse) => {
+      const sameTarget = sender.id === chrome.runtime.id;
+      const msgToForeground = isMessageToForeground(msg);
+      const msgToPersonaProvider = msg.action === MessageToForegroundAction.TransactionsChanges;
+      if (!sameTarget || !msgToForeground || !msgToPersonaProvider) {
```
albertandrejev: Possible code duplication

packages/sanes-chrome-extension/src/context/RequestProvider.tsx
```diff
@@ -0,0 +1,57 @@
...
+  React.useEffect(() => {
+    if (!extensionContext()) {
+      return;
+    }
+
+    chrome.runtime.onMessage.addListener((message, sender, _sendResponse) => {
+      const sameTarget = sender.id === chrome.runtime.id;
+      const msgToForeground = isMessageToForeground(message);
+      const msgToRequestProvider = message.action === MessageToForegroundAction.RequestChanges;
+      if (!sameTarget || !msgToForeground || !msgToRequestProvider) {
```
albertandrejev: Possible code duplication

packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
webmaster128: I think the question is: Should both providers listen to background messages individually? Or do we want one place that listens to messages from background and distributes them to the providers. Both ways are fine I think

apanizo:
@webmaster128 @albertandrejev will move the conversation regarding event listeners duplication into a different ticket. Will come back to you soon.

PR #199, opened 2019-05-20
Feature - Redirect requests to their views

apanizo:
**Description**
With this PR when a request is enqueued the UI knows the destination route of each request.

Screenshot 2019-05-20 23 28 29

**DONE**
- Added counter for identifying correctly requests in UI (unfortunately requests cannot be passed in the location' state
- Implemented logic handling of requests in share identify and tx-request views
- Added type to Request interface for dispatching them to their correct view.
- Added firstRequest to RequestContext

### DIFF
--- a/packages/sanes-chrome-extension/src/routes/requests/components/RequestList.tsx
+++ b/packages/sanes-chrome-extension/src/routes/requests/components/RequestList.tsx
@@ -38,7 +38,7 @@ function buildClickHandlerFrom(request: Request): () => void {
     history.push({
       pathname,
       state: {
-        [REQUEST_FIELD]: request,
+        [REQUEST_FIELD]: request.id,
       },
     });
 }
--- a/packages/sanes-chrome-extension/src/routes/requests/components/RequestList.tsx
+++ b/packages/sanes-chrome-extension/src/routes/requests/components/RequestList.tsx
@@ -61,6 +61,7 @@ const RequestList = ({ requests }: Props): JSX.Element => {
           const first = index === 0;
           const firstElemClass = first ? classes.first : undefined;
           const onRequestClick = buildClickHandlerFrom(req);
+          const text = req.type === 'getIdentities' ? 'Get Identities Request' : 'Sign Request';
 
           return (
             <React.Fragment>
@@ -71,7 +72,7 @@ const RequestList = ({ requests }: Props): JSX.Element => {
                 onClick={first ? onRequestClick : undefined}
               >
                 <ListItemText
-                  primary="Type of request"
+                  primary={text}
                   secondary={req.reason}
                   secondaryTypographyProps={secondaryProps}
                 />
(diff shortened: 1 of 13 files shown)


apanizo:
I will fix DOM test tomorrow, now you can give feedback, the UI how I have isolated and checked txs when dispatching them...

Thanks!!

packages/sanes-chrome-extension/src/routes/share-identity/index.tsx
```diff
@@ -1,19 +1,32 @@
...
 import RejectRequest from './components/RejectRequest';
 import ShowRequest from './components/ShowRequest';
 
-const ShareIdentity = (): JSX.Element => {
+const ShareIdentity = ({ location }: RouteComponentProps): JSX.Element => {
   const [action, setAction] = React.useState<'show' | 'reject'>('show');
+  const toast = React.useContext(ToastContext);
+  const requestContext = React.useContext(RequestContext);
+
+  const request = requestContext.firstRequest;
```
webmaster128: Suggestion: you can write `const request = requestContext.requests.find(() => true);`, which looks a bit odd but gives you the first element if there is one. This allows you to avoid the duplicated data.

packages/sanes-chrome-extension/src/routes/tx-request/index.tsx
```diff
@@ -1,19 +1,32 @@
...
 import RejectRequest from './components/RejectRequest';
 import ShowRequest from './components/ShowRequest';
 
-const TxRequest = (): JSX.Element => {
+const TxRequest = ({ location }: RouteComponentProps): JSX.Element => {
   const [action, setAction] = React.useState<'show' | 'reject'>('show');
+  const toast = React.useContext(ToastContext);
+  const requestContext = React.useContext(RequestContext);
+
+  const request = requestContext.firstRequest;
```
webmaster128: same here

packages/sanes-chrome-extension/src/routes/requests/components/RequestList.tsx
```diff
@@ -43,12 +60,19 @@ const RequestList = ({ requests }: Props): JSX.Element => {
...
-              <ListItem key={`${index}`} className={firstElemClass} disabled={!first}>
+              <ListItem
+                key={`${index}`}
+                className={firstElemClass}
+                disabled={!first}
+                onClick={first ? onRequestClick : undefined}
+              >
                 <ListItemText
-                  primary="Type of request"
+                  primary={text}
```
albertandrejev: I think we should show who is asking for the permissions (resource name). Because there is no information about source of the request. I understand that customer will see this info on next screen, but I think it is better for reviewing purpose.

apanizo: Sure, I will keep in mind, we can discuss it when #201 is done and surrounding tests we need to implement.

packages/sanes-chrome-extension/src/routes/share-identity/index.tsx
apanizo: I was going to modify thinking here I was using `requests` and `firstRequest`, but I realised only the `firstRequest` is used. I prefer to keep it in this way because it explains better internal requirements and it is encapsulated correctly.

webmaster128:
Merge?

PR #205, opened 2019-05-21
Feature #201 - Show real data tx-request and share-identity requests 

### LINKED ISSUE #201: Show real data tx-request and share-identity requests
**Description**
Right now tx-request and share-identity views use mocked data, once #199 is merged we can update those views with real data

- [ ] Update views with real data

albertandrejev:
**Description**
This PR will show real data from requests to appropriate views.
It will close #201 

**Screens**
Transaction request
tx-req-show
Transaction reject
tx-req-reject

Share identity
share-identity-main
Share identity reject
share-identity-reject

**Developer notes**
This PR was not fully completed because of lack of data in request object

### DIFF
--- a/packages/sanes-chrome-extension/src/routes/share-identity/index.tsx
+++ b/packages/sanes-chrome-extension/src/routes/share-identity/index.tsx
@@ -4,7 +4,7 @@ import { RouteComponentProps, withRouter } from 'react-router';
 import { RequestContext } from '../../context/RequestProvider';
 import { history } from '../../store/reducers';
 import { REQUEST_ROUTE } from '../paths';
-import { checkRequest } from '../requests';
+import { validRequest } from '../requests';
 import RejectRequest from './components/RejectRequest';
 import ShowRequest from './components/ShowRequest';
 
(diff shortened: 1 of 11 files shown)


apanizo:
@albertandrejev I have read your note of "This PR is not completed because of lack of data in request object", but can you show (including screenshots) how they look like right now?

packages/sanes-chrome-extension/src/routes/share-identity/index.tsx
```diff
@@ -31,8 +31,14 @@ const ShareIdentity = ({ location }: RouteComponentProps): JSX.Element => {
 
   return (
     <React.Fragment>
-      {action === 'show' && <ShowRequest onAcceptRequest={onAcceptRequest} showRejectView={showRejectView} />}
-      {action === 'reject' && <RejectRequest onBack={showRequestView} onRejectRequest={onRejectRequest} />}
+      {action === 'show' && (
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
apanizo: For avoid the constant eslint disable check of `request!` I suggest to refactor code and do:
```
const request = requestContext.firstRequest;
const valid = validRequest(request, location, toast) <-- from checkRequest to validRequest
if (!valid) {
  return;
}
```

packages/sanes-chrome-extension/src/routes/tx-request/index.tsx
```diff
@@ -31,8 +31,14 @@ const TxRequest = ({ location }: RouteComponentProps): JSX.Element => {
 
   return (
     <React.Fragment>
-      {action === 'show' && <ShowRequest onAcceptRequest={onAcceptRequest} showRejectView={showRejectView} />}
-      {action === 'reject' && <RejectRequest onBack={showRequestView} onRejectRequest={onRejectRequest} />}
+      {action === 'show' && (
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
apanizo: Same here

packages/sanes-chrome-extension/src/routes/share-identity/index.tsx
albertandrejev: Yes, that's was one of my first thoughts. But after this check I will be required to assign `request` variable to another non-undefined var.
like: 
```
// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
const requestValid = request!
```
And only after this I can use it in prop without any comment :(

PR #210, opened 2019-05-22
Feature #207 Reject automatically enqueued txs

### LINKED ISSUE #207: Reject automatically requests if one blocked permanently
**Description**
Having queue of requests if, one request is permanently rejected, we should check the queue and reject automatically those ones which have the same sender.

apanizo:
**Description**
This PR removes from the background script's request queue all requests belonging to a permanently blocked sender.

It closes #207 

See this [feature in action](https://drive.google.com/open?id=1Z5WypTDvHHp56GnQDLOuPFR94NhgYKfE)

**Note for developers**
- Found a bug doing the conversion of permanent value from the form. Take a look!

### DIFF
--- a/packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestHandler.ts
+++ b/packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestHandler.ts
@@ -49,4 +49,22 @@ export class RequestHandler {
 
     return size;
   }
+
+  public static purge(senderUrl: string): void {
+    const initialSize = RequestHandler.instance.length;
+    for (let i = 0; i < initialSize; i++) {
+      const req = RequestHandler.instance[i];
+      if (req.sender !== senderUrl) {
+        continue;
+      }
+
+      // Note here we are mutating the array placing in first positions request to be rejected
+      RequestHandler.instance.splice(i, 1);
+      RequestHandler.instance.splice(0, 0, req);
+    }
+
+    // Note here we only get references
+    const reqToCancel = RequestHandler.instance.filter(req => req.sender === senderUrl);
+    reqToCancel.forEach(req => req.reject(false));
+  }
 }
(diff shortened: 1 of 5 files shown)


packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestHandler.ts
```diff
@@ -49,4 +49,22 @@ export class RequestHandler {
...
+    const initialSize = RequestHandler.instance.length;
+    for (let i = 0; i < initialSize; i++) {
+      const req = RequestHandler.instance[i];
+      if (req.sender !== senderUrl) {
+        continue;
+      }
+
+      // Note here we are mutating the array placing in first positions request to be rejected
+      RequestHandler.instance.splice(i, 1);
+      RequestHandler.instance.splice(0, 0, req);
```
webmaster128: Simpler: `RequestHandler.instance.unshift(req)`

packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestHandler.ts
```diff
@@ -49,4 +49,22 @@ export class RequestHandler {
...
+    }
```
webmaster128: Hmm, if reject() would not automatically solve the first element, we would not need this reordering. Anyways, this seems to be correct

packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestHandler.ts
apanizo: The thing is, when we reject the request, in the callback it calls to `solve()` which removes the first element.

I believe that was my first approach and it failed, anyway I will give another shoot while @albertandrejev reviews it. I have something in mind.

packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestHandler.ts
apanizo: Same as before, I will check again.

packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestHandler.ts
webmaster128: This particular comment is just about readability of the existing logic. `unshift` is like `push` but adds in the front instead of the back

PR #214, opened 2019-05-22
Feature #200 - Unit tests for RequestHandler

### LINKED ISSUE #200: Test request handler counter
We have added an internal id for assuring the request which is going to be accepted or rejected is the same of the selected in the list.

We should unit test all options for assuring the counter is implemented correctly, kind of black-box test, queuing several requests, dispatching them, then add more, and check counter should work as expected.

Also test given a queue of requests the first one to be dispatched is the 0, then the 1, and so on.

abefernan:
This PR adds unit testing to `RequestHandler` with the goal of:

- Testing the basic functionality of its public methods.
- Checking the reliable update of its private fields `instance` and `counter`.
- Ensure the interaction between the `counter` field and `add()`'s `id` assignment is correct.

It closes #200.


packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestHandler/index.unit.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    id: 2999,
+    type: 'signAndPost',
+    reason: 'Reason 2',
+    sender: 'Sender 2',
+    accept: () => {},
+    reject: () => {},
+  };
+
+  const REQUEST_THREE: Request = {
+    id: 3999,
```
apanizo: There is no need to declare `id`, it should be added automatically when `add`. In fact this is something we need to unit test.

` add | add | solve | solve | add | reject | add` <--- the last one should be 3 (counter starts on 0) `

packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestHandler/index.unit.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    accept: () => {},
+    reject: () => {},
+  };
+
+  const REQUEST_TWO: Request = {
+    id: 2999,
+    type: 'signAndPost',
+    reason: 'Reason 2',
+    sender: 'Sender 2',
+    accept: () => {},
```
apanizo: I would use `jest.fn()` more JEST friendly 😉

packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestHandler/index.unit.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    RequestHandler.load();
+    RequestHandler.add(REQUEST_ONE);
+    RequestHandler.add(REQUEST_TWO);
+    RequestHandler.add(REQUEST_THREE);
+  });
+
+  it('re-initializes itself', () => {
+    RequestHandler.load();
+
+    expect(RequestHandler['instance'].length).toBe(0);
```
apanizo: As I said in the review of the other PR, no more violating private accessors 👴 . You can use `RequestHandler.requests(). length`

packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestHandler/index.unit.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+    expect(RequestHandler['counter']).toBe(0);
+  });
+
+  it('updates counter when a new request is added', () => {
+    expect(RequestHandler['counter']).toBe(3);
+    RequestHandler.solved();
+    expect(RequestHandler['counter']).toBe(3);
+    RequestHandler.add(REQUEST_ONE);
+    expect(RequestHandler['counter']).toBe(4);
```
apanizo: The example I posted two comments before can be added here.

packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestHandler/index.unit.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  it('re-initializes itself', () => {
+    RequestHandler.load();
+
+    expect(RequestHandler['instance'].length).toBe(0);
+    expect(RequestHandler['counter']).toBe(0);
+  });
+
+  it('updates counter when a new request is added', () => {
+    expect(RequestHandler['counter']).toBe(3);
```
apanizo: In this case, you have to test the counter, I understand it, add a comment explaining it.

packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestHandler/index.unit.spec.ts
abefernan: I'm aware the `id`s are inserted by `add`, it is tested on the last test. I put those `id`s to comply with the Request interface so that the `add` method accepts this objects as arguments.

Is there any way to remove the id property from these objects and `add` still accepts them as arguments? Because as soon as I remove `id` it can no longer be a `Request` object.

apanizo: You can use: ```Omit<Request, 'id'>``` in TS 3.5 it will be added by default, right now just use  `import { Omit } from '@material-ui/core';`

packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestHandler/index.unit.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    expect(requests[0].reason).toBe(REQUEST_ONE.reason);
+    expect(requests[1].reason).toBe(REQUEST_TWO.reason);
+    expect(requests[2].reason).toBe(REQUEST_THREE.reason);
+
+    RequestHandler.solved();
+    RequestHandler.solved();
+    requests = RequestHandler.requests();
+
+    expect(requests.length).toBe(1);
+    expect(requests[0].reason).toBe(REQUEST_THREE.reason);
```
apanizo: You should use `next()`:
```
expect(RequestHandler.requests().length).toBe(1);
const request = RequestHandler.next();
expect(request.reason).toBe(REQUEST_THREE.reason);
```

PR #215, opened 2019-05-22
Fix - Simplifies access to blocked list in SenderWhitelist unit test

abefernan:
This PR refactors the unit testing suite for `SenderWhitelist` in order to:

- Make use of array access notation to simplify the access to `SenderWhitelist`'s blocked senders list.
- Removes no longer needed constants that are, additionaly, duplicated from `SenderWhitelist`.

### DIFF
--- a/packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestSenderWhitelist/index.unit.spec.ts
+++ b/packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestSenderWhitelist/index.unit.spec.ts
@@ -1,8 +1,6 @@
 import { SenderWhitelist } from './index';
 
 describe('SenderWhitelist', () => {
-  const BLOCKED_URLS_KEY = 'buk';
-  const SEPARATOR = ',';
   const SENDER_ONE = 'Example Sender 1';
   const SENDER_TWO = 'Example Sender 2';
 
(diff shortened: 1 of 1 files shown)


packages/sanes-chrome-extension/src/extension/background/actions/createPersona/requestSenderWhitelist/index.unit.spec.ts
```diff
@@ -19,9 +17,7 @@ describe('SenderWhitelist', () => {
     SenderWhitelist.block(SENDER_TWO);
 
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-    const numBlocked = localStorage.getItem(BLOCKED_URLS_KEY)!.split(SEPARATOR).length;
-
-    expect(numBlocked).toBe(2);
+    expect(SenderWhitelist['blocked']!.length).toBe(2);
```
apanizo: You can call me old, but accessing a private static field in this way (first of all, I did not know that was possible) makes me feel strange.

But I understand your point, and seeing this is a test I would add this comment:
```
// SenderWhitelist does not expose the array of blocked domains, so this is an exceptional
// use case which can not be used in regular code because it violates private modifier of blocked variable.
```

abefernan: Sure, I believe it would be better to have a getBlocked() method, but I didn't want to extend the API only for the sake of tests.

PR #218, opened 2019-05-22
Feature #204 - Bierzo wallet. Extension's badge text e2e tests.

### LINKED ISSUE #204: e2e test bierzo-wallet extension's badge icon
This test should create an e2e test checking if the user in bierzo wallet sends a couple of requests for getting identities, the badge text in the extension is updated correctly.

**NOTE**
For implementing this ticket we should assure chrome global is accessible from puppeteer.

albertandrejev:
**Description**
This PR will add E2E tests to Bierzo Wallet and will check extension's badge number incrementation in case of request added to the queue. It will close #204.

### DIFF
--- a/packages/bierzo-wallet/package.json
+++ b/packages/bierzo-wallet/package.json
@@ -16,9 +16,11 @@
     "@types/node": "11.12.2",
     "@types/react": "16.8.10",
     "@types/react-dom": "16.8.3",
+    "concurrently": "^4.1.0",
     "medulas-react-components": "^0.0.0",
     "normalize.css": "^8.0.1",
-    "react-scripts": "3.0.0"
+    "react-scripts": "3.0.0",
+    "serve": "^11.0.0"
   },
   "devDependencies": {
     "@types/chrome": "^0.0.82"
@@ -28,7 +30,8 @@
     "format": "prettier --write --loglevel warn './src/**/*.{ts,tsx,json,md,css}'",
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "CI=true react-scripts test '\\.dom\\.(spec|test)\\.[jt]s[x]?$' --env=jsdom"
+    "test": "CI=true react-scripts test '\\.dom\\.(spec|test)\\.[jt]s[x]?$' --env=jsdom",
+    "test-e2e": "CI=true concurrently -s \"first\" -k \"serve -s build -l 3000\" \"react-scripts test '\\.e2e\\.(spec|test)\\.[jt]s[x]?$' --env=jsdom\""
   },
   "browserslist": [
     ">0.2%",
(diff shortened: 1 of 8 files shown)


packages/bierzo-wallet/package.json
```diff
@@ -28,7 +30,9 @@
     "format": "prettier --write --loglevel warn './src/**/*.{ts,tsx,json,md,css}'",
     "start": "react-scripts start",
     "build": "react-scripts build",
-    "test": "CI=true react-scripts test '\\.dom\\.(spec|test)\\.[jt]s[x]?$' --env=jsdom"
+    "test-dom": "CI=true react-scripts test '\\.dom\\.(spec|test)\\.[jt]s[x]?$' --env=jsdom",
+    "test-e2e": "CI=true concurrently -s \"first\" -k \"serve -s build -l 3000\" \"react-scripts test '\\.e2e\\.(spec|test)\\.[jt]s[x]?$' --env=jsdom\"",
```
apanizo: What is `concurrently -s \"first\" -k \"serve -s build -l 3000\"`?

packages/bierzo-wallet/src/routes/welcome/index.e2e.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+    let browser: Browser;
+    let page: Page;
+    let backgroundPage: Page;
+    let extensionPage: Page;
+
+    beforeEach(async (): Promise<void> => {
+      browser = await launchBrowser();
+      page = await createPage(browser);
+      extensionPage = await createExtensionPage(browser);
+      backgroundPage = await getBackgroundPage(browser);
```
apanizo: I like the `backgroundPage` approach. The initial approach I had in mind was:
```
const badgeText = await page.evaluate(    <------ Note page here
    (): Promise<string | undefined> => {
        return new Promise(resolve => {
            chrome.browserAction.getBadgeText({}, text => resolve(text));
        });
   },
);
```

I thought having the page tab with the extension installed you already had access to `chrome.browserAction`.

But the `getBackgroundPage` is so cool. It is so damn cool that despite we can have other alternatives we are going to leave it in the repo. 

Just curiosity, did you try my approach?

albertandrejev: Yes I tried it, but you do not have access to `chome.browserAction` from it.

packages/bierzo-wallet/package.json
albertandrejev: full command is `concurrently -s "first" -k "serve -s build -l 3000" "react-scripts test '\\.e2e\\.(spec|test)\\.[jt]s[x]?$' --env=jsdom"
`concurrently` - run commands in parallel 
`-s "first" -k` - kill all threads after first finished thread and return command execution status code from first finished thread.
`serve -s build -l 3000` - serves web static content from `build` directory on port 3000
`react-scripts test '\\.e2e\\.(spec|test)\\.[jt]s[x]?$' --env=jsdom` - I think you know what it is for.

PR #219, opened 2019-05-23
Feature – Persona persistency

### LINKED ISSUE #177: Persistency! Store persona to database and load from database

webmaster128:
Closes #177


packages/sanes-chrome-extension/src/extension/background/handlers/externalHandler.ts
```diff
@@ -15,6 +16,7 @@ function resolveToError(
 }
 
 export function handleExternalMessage(
+  db: StringDb,
```
apanizo: Not needed, right?

webmaster128: oh, indeed, thanks. Will remove

webmaster128:
All updated, CI fail is because of failing master

PR #220, opened 2019-05-23
Fix – Use stronger linting settings

webmaster128:
Before, we used `--max-warnings 1`, which was a stupid mistake by me as it allowsed up to 1 warning per package.

Also added `eslint:recommended` as a baseline ruleset.


webmaster128:
> Do not know what exactly introduces `eslint:recommended` different from `@typescript-eslint/recommended`, but as long as we check more things, I am happy with it.

I did not look into every single rule either but basically eslint:recommended is a set of JavaScript rules and @typescript-eslint/recommended is what the typescript plugin recommends as TypeScript defaults.

PR #222, opened 2019-05-23
Feature #193 - DOM tests for Requests view

### LINKED ISSUE #193: DOM test for dispatch enqueued requests to their proper view
**Description**
We have our routes/request in #199 we are dispatching from request view to their corresponding view. Now we need to assure those redirections are done correctly.

- [x] Done in #199 Connect logic for each request to `share-identity` or `tx-request` respectively
- [ ] Add DOM test for request view to handle correctly each request to their view
- [x] Review externalHandler unit test for checking enqueues everything correctly (including type and counter.

abefernan:
**Description**
In the Requests view, the user can click on the first request to solve it. Depending on the type of the request, it will be redirected to a different rout. If the request is of type "getIdentities" it will be redirected to the Share Identity view, whereas if it is of type "signAndPost" it will be redirected to the TX Request view.

The tests that have been implemented aim to check that each type of request redirect to its corresponding view when clicked.

**Bug**
There is currently a bug that makes the test not pass if executed on the same run. That is to say, the tests pass if executed individually by setting `it.only()`.

It closes #193.

### DIFF
--- a/packages/sanes-chrome-extension/src/routes/requests/index.dom.spec.ts
+++ b/packages/sanes-chrome-extension/src/routes/requests/index.dom.spec.ts
@@ -1,6 +1,6 @@
 import TestUtils from 'react-dom/test-utils';
 import { Store } from 'redux';
-import { Request } from '../../extension/background/actions/createPersona/requestHandler';
+import { Request } from '../../extension/background/model/signingServer/requestQueueManager';
 import { aNewStore } from '../../store';
 import { RootState } from '../../store/reducers';
 import { whenOnNavigatedToRoute } from '../../utils/test/navigation';
--- a/packages/sanes-chrome-extension/src/routes/requests/index.dom.spec.ts
+++ b/packages/sanes-chrome-extension/src/routes/requests/index.dom.spec.ts
@@ -42,7 +42,7 @@ describe('DOM > Feature > Requests', () => {
     });
 
     await whenOnNavigatedToRoute(store, SHARE_IDENTITY);
-  });
+  }, 55000);
 
   it('redirects to the TX Request view when a Request of type "signAndPost" is clicked', async () => {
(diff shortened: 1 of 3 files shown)


packages/sanes-chrome-extension/src/routes/requests/index.dom.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+    const identityRequest = TestUtils.scryRenderedDOMComponentsWithTag(requestsDom, 'li')[0];
+
+    TestUtils.act(
+      (): void => {
+        TestUtils.Simulate.click(identityRequest);
+      },
+    );
+
+    await whenOnNavigatedToRoute(store, SHARE_IDENTITY);
+  });
```
apanizo: You need to destroy persona after each test.
```
const persona = getCreatedPersona();
persona.destroy();
```

Otherwise connections are open and probably make the other tests to fail.

abefernan: I have tried that but it does not return any Persona since it's not created anywhere.

packages/sanes-chrome-extension/src/routes/requests/index.dom.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  it('redirects to the TX Request view when a Request of type "signAndPost" is clicked', async () => {
+    requestsDom = await travelToRequests(store, [REQUEST_TWO, REQUEST_ONE]);
+    const txRequest = TestUtils.scryRenderedDOMComponentsWithTag(requestsDom, 'li')[0];
+
+    TestUtils.act(() => {
+      TestUtils.Simulate.click(txRequest);
+    });
+
+    await whenOnNavigatedToRoute(store, TX_REQUEST);
+  });
```
albertandrejev: Please update this test by setting test timeout value and it will pass.

PR #226, opened 2019-05-26
Feature #111 - DOM tests for Account Status view

### LINKED ISSUE #111: Create DOM test in account status view
The DOM test should cover: Create a persona, then navigate to account status, mock a send transaction.

This test should check: all UI elements are correct (tx is shown in the box, the hamburguer menu is there, select has one account...)

**Note**
This is not blocked because we are mocking the `send transaction` step.

abefernan:
**Description**
This PR aims to add DOM testing to the Account Status view in order to test that the basic UI elements are present and that the transaction box works as expected.

- Test that the hamburger button is present.
- Test that the Account dropdown has an account.
- Test that the transaction box has the expected mocked transaction.

It closes #111.

### DIFF
--- a/packages/sanes-chrome-extension/src/routes/account/index.dom.spec.ts
+++ b/packages/sanes-chrome-extension/src/routes/account/index.dom.spec.ts
@@ -1,18 +1,13 @@
 import { TokenTicker } from '@iov/bcp';
 import TestUtils from 'react-dom/test-utils';
 import { Store } from 'redux';
-import * as messages from '../../extension/background/messages';
-import { CreatePersonaResponse, GetPersonaResponse } from '../../extension/background/messages';
+import { GetPersonaResponse } from '../../extension/background/messages';
 import { PersonaAcccount, ProcessedTx } from '../../logic/persona';
 import { aNewStore } from '../../store';
 import { RootState } from '../../store/reducers';
-import { whenOnNavigatedToRoute } from '../../utils/test/navigation';
-import { withChainsDescribe } from '../../utils/test/testExecutor';
-import { ACCOUNT_STATUS_ROUTE } from '../paths';
-import { processSignUp } from '../signup/test/travelToSignup';
 import { travelToAccount } from './test/travelToAccount';
 
-withChainsDescribe('DOM > Feature > Account Status', () => {
+describe('DOM > Feature > Account Status', () => {
   const accountMock: PersonaAcccount = { label: 'Account 0' };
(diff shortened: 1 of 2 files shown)


packages/sanes-chrome-extension/src/routes/account/index.dom.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+  beforeEach(async () => {
+    store = aNewStore();
+
+    jest.spyOn(messages, 'sendCreatePersonaMessage').mockResolvedValue(createPersonaResponse);
+
+    await processSignUp(store);
+    await whenOnNavigatedToRoute(store, ACCOUNT_STATUS_ROUTE);
+
+    accountStatusDom = await travelToAccount(store, getPersonaResponse);
+    await whenOnNavigatedToRoute(store, ACCOUNT_STATUS_ROUTE);
```
albertandrejev: To be honest I am not sure that you need to call `processSignUp` and check for `ACCOUNT_STATUS_ROUTE` because you have called `travelToAccount(store, getPersonaResponse)`, but I am not 100% sure about this, maybe I am wrong. But I am sure that you need to check for `ACCOUNT_STATUS_ROUTE` after calling `travelToAccount` because this method (`travelToAccount`) already contains this checking.

albertandrejev: But in case if I am wrong about `processSignUp` you do not need to check for `await whenOnNavigatedToRoute(store, ACCOUNT_STATUS_ROUTE);` on the next line, because this is redundant in this test and not required, you are not testing sign up process in this test case.

packages/sanes-chrome-extension/src/routes/account/index.dom.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+  let store: Store<RootState>;
+  let accountStatusDom: React.Component;
+
+  beforeEach(async () => {
+    store = aNewStore();
+
+    jest.spyOn(messages, 'sendCreatePersonaMessage').mockResolvedValue(createPersonaResponse);
+
+    await processSignUp(store);
+    await whenOnNavigatedToRoute(store, ACCOUNT_STATUS_ROUTE);
```
albertandrejev: I think you do not need to check for `await whenOnNavigatedToRoute(store, ACCOUNT_STATUS_ROUTE);` , because this is redundant in this test and not required, you are not testing sign up process in this test case.

PR #229, opened 2019-05-27
Fix #223 - Update extension badge after brower restart

### LINKED ISSUE #223: Badge text should be cleared
At the moment, If you close the browser having requests in queue we do not persist them to being dispatched when the browser is opened again. In fact, the queue is reseted.

So, in the background script, which is a file only loaded once, we should clear the icon status like if no requests are handled.

albertandrejev:
**Description**
This PR will fix extension badge appearance after browser restart. Also it will close #223.

### DIFF
--- a/packages/sanes-chrome-extension/src/extension/backgroundscript.ts
+++ b/packages/sanes-chrome-extension/src/extension/backgroundscript.ts
@@ -1,9 +1,13 @@
 /*global chrome*/
 import { generateErrorResponse } from './background/errorResponseGenerator';
 import Backgroundscript from './background/model/backgroundscript';
+import { updateExtensionBadge } from './background/updaters/extensionBadgeUpdater';
 
 const backgroundScript = new Backgroundscript();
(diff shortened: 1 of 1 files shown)


packages/sanes-chrome-extension/src/extension/backgroundscript.ts
```diff
@@ -3,6 +3,10 @@ import { createBrowserDb } from '../logic/db';
 import { Request, RequestHandler } from './background/actions/createPersona/requestHandler';
 import { handleExternalMessage } from './background/handlers/externalHandler';
 import { internalHandler } from './background/handlers/internalHandler';
+import { updateExtensionBadge } from './background/actions/createPersona/requestExtensionBadge';
+
+//Reset extension badge to clear one after browser start because of queue reset
+updateExtensionBadge(0);
```
apanizo: I do not really understand why the CI fails, let's wait for #230 to be merged, it should work fine.

About the solution implemented in this PR LGTM.

Issue #89, opened 2019-04-15
Add login logic


apanizo:
Once we support db storage of user profiles, we need to add the login logic for unlocking stored db and being able to recreate the stored persona correctly.

- [x] If profile is stored the app should redirect to login view for introducing password.

~Blocked by #177~

PR #230, opened 2019-05-28
Feature - Refactor background logic

apanizo:
**Description**
This PR refactors the whole **backgroundscript** logic. The main aim for this refactor was was the wrong approach we did create all logic close to the extension instead of the background script. During lasts weeks on every feature we have been adding, the architecture of the extension was more complex and more difficult to understand.

Not now:
- We have created a background entity model having the DB, the persona and the signingServer.
- The signing server, is another entitiy, which contains the request queue, sender whitelist (for knowing when a sender is blocked). It also dispatches requests from external websites.
- The background script exposes in `getBackgroundPage` all internal functionality needed in the extension.
- Persona entity, now, is just a persona. Internally it has all low-level blockchain features offered by iov-core, and exposes all data necessary for the extension and is related to its own domain. Since the object is really complex I have created a PersonaBuilder for helping to create stuff.
- Now all complexity is more balanced, we have 3 main entities (Persona, Backgroundscript and SigningServer) instead of only one (Persona) depending everything on one entity. SigningServer is focused on attending stuff happening outside of the extension. Persona does the same, but internally speaking, and Backgroundscript handles the interaction between both of them.
- We are more flexible for the future, adding multi-persona support can be done refactored the backgroundscript entity in an easier way.

**Notes:**
- The old `src/logic` folder is going to be refactored & finally removed when #89, so bear a bit with me 😉 
- For correct review I recommend just `git pull` this branch start from `extension/bacgkroundscript.ts` and pull the rope from it. 

IOV ❤️ refactoring


packages/sanes-chrome-extension/src/extension/background/model/accountManager/index.ts
```diff
@@ -0,0 +1,124 @@
...
+      });
+
+      return {
+        index: accountIndex,
+        identities: identities,
+      };
+    });
+  }
+
+  /** This must be called with the next unused derication index. Gaps are not supported. */
```
albertandrejev: Typo in comment

ethanfrey:
It looks like a good approach to keep all logic in one central place. And I guess the background script is that.

Hard to review as so many moved files are marked "deleted" and "new", not "moved".

I also wanted to understand how this fit together (bapp, signing server, background script, extension UI, bcp-connections, keystore, other items?). Like what goes where and what talks to what.

I am sure you all have this in your head, but I think an hour or two to sketch out the rough architecture in the README would be very useful to help someone new (me) look at the code.

packages/sanes-chrome-extension/src/extension/background/model/signingServer/index.ts
```diff
@@ -0,0 +1,127 @@
...
+      true,
+      false,
+    );
+  };
+
+  public getPendingRequests(): Request[] {
+    return this.requestHandler.requests();
+  }
+
+  public start(core: SigningServerCore): void {
```
webmaster128: This creates a circular dependency: Persona depends on SigningServer and SigningServer depends on Persona, which makes the flow very very hard to follow.

packages/sanes-chrome-extension/src/extension/background/model/signingServer/index.unit.spec.ts
```diff
@@ -25,27 +23,32 @@ const buildGetIdentitiesRequest = (method: string, customMessage?: string): Json
...
   });
   afterAll(() => {
     jest.spyOn(txsUpdater, 'transactionsUpdater').mockReset();
-    jest.spyOn(txsUpdater, 'requestUpdater').mockReset();
+    jest.spyOn(txsUpdater, 'updateRequestProvider').mockReset();
   });
 
   function checkNextRequest(request: JsonRpcRequest, sender: string): void {
-    const req = RequestHandler.next();
+    const req = signingServer['requestHandler'].next();
```
webmaster128: didn't you just tell Abel to not access private members? Usually accessing private data should not be necessary to test the bahaviour of a class

packages/sanes-chrome-extension/src/extension/background/model/signingServer/requestQueueManager/index.ts
```diff
@@ -39,62 +48,62 @@ export interface Request {
   readonly reject: (permanently: boolean) => void;
 }
 
-export class RequestHandler {
-  private static instance: Request[] = [];
-  private static counter = 0;
+export class RequestQueueManager {
+  private instance: Request[] = [];
```
webmaster128: 👍 good to get rid of static state

packages/sanes-chrome-extension/src/utils/chrome/index.ts
```diff
@@ -1,4 +1,47 @@
...
+
+  return response;
+}
+
+export function getQueuedRequests(): ReadonlyArray<Request> {
+  const extensionWindow = chrome.extension.getBackgroundPage() as IovWindowExtension;
+  const response = extensionWindow.getQueuedRequests();
+
+  return response;
+}
```
webmaster128: 👍

packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
```diff
@@ -54,6 +54,8 @@ export const PersonaProvider = ({ children, persona }: Props): JSX.Element => {
         return;
       }
 
+      // TODO Refactor to do not hard check everything, if the message is not for here
+      // just ignore it. Change the above checks as well.
```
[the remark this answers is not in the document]
apanizo: We started a conversation last week about how the listener should work. I think it is easier to just check `!sameTarget || !msgToForeground` and remove the `throw` inside `switch`.

It is like: "Ey if the msg is for the extension let's move forward. Then inside the switch only "reacts" if it applies to my context.

I think the `msgToPersonaProvider` can be removed.

packages/sanes-chrome-extension/src/extension/background/model/accountManager/index.ts
```diff
@@ -0,0 +1,113 @@
...
+  readonly identities: ReadonlyArray<PublicIdentity>;
+}
+
+export interface AccountManagerChainConfig {
+  readonly chainId: ChainId;
+  readonly algorithm: Algorithm;
+  readonly derivePath: (account: number) => ReadonlyArray<Slip10RawIndex>;
+}
+
+export class AccountManager {
```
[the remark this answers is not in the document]
apanizo: Yeah, I know, this is why I wrote in the PR's description: "The old src/logic folder is going to be refactored & finally removed when #89, so bear a bit with me".

Will do tomorrow.

packages/sanes-chrome-extension/src/extension/background/model/backgroundscript/db/index.ts
```diff
@@ -0,0 +1,16 @@
+import { AbstractLevelDOWN } from 'abstract-leveldown';
```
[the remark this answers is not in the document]
apanizo: Yep it is 1-1

packages/sanes-chrome-extension/src/extension/background/model/persona/config/codec.ts
```diff
@@ -0,0 +1,74 @@
+import { Algorithm, ChainConnector } from '@iov/bcp';
```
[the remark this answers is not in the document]
apanizo: Yes, it is 1:1

packages/sanes-chrome-extension/src/extension/background/model/persona/config/configurationfile.ts
```diff
@@ -0,0 +1,46 @@
+import { singleton } from '../../../../../utils/singleton';
```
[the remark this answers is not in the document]
apanizo: Yes it is 1:1

packages/sanes-chrome-extension/src/extension/background/model/persona/index.ts
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * All fields must be losslessly JSON serializable/deserializable to allow
+ * messaging between background script and UI.
+ */
+export interface PersonaAcccount {
+  /** human readable address or placeholder text */
+  readonly label: string;
+}
+
+export class Persona {
```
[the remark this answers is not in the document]
apanizo: I understand your point. But I faced 2 situations:
- I did not want to refactor the helper methods for knowing if the persona exists in the db (needed for my next ticket #89)
- I need to do this refactor as fast as I can, and seeing we have distributed complexity without losing entity's domain spec, this is the only way I found to do it in a reasonable amount of time.

packages/sanes-chrome-extension/src/extension/background/model/signingServer/index.ts
apanizo: I think the problem here is you are not familiar with it (yet). Just in the moment you know this SigningServer class is for dealing with external requests.

I know ideally should be independent, but at the end Persona and SigningServer should be connected somehow, because without Persona cannot exist SigningServer, and I prefer to just pass pack the "core" here rather than start the signing server in Persona which does not make really sense (inherited feature caused for bad design thinking persona should live in the app instead of the background).

Anyway this relation is just 1:1 from the main backgroundscript entity, so it does not hurt 😉

packages/sanes-chrome-extension/src/extension/background/model/signingServer/index.unit.spec.ts
apanizo: This is a fair and good comment. I forgot to put the doc message. Unfortunately RequestHandler now it is well encapsulated under SigningServer and for obtaining the next request under a "unit-test", this is the most efficient and simplest way to do it.

Can you figure out an alternative?

packages/sanes-chrome-extension/src/utils/chrome/index.ts
apanizo: I do not really like too much, ideally, we should be able to call them directly from views, but for avoiding DOM tests problems due `chrome` global variable I had to create this class.

It works really well.

apanizo:
That is true. At least we have not lost any functionality because tests are green. Once we release the MVP we can go feature by feature refactoring, I needed to have something usable to explain others how everything works without having big inconsistencies (described in the description).

PR #231, opened 2019-05-29
Remove BCP/BOV blockchain

webmaster128:
As discussed with the backend team, this blockchain will not be used anymore. Upcoming networks will only have a BNS.


webmaster128:
> why this?

because BCP is not needed anymore. We have multiple chains with BNS, Lisk and ETH and can show cross-chain atomic swaps between BNS-ETH.

PR #232, opened 2019-05-29
Feature #89 login logic

### LINKED ISSUE #89: Add login logic

apanizo:
**Description**
This PR checks if there is a persona in the db, and if it has not been loaded, force the app to be redirected to the login view. Then the user can introduce his password and reload the persona.

It closes #89 

This PR:
- Reverts back all remaining unit tests (persona, signingserver, accountManager, db...)
- Removes already refactored code.
- Finishes the refactor process.


packages/sanes-chrome-extension/src/extension/background/model/backgroundscript/index.ts
```diff
@@ -87,6 +92,7 @@ class Backgroundscript {
     (window as IovWindowExtension).loadPersona = pss => this.loadPersona(pss);
     (window as IovWindowExtension).createAccount = () => this.createAccount();
     (window as IovWindowExtension).getPersonaData = () => this.getPersonaData();
+    (window as IovWindowExtension).hasStoredPersona = () => this.hasStoredPersona();
```
albertandrejev: I think it is better to select this type casting into separate variable like `windowExtension` or `iovWindowExtension`, shorter resulting code and easier to maintenance.

webmaster128: I agree. Even better would be to explicitly pass a `target: IovWindowExtension` argument into this function instead of using the global reference

PR #243, opened 2019-05-31
Feature #235 - Fix height issues with PageLayout

### LINKED ISSUE #235: Auto height size to extension
Sometimes when one dynamic view in the extension is higher than the extension' size, the layout container does not fit the entire screen.
See: 
Screenshot 2019-05-29 15 34 40

It seems: 
```
height: "auto"
 minHeight: "100vh" 
```
solve the problem. Check it out.

Also, if the screen is too small, the IOV icon shrinks to the panel. See:
Screenshot 2019-05-29 15 48 21
And it should stay at the bottom of the window.

abefernan:
**Description**
The hardcoded CSS rule in the `body` element `"height: 500px"` means that every view in the extension will have a viewport height of 500px. However, the `height` properties of some of its children are not properly implemented, resulting in the following issues: 

- When a container dynamically grows, its parent container does not adjust its own height, resulting in visual issues.
- The IOV logo at the bottom is not vertically aligned to the end of the view.

There is a color inconsistency addressed in this [#235 comment](https://github.com/iov-one/ponferrada/issues/235#issuecomment-497680085).

Closes #235.

**Screenshots**
The following screenshots are presented as a counterpart of the ones present in issue #235:
imagen
imagen

### DIFF
--- a/packages/medulas-react-components/src/components/PageLayout/index.tsx
+++ b/packages/medulas-react-components/src/components/PageLayout/index.tsx
@@ -7,6 +7,8 @@ import Block from '../Block';
 import Image from '../Image';
 import Typography from '../Typography';
 
+const PAGE_HEIGHT = 500;
+
 interface Props extends StyleProps {
   readonly id?: string;
   readonly children: React.ReactNode;
@@ -17,7 +19,7 @@ interface Props extends StyleProps {
 
 interface StyleProps {
   readonly color?: 'white' | 'transparent';
-  readonly minHeight?: string;
+  readonly minHeight?: string | number;
 }
 
 const useStyles = makeStyles<Theme, StyleProps>({
@@ -34,7 +36,7 @@ const PageLayout = ({
   primaryTitle,
   onBack,
   color = 'transparent',
-  minHeight = '500px',
+  minHeight = PAGE_HEIGHT,
 }: Props): JSX.Element => {
   const showBackArrow = !!onBack;
   const classes = useStyles({ color, minHeight });
--- a/packages/medulas-react-components/src/components/PageLayout/index.tsx
+++ b/packages/medulas-react-components/src/components/PageLayout/index.tsx
@@ -17,24 +17,27 @@ interface Props extends StyleProps {
 
 interface StyleProps {
   readonly color?: 'white' | 'transparent';
+  readonly minHeight?: string;
 }
 
 const useStyles = makeStyles<Theme, StyleProps>({
   root: props => ({
     backgroundColor: props.color,
+    minHeight: props.minHeight,
   }),
 });
 
 const PageLayout = ({
   id,
   children,
-  color = 'transparent',
   title,
   primaryTitle,
   onBack,
+  color = 'transparent',
+  minHeight = '500px',
 }: Props): JSX.Element => {
   const showBackArrow = !!onBack;
-  const classes = useStyles({ color });
+  const classes = useStyles({ color, minHeight });
 
   return (
     <Block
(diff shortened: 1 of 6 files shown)


packages/medulas-react-components/src/components/PageLayout/index.tsx
```diff
@@ -17,24 +17,27 @@ interface Props extends StyleProps {
...
 
 const PageLayout = ({
   id,
   children,
-  color = 'transparent',
   title,
   primaryTitle,
   onBack,
+  color = 'transparent',
+  minHeight = '500px',
```
apanizo: You can use: `EXTENSION_HEIGHT` constant.

packages/sanes-chrome-extension/src/routes/account/index.tsx
```diff
@@ -59,7 +59,7 @@ const AccountView = (): JSX.Element => {
 
   return (
     <Drawer items={items}>
-      <PageLayout id={ACCOUNT_STATUS_ROUTE} primaryTitle="Account" title="Status">
+      <PageLayout id={ACCOUNT_STATUS_ROUTE} primaryTitle="Account" title="Status" minHeight="444px">
```
apanizo: Why 444px?

abefernan: 500-56 because the Drawer is 56px high.

packages/medulas-react-components/src/components/PageLayout/index.tsx
```diff
@@ -17,24 +17,27 @@ interface Props extends StyleProps {
 
 interface StyleProps {
   readonly color?: 'white' | 'transparent';
+  readonly minHeight?: string;
```
albertandrejev: Better use `string | number` instead of `string`. Because it possible to use `number` in case of pixel height.

packages/medulas-react-components/src/components/PageLayout/index.tsx
albertandrejev: That's why you should use `string | number` because `EXTENSION_HEIGHT` is number type constant.

packages/medulas-react-components/src/components/PageLayout/index.tsx
abefernan: But in order to set pixel height you usually do "500px", not "500", and that way it supports every value for height, such as "auto", "100vh", "50rem".

packages/sanes-chrome-extension/src/routes/account/index.tsx
apanizo: The extract to a constant the Drawer height, place it inside Drawer and make the calculation, something like:
```
const MAIN_LAYOUT = CHROME_EXTENSION_HEIGHT - DRAWER_HEIGHT
```

packages/medulas-react-components/src/components/PageLayout/index.tsx
albertandrejev: But material UI allows you to set it as number too. Just for consistency.

albertandrejev: It is even possible to set breakpoints with values for the property (you just need to provide specific order as value). This is not required in our case but it is possible. See https://material-ui.com/system/basics/ (https://material-ui.com/system/basics/#demo)

packages/medulas-react-components/src/components/PageLayout/index.tsx
abefernan: Alright, but `EXTENSION_HEIGHT` is a Sanes' constant, not Medulas', and I don't think there should be imports from Sanes to Medulas. For now I've added the `PAGE_HEIGHT` constant to the PageLayout component to remove the magic number.

PR #244, opened 2019-06-01
Feature - Update libraries

apanizo:
**Description**
In this PR we update libraries reducing the technical debt of the project, also, we organise better dependencies in lerna monorepo.


packages/medulas-react-components/package.json
```diff
@@ -15,15 +15,7 @@
     "not op_mini all"
   ],
   "dependencies": {
-    "@material-ui/core": "4.0.0-alpha.6",
```
webmaster128: Those dependencies belong here. Otherwise a user of medulas-react-components cannot use medulas-react-components due to missing dependencies.

package.json
```diff
@@ -64,19 +70,23 @@
...
+    "final-form": "^4.12.0",
+    "react-final-form-hooks": "^2.0.0",
+    "react": "16.9.0-alpha.0",
+    "react-dom": "16.9.0-alpha.0",
+    "react-redux": "^7.0.2",
+    "react-router": "^5.0.0",
+    "react-router-dom": "^5.0.0",
+    "react-scripts": "3.0.0",
+    "react-test-renderer": "^16.8.6",
+    "redux": "^4.0.1"
```
webmaster128: A lerna monorepo should not have any "dependencies", as only developers will ever install those dependencies, not users.

packages/medulas-react-components/package.json
apanizo: The main reason because I extracted them was not having duplicated versions of the same library, do you know if it is possible to create constants?

package.json
apanizo: 🤔 let me understand this. Which can be the use case of users using our repo without being developers? The only one I can think of is the usage of medulas-react-components, and for that.... maybe is just worth use `peerDependencies`?

Basically what I want to do is, have a way to autodeploy the extension to the chrome store #55, and also autodeploy our example webapps to our AWS instance, so, under this use case, does it hurt, under this circumstances, to still have those dependencies here?

Sorry but I am kind of lost. Please help! 😉

packages/medulas-react-components/package.json
webmaster128: Since .json does not allow coding, the only way I see is to use a code generator that updates all `package.json`s. However, it is much simpler to search the whole project for "@material-ui/core" (including the quotes) via CMD+Shift+F and update 2 or 3 places.

package.json
webmaster128: > Which can be the use case of users using our repo without being developers? 

Such a use case cannot exist. The repo root package.json is only for developers.

> maybe is just worth use peerDependencies

Nope, peerDependencies are something different

> does it hurt, under this circumstances, to still have those dependencies here?

The point is: only developers (and the CI) install the dependencies from repo root package.json. Thus all the dependencies in that file should be in "devDependencies". "dependencies" must be empty.

webmaster128: By the way, a "user" for which the "dependencies" field exists, is someone installing a package from npm.

apanizo: Ok, I can move the `dependencies` to `devDepencies` based on the premise of root's package belongs to developers. So far so good.

But I do not really understand why libraries which are needed for running deployed apps should be in `devDependencies`.

webmaster128: > But I do not really understand why libraries which are needed for running deployed apps should be in devDependencies.

Good point, see my other comment. If a runtime dependency is in the root package.json, this is an error.

package.json
```diff
@@ -66,11 +73,6 @@
     "prettier": "^1.16.4",
     "react": "16.9.0-alpha.0",
```
apanizo: The react library is kept here to avoiding the installation of @storybook's custom version (storybook 5.1.0 uses react's  16.8) and therefore has problems running projects.

One possible solution is to move storybook dependencies to all projects under the `devDepencies` entry, not sure if this could be a better solution.

@webmaster128 thoughts?

webmaster128: The issue is that 16.9.0-alpha.0 [is not compatible](https://jubianchi.github.io/semver-check/#/^16.8.4/16.9.0-alpha.0) to the storybook's version range `^16.8.4` because it is a pre-release version.

When we use react version "^16.8.6" instead, everything is resolved to a single installation:

```
react@^16.7.0, react@^16.8.4, react@^16.8.6:
  version "16.8.6"
  resolved "https://registry.yarnpkg.com/react/-/react-16.8.6.tgz#ad6c3a9614fd3a4e9ef51117f54d888da01f2bbe"
  integrity sha512-pC0uMkhLaHm11ZSJULfOBqV4tIZkx87ZLvbbQYunNixAAvjnC+snJCg0XQXn9VIsttVsbZP/H/ewzgsd5fxKXw==
  dependencies:
    loose-envify "^1.1.0"
    object-assign "^4.1.1"
    prop-types "^15.6.2"
    scheduler "^0.13.6"
```

So, do we really need this pre-release version of react?

apanizo: yes, otherwise we suffer a nasty error testing hooks, and when using material-ui components.

webmaster128: Good, the other option is to force storybook to use 16.9.0-alpha.0 as well:

In root `package.json` add

```
  "resolutions": {
    "react": "16.9.0-alpha.0",
    "react-dom": "16.9.0-alpha.0"
  }
```

apanizo: I do really like this! Will update now.

Just FYI, I have just tested test suite using 16.8 and the problem is React cannot deal with async TestUtils.act calls, which are necessary to perform user interactions (and wait for them).

apanizo: Done!!

PR #255, opened 2019-06-04
Feature – Show human readable address in accounts list

### LINKED ISSUE #106: Account Status view : display human readable address
The middle section of the Account Status view has an account selection box. On the same row, right justified show the human readable address associated with the account, if the user has created one (for example in the wallet application).

**Note**
-  if the user switches accounts the human readable account would change.
- See ticket comments
- This is the place where we should query for the human readable address: https://github.com/iov-one/ponferrada/blob/master/packages/sanes-chrome-extension/src/logic/persona/persona.ts#L154

**Aditional comments**
- Since an account has multiple addresses, only do one query per account (e.g. find the BNS address and query the name for the BNS address)
- Just assume the first chain of `this.signer` is the BNS chain (which we configure that way) and cast it to BnsConnection

webmaster128:
Closes #106

### DIFF
--- a/packages/sanes-chrome-extension/src/extension/background/model/persona/index.unit.spec.ts
+++ b/packages/sanes-chrome-extension/src/extension/background/model/persona/index.unit.spec.ts
@@ -1,5 +1,8 @@
-import { EnglishMnemonic } from '@iov/crypto';
+import { bnsCodec, BnsConnection, RegisterUsernameTx } from '@iov/bns';
+import { Ed25519HdWallet, HdPaths, UserProfile } from '@iov/core';
+import { Bip39, EnglishMnemonic, Random } from '@iov/crypto';
 import { withChainsDescribe } from '../../../../utils/test/testExecutor';
+import { sleep } from '../../../../utils/timer';
 import * as txsUpdater from '../../updaters/appUpdater';
 import { Db } from '../backgroundscript/db';
 import SigningServer from '../signingServer';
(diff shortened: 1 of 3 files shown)


packages/sanes-chrome-extension/src/extension/background/model/persona/index.unit.spec.ts
```diff
@@ -107,6 +110,91 @@ withChainsDescribe('Persona', () => {
...
+      await persona.createAccount(db); // index 1
+      await persona.createAccount(db); // index 2
+      await persona.createAccount(db); // index 3
+
+      const accounts = await persona.getAccounts();
+      expect(accounts.length).toEqual(4);
+      expect(accounts[0].label).toEqual(`${name0}*iov`);
+      expect(accounts[1].label).toEqual(`${name1}*iov`);
+      expect(accounts[2].label).toEqual('Multiple names');
+      expect(accounts[3].label).toEqual('Account 3');
```
apanizo: I was thinking that maybe we should increase the counter, starting from 0 and having some accounts with name, then see "Account 3"  but being the fourth one, could be misleading.

Thoughts?

webmaster128: Yeah, right, I don't know which one is better. On the one hand we have the natural counting account 1, 2, 3, …. On the other hand each account belings to one derication index 0, 1, 2, 3.

So far we started with "Account 0", but this can easily be changed. I have no preference.

apanizo: Me too, no preference.

webmaster128: Let the users decide …

PR #261, opened 2019-06-05
Feature - Show creator when signing tx

apanizo:
**Description**
In this PR when the user signs a transaction we also show the native blockchain of the creator for giving more context. The problem could arise when dealing with multiple Persona's accounts, so it is good to see the address just in case.

Also, some styles have been improved.

Now:
Screenshot 2019-06-05 15 30 32


packages/sanes-chrome-extension/src/extension/background/model/signingServer/index.requests.unit.spec.ts
```diff
@@ -207,4 +200,28 @@ withChainsDescribe('background script handler for website request', () => {
     expect(chromeBazRequest.id).toBe(2);
     expect(chromeBazRequest.reason).toBe('Reason baz');
   }, 8000);
+
+  it.only('generates a creator correctly when signAndPost', async () => {
```
webmaster128: no .only please

PR #262, opened 2019-06-05
Feature – Deploy to Chrome Web Store

### LINKED ISSUE #55: Deploy the extension to Chrome Web Store 🚀⚒

webmaster128:
Closes #55

This is not yet fully tested since I don't have IOV developer credentials for the Chrome Web Store, but all the steps worked for a personal account locally.


packages/sanes-chrome-extension/package.json
```diff
@@ -3,23 +3,31 @@
...
-    "start:firefox:development": "yarn use-config-development && yarn build && web-ext run --source-dir ./build/",
-    "start:firefox:production": "yarn use-config-production && yarn build && web-ext run --source-dir ./build/",
+    "start:firefox:test": "yarn build-test && web-ext run --source-dir ./build/",
+    "start:firefox:development": "yarn build-development && web-ext run --source-dir ./build/",
+    "start:firefox:production": "yarn build-production && web-ext run --source-dir ./build/",
     "test-dom": "yarn use-config-test && CI=true react-scripts test '\\.dom\\.(spec|test)\\.[jt]s[x]?$' --env=jsdom",
     "test-e2e": "yarn use-config-test && CI=true react-scripts test '\\.e2e\\.(spec|test)\\.[jt]s[x]?$' --runInBand --env=jsdom",
     "test-unit": "yarn use-config-test && CI=true react-scripts test '\\.unit\\.(spec|test)\\.[jt]s[x]?$' --env=jsdom",
     "test-single": "yarn use-config-test && CI=true react-scripts test --env=jsdom",
-    "test": "yarn use-config-travis && yarn test-unit && yarn test-dom && yarn test-e2e"
```
apanizo: These are needed for adding config files after a clean build, otherwise, it will fail.

packages/sanes-chrome-extension/public/manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "manifest_version": 2,
   "name": "IOV - Chrome extension",
-  "version": "0.0.0",
+  "version": "will_be_set_in_build_step",
```
apanizo: This is because for updating the extension, we need to increase version, right? Which is the version nomenclature?

packages/sanes-chrome-extension/scripts/remove_key.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
apanizo: Why deleting this key? As far as I remember it was created in a deterministic way to be unique. Why not use it for uploading it to the chrome extension?

packages/sanes-chrome-extension/package.json
webmaster128: The the config file is written into the build folder in `build-test` script now, which is called in a clean `yarn build`. So it is always there, no matter if you test or not.

packages/sanes-chrome-extension/public/manifest.json
webmaster128: The version is copied 1:1 from from package.json into manifest.json during the build. This is just a placeholder to make clear that the manifest version is auto-generated.

apanizo: 👌

packages/sanes-chrome-extension/scripts/remove_key.sh
webmaster128: the `key` should not be used at all:

> You don't usually need to use this value. Instead, write your code so that the key value doesn't matter by using relative paths and extension.getURL. 

https://developer.chrome.com/apps/manifest/key

Foir development it is conveneient to ensure you always get the same ID. When uploading the extension to the web store, the `key` property is forbidden. If it is set, the upload is rejected. The extension ID is assigned by the web store.
```

apanizo: Perfect. 👌

Issue #260, opened 2019-06-04
Implement missed tests


apanizo:
Implement missed tests, described:
https://docs.google.com/spreadsheets/d/1U00bsWQNOT2IGd1cBrc99XRlV2nBHUZFxx_jubNVteo/edit#gid=1209034653

webmaster128:
Closing since this is in Done

PR #266, opened 2019-06-06
Feature - Adds testing utilities for the simplification of future tests

abefernan:
This PR provides methods for the simplification of tests and also removes some code duplication.

DOM manipulation:
- click(): clicks on an element.
- input(): modifies the target.value of an element.
- check(): sets to true the target.checked of an element
- submit(): submits a form.

Route navigation:
- travelTo(): redirects a user to a route. Persona and Requests can be set.
- travelTo<route>(): redirects to <route>. It delegates on the travelTo() method.

### DIFF
--- a/packages/sanes-chrome-extension/src/utils/test/dom.tsx
+++ b/packages/sanes-chrome-extension/src/utils/test/dom.tsx
@@ -7,8 +7,8 @@ import { Provider } from 'react-redux';
 import { Store } from 'redux';
 import { PersonaProvider } from '../../context/PersonaProvider';
 import { RequestProvider } from '../../context/RequestProvider';
-import { Request } from '../../extension/background/actions/createPersona/requestHandler';
-import { GetPersonaResponse } from '../../extension/background/messages';
+import { GetPersonaResponse } from '../../extension/background/model/backgroundscript';
+import { Request } from '../../extension/background/model/signingServer/requestQueueManager';
 import Route from '../../routes';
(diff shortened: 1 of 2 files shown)


packages/sanes-chrome-extension/src/utils/test/dom.tsx
```diff
@@ -31,8 +31,28 @@ export const createDom = (
...
+  ) as any;
 
-export const expectRoute = (store: Store, route: string): void => {
-  expect(store.getState().router.location.pathname).toBe(route);
+export const click = (element: Element): void => {
+  TestUtils.act(() => TestUtils.Simulate.click(element));
+};
+
+export const input = (field: Element, value: any): void => {
+  TestUtils.act(() => {
```
apanizo: Can you give a spin through the tests and update them to use those helpers?

abefernan: I'm doing that for #260, besides adding the new tests.

I'm not sure if it would be worth it to refactor all the existing tests on this PR because by adding the tests documented on the spreadsheet some of them become obsolete or redundant.

Maybe I could add to this PR the refactor of Welcome and Login, that are short and to the point, and showcase this new methods.

Let me know the best course of action.

apanizo: Sounds good, continue doing that in #260. I will merge this PR now. Good job.

Issue #265, opened 2019-06-06
Call the extension to find a user's state


kmw101:
Request to the extension to find the `state` of the user.

There are a number of states which are returned.

- [ ] No extension present - This would be capturing a 404 or something equivalent.
- [ ] Extension installed but no profile
- [ ] Logged out/Not active
- [ ] Logged in/Active

kmw101:
This will be done using error codes and not a specific module to call.

PR #271, opened 2019-06-06
Feature - Tx id available, improved design of account view

apanizo:
**Description**
The account status view required more work than I expected, that is why the logic for introducing in Block Explorer urls via config will be created in a separate PR.

This PR:
- Enhances Link component allowing now external hosts urls.
- Added copy to clipboard icon to copy the transaction's id.
- Added link to block explorer transaction for navigating to etherscan... if it applies
- Refactored Msg and MsgError component making them more user-friendly.
- Fixed problems with typography (link and weight styles)
- Proper storybook store representing all different types of `ProcessedTx`

See:
Screenshot 2019-06-06 15 42 15

### DIFF
--- a/packages/medulas-react-components/src/components/Link/index.tsx
+++ b/packages/medulas-react-components/src/components/Link/index.tsx
@@ -5,10 +5,22 @@ interface Props extends LinkProps {
   readonly children: React.ReactNode;
 }
 
-const Link = ({ children, to, ...rest }: Props): JSX.Element => (
-  <RouterLink to={to} {...rest}>
-    {children}
-  </RouterLink>
-);
+const Link = ({ children, to, ...rest }: Props): JSX.Element => {
+  if (typeof to === 'string') {
+    if (/^(mailto:)|((https?:)?\/\/)/.test(to)) {
+      return (
+        <a href={to} target="_blank" rel="noopener noreferrer" {...rest}>
+          {children}
+        </a>
+      );
+    }
+  }
+
+  return (
+    <RouterLink to={to} {...rest}>
+      {children}
+    </RouterLink>
+  );
+};
 
 export default Link;
--- a/packages/medulas-react-components/src/components/Link/index.tsx
+++ b/packages/medulas-react-components/src/components/Link/index.tsx
@@ -7,7 +7,7 @@ interface Props extends LinkProps {
 
 const Link = ({ children, to, ...rest }: Props): JSX.Element => {
   if (typeof to === 'string') {
-    if (/^(mailto:)|((https?:)?\/\/)/.test(to)) {
+    if (/^((mailto:)|((https?:)?\/\/))/.test(to)) {
       return (
         <a href={to} target="_blank" rel="noopener noreferrer" {...rest}>
           {children}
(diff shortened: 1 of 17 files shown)


packages/medulas-react-components/src/components/Link/index.tsx
```diff
@@ -5,10 +5,22 @@ interface Props extends LinkProps {
...
 }
 
-const Link = ({ children, to, ...rest }: Props): JSX.Element => (
-  <RouterLink to={to} {...rest}>
-    {children}
-  </RouterLink>
-);
+const Link = ({ children, to, ...rest }: Props): JSX.Element => {
+  if (typeof to === 'string') {
+    if (/^(mailto:)|((https?:)?\/\/)/.test(to)) {
```
webmaster128: This also matches the string `xxxx//yyyyy`. I think the last `?` should not be there

packages/sanes-chrome-extension/src/extension/background/model/persona/index.ts
```diff
@@ -27,6 +27,7 @@ export interface ProcessedTx {
   readonly amount: Amount;
   readonly memo?: string;
   readonly time: string;
+  readonly blockExplorerUrl: string | null;
```
webmaster128: This is the direct link to the transaction, right?

packages/sanes-chrome-extension/src/extension/background/model/persona/index.ts
```diff
@@ -136,6 +137,7 @@ export class Persona {
       signer: Encoding.toHex(t.transaction.creator.pubkey.data),
       memo: t.transaction.memo,
       amount: t.transaction.amount,
+      blockExplorerUrl: null,
```
webmaster128: Is there a different ticket to set this?

packages/sanes-chrome-extension/src/extension/background/model/persona/index.ts
apanizo: yep, link to tx in etherscan or other block explorer.

packages/sanes-chrome-extension/src/extension/background/model/persona/index.ts
apanizo: Yes my next PR :) which will be done tomorrow for closing #265 (wrote in the description)

webmaster128: Ah, sorry, I only read the list items

packages/medulas-react-components/src/components/Link/index.tsx
apanizo: Removing the last `?` we won't accept `//foo.bar/`, I think based on our usage we can remove it. Thoughts?

webmaster128: Ah, I see. The issue I found comes because the `^` only binds to the mailto alternative (`^` binds stronger than `|`). So adding addional brackets resolves it: `/^((mailto:)|((https?:)?\/\/))/`. This way, `//foo.bar/` is accepted as well

PR #272, opened 2019-06-06
Feature - Enhance bierzo wallet store configuration.

### LINKED ISSUE #277: Core : Prepare redux store configuration
Create redux store configuration to store extension states as well as methods to set and control extension state during runtime.

- [x] Prepare initial redux store configuration (PR #272 );
- [ ] Add helper methods to control extension state.

albertandrejev:
**Description**
This PR will enhance `bierzo-wallet` store configuration and will add additional actions to check extension state. It will close #277.


packages/bierzo-wallet/src/sequences/extension.ts
```diff
@@ -0,0 +1,29 @@
+/* global chrome */
+import { ThunkAction } from 'redux-thunk';
```
apanizo: This is complex, supposedly redux is a library for making your life easier, not complex. Sequences are something inherited from wallet-demo, which, as the `sanes-chrome-extension` has demonstrated are not necessary.

Keep in mind: In redux we should store information needed for customizing views. In this case, as you have defined, I would create a **user** reducer, using your still keep:
```
const initState: ExtensionState = {
  connected: false,
  installed: false,
};
```

Then have two actions:
```
SET_PERSONA_CONNECTED_ACTION <-- if instead of persona you prefer to use user, 
                                                                        for facing web app approach that is fine with me
SET_EXTENSION_INSTALLED_ACTION
```
On the other hand, please do not wrap chrome communication inside thunk actions, due its nature it is better for having controlled complexity create a folder named `communication` where you should place all logic needed for interact with the extension.

Something like:
```
src/communication/status.ts

  return new Promise(resolve => {
     try {
        chrome.runtime........
     ....
  })
}
```

Then what I would do in the src/index

```
getExtensionStatus()
  .then(status => {
     // create store with initial the just obtained status data
     //  mount react tree as usual
     //  navigate to welcome
  })
```

And finally in welcome route:
```
    const connected = useSelector(state => state.extension.connected)
    const installed = useSelector(state => state.extension.installed)
 
    ....
```

Maybe this big refactor sounds a bit strange at first, but you will be happy once you have finished it, because, when you start testing, you can easily mock using jest the `communication` folder.

And you avoid much difficult code.

Happy coding! 💪

albertandrejev: Yes, I think about it. But you should control state of the extension during all requests. Because user can remove application or logout somehow etc. So you need to be ready to update state during each request, that's why I decided to use thunk action. It will allow you to make request, update state in case of any operation result. And everything in one place. Unfortunately it is no possible to check extension state on during startup, you should do this all the time.

albertandrejev: Also I used same approach 
> SET_PERSONA_CONNECTED_ACTION <-- if instead of persona you prefer to use user, 
>                                                                        for facing web app approach that is fine with me
> SET_EXTENSION_INSTALLED_ACTION
But code start looks ugly because I was required to update both states (`connected`, `installed`) simultaneously in different places depends of operation result and decided to use one action for this

apanizo: This is a good and fair point. 

Let's start doing it on start-up. Once this is clean and done, we can evaluate to do the check on every request. We could do something like:

```
src/routes/foo/index.tsx

const Foo = ({...}: Props): JSX.Element => {
     ...  

    const onWhateverRequestToExtentionInSubmit = async (...) => {
        const extensionStatus = await getExtensionStatus();

       const somethingIsBad = !extensionStatus.connected || !extensionStatus.installed
       if (somethingIsBad) {
           // do whatever, update react, and force to navigate to other place where react is updated
          // or just throw error
       }
       
      await talkToExtension(....)
    }
}
```

But first, let's code the startup solution.

albertandrejev: There is also another issue, we should assume by default that everything is ok with the extension. In case if extension is ready and logged in we can not receive any extension status until user will accept/reject incoming identity sharing request.
Let me describe simple scenario. 
1. Extension was not installed -> we will get exception during `chrome.runtime.sendMessage(extensionId....` step
2. User was not logged in or Persona was not created -> we can understand this during 
```
const parsedResponse = parseJsonRpcResponse2(response);
        if (isJsonRpcErrorResponse(parsedResponse)) {
....
```` 
step
3. Everything is ok -> we can not get this state until user will click on Accept in the extension. So again we at the moment where we should set extension state.

so approach 
```
getExtensionStatus()
  .then(status => {
     // create store with initial the just obtained status data
     //  mount react tree as usual
     //  navigate to welcome
  })
```
will not work. Because in case if everything is ok `then` will be called only after user will accept share identity request. I think this is possible to avoid somehow but with usage of ugly hacks with `timeout` inside `getExtensionStatus()` Promise.

Comment (apanizo) [reply packages/bierzo-wallet/src/sequences/extension.ts]: That is right, good comment. In this case, we need to give a second thought. Seeing this project is a conversion of the old wallet-demo, we need to create the login view first, same as the wallet demo.

So, first is load the reducer using default props:
```
const initState: ExtensionState = {
  connected: false,
  installed: false,
};
```

Then in src/index.tsx
```
1. Create react instance as usual
2. Navigate to login view (in the future we'll navigate automatically to account view)
```

Third in login view:
```
const Login = ({...}) => {
  const dispatch = useDispatch();
  const onLogin = () => {
      toast.show("Please check your extension")
      getExtensionStatus()
        .then(status => {
            if (everythingIsOk) {
              // update redux
              // travel to welcome view, which will be refactored into account view in further steps
           }
           toast.show("Error!!");
        });
  }
}
```

Then in the `Header` component you can do:
```
    const connected = useSelector(state => state.extension.connected)
    const installed = useSelector(state => state.extension.installed)
```

And base on the connected and installed values, you can customize it.

Comment (apanizo) [reply packages/bierzo-wallet/src/sequences/extension.ts]: By the way, create a ticket for investigating what happens if you visit another view while the user interacts with the extension. Once the request is handled what would happen with the response? Is going to be received here?

Comment (albertandrejev) [reply packages/bierzo-wallet/src/sequences/extension.ts]: Yes, it should be received there. Also calling getExtensionStatus will help us only in case of exception. Because in this method we are relaying on requestShareIdentities API call

**Comment by albertandrejev** [comment]
Concrete views/routes implementation will be added in another PR (#273).

File: package.json [hunk 2]
```diff
@@ -48,6 +48,7 @@
     "@storybook/addons": "^5.1.0-alpha.27",
     "@storybook/react": "^5.1.0-alpha.27",
     "@types/classnames": "^2.2.7",
+    "@types/express": "^4.16.1",
```
apanizo: Nop, as long as we use express only in `bierzo-wallet`, this type should remain in `bierzo-wallet` project 😉

packages/bierzo-wallet/src/routes/welcome/index.tsx
```diff
@@ -33,8 +35,15 @@ const onPayment = (): void => {
 
 const Welcome = (): JSX.Element => {
   const toast = React.useContext(ToastContext);
+  //TODO: Fix this as soon as proper react-redux definitions will be available
+  const dispatch = (ReactRedux as any).useDispatch();
   const classes = useStyles();
 
+  const onConnect = async (): Promise<void> => {
+    const request = generateGetIdentitiesRequest();
+    dispatch(getExtension(request));
```
apanizo: instead of `getExtension`, I would say: `setExtensionStatus`.

packages/bierzo-wallet/src/store/types.ts
```diff
@@ -0,0 +1,5 @@
+import { ThunkDispatch } from 'redux-thunk';
```
apanizo: I believe this file can be deleted, and also the redux-thunk library, or am I missing something?

packages/bierzo-wallet/src/store/reducers/extension/types.ts
```diff
@@ -0,0 +1,7 @@
+import { Action } from 'redux';
```
apanizo: For now, those types can live perfectly fine in `reducers/extension/reducer.ts`, the main reason is reducer is the root of everything, and we have this type because we are using redux, and as you can imagine, having 22 lines on `reducer.ts` it is good to move there the type.

Or what do you think? What was the reason to slip those files?

packages/bierzo-wallet/src/store/reducers/extension/state.ts
```diff
@@ -0,0 +1,4 @@
+export interface ExtensionState {
```
apanizo: Same happens here (read the comment written in types.ts).

packages/bierzo-wallet/src/store/types.ts
albertandrejev: But how we will control extension availability during wallet usage session? Should we set state inside the component where we make a request to the extension's data? Or we should set state only during startup and assume that nothing will happens with extension during session?

apanizo: As far as I see the app:
1. The app on startup goes to login page
2. The user logs in, and we update the store
3. The user goes to any view different than login, we use useSelector for knowing the user status, if the user is not logged, redirected to login view.
4. If you are going to perform an operation (sign a tx), first we check the `getExtensionStatus` if we receive bad request (not extension installed or not logged) we show a toast error, we update redux showing the user is not connected, and then, we redirect the app to the login page.

I believe those 4 steps cover the basic UX flow. By the way, we need to create tickets for it. Let me know if we want to do it together.

albertandrejev: Unfortunately there is not `getExtensionStatus` request in background script. As you can remember we discussed this, so the only possible way to get extension status is to call `GetIdentitiesRequest` on the background script. So this call will return offline/not installed state and will return `connected` state only after user will click `Accept` on share identities view of the extension. Are you sure that we should call `GetIdentitiesRequest` every time we want to perform status check. Or we should implement another routine for this in background script?
Also we do not need to check `getExtensionStatus` before performing any operation, because any operation called on the background script will return same error like `getExtensionStatus` so we just need single point where we can check background script responses and set state accordingly. That's was my idea to use Thunk.

packages/bierzo-wallet/src/routes/welcome/index.tsx
albertandrejev: If we want to remove Thunk then we do not need this action and call.

packages/bierzo-wallet/src/store/reducers/extension/types.ts
albertandrejev: I don't mind

packages/bierzo-wallet/src/routes/welcome/index.tsx
apanizo: But what you need to do here is update the redux store with the received information. Once you create the login view which is one of the ticket with the biggest priority (because it unblocks many things), you will remove this from here, but for now, the place where we "log in" is here, right?

packages/bierzo-wallet/src/store/types.ts
apanizo: > Unfortunately there is not getExtensionStatus request in background script.

No, this method lives in the app, is your **encapsulated** helper to know the status of the extension (/src/communication/status/index.ts). 

> As you can remember we discussed this, so the only possible way to get extension status is to call GetIdentitiesRequest on the background script. So this call will return offline/not installed state immediately and will return connected state only after user will click Accept on share identities view of the extension.

As we discussed, the SigningServer protocol is closed, after the MVP maybe we can refactor this part. Create a ticket and put it in the "After MVP" column. So, yeah, use the "getIdentities" trick to know the extension' status. This is not a real problem because, as explained in the last point, inside /src/communication/status/index.ts the communication is well encapsulated. The day we introduce another message in the background script (because the signing server supports it, we just need to update this method, and that is all!!)

> Are you sure that we should call GetIdentitiesRequest every time we want to perform status check. Or we should implement another routine for this in background script?

No, not a requirement from my side. You proposed it, I think we can safely assume, if redux says we are connected, we have not closed the extension. So for me, before sending any signAndPost request to the extension you want to omit this check, is fine with me.

Any other doubts?

albertandrejev: ok, got it. Thank you!

packages/bierzo-wallet/src/routes/welcome/index.tsx
albertandrejev: Yes, you are right

PR #274, opened 2019-06-07
Feature #246 redirect to block explorers

### LINKED ISSUE #246: Show basic info in transactions list (account status page)
Steps for filling info in account status view for the txs box:
```
1. Text tx box is filled (in account status view)
2. Being able to represent in storybook a view showing content in the box
3. Implement urls in config file like names, allowing some urls to be undefined (being able to copy the tx id easily in the frontend)
4. Being able to produce a test where we check the link is correct
```

We have faucets running that you can access from the tests via:

```
const faucet = new IovFaucet("https://local faucet URL");
await faucet.credit(myAddress, "IOV" as TokenTicker);
```

Alternatively, you can use the mnemonic from [here](https://github.com/iov-one/ponferrada/tree/master/scripts/ethereum#generate-secret-keys-for-ganache-cli), which has at least ETH funds

apanizo:
**Description**
In this PR we introduce the available block explorers for transactions.

It closes #246 

Screenshot 2019-06-07 17 17 18

### DIFF
--- a/packages/sanes-chrome-extension/config/development.json
+++ b/packages/sanes-chrome-extension/config/development.json
@@ -4,6 +4,11 @@
     "198f2b61a8eb95fbeed58b8216780b68f697f26b849acf00c8c93bb9b24f783d": "Lisk Devnet",
     "ethereum-eip155-5777": "Ganache"
   },
+  "blockExplorers": {
+    "local-bns-devnet": null,
+    "198f2b61a8eb95fbeed58b8216780b68f697f26b849acf00c8c93bb9b24f783d": null,
+    "ethereum-eip155-5777": null
+  },
   "chains": [
     {
       "chainSpec": {
--- a/packages/sanes-chrome-extension/config/development.json
+++ b/packages/sanes-chrome-extension/config/development.json
@@ -7,7 +7,7 @@
   "blockExplorers": {
     "local-bns-devnet": null,
     "198f2b61a8eb95fbeed58b8216780b68f697f26b849acf00c8c93bb9b24f783d": null,
-    "ethereum-eip155-5777": null
+    "ethereum-eip155-5777": "https://ganache.etherscan.io/tx/"
   },
   "chains": [
     {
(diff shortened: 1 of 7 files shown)


apanizo:
@webmaster128 do you see an easy solution for enabling any block explorer in development environments? I have researched a bit and it seems there are no reliable repositories available. 

I have inspected ganache, and it seems it would not be really difficult to hack the code for enabling browser support.

Thoughts?

packages/sanes-chrome-extension/config/development.json
```diff
@@ -4,6 +4,11 @@
     "198f2b61a8eb95fbeed58b8216780b68f697f26b849acf00c8c93bb9b24f783d": "Lisk Devnet",
     "ethereum-eip155-5777": "Ganache"
   },
+  "blockExplorers": {
+    "local-bns-devnet": null,
+    "198f2b61a8eb95fbeed58b8216780b68f697f26b849acf00c8c93bb9b24f783d": null,
+    "ethereum-eip155-5777": "https://ganache.etherscan.io/tx/"
```
apanizo: @webmaster128  @albertandrejev @abefernan  This is fake, but for testing we generate the link correctly, I have used it. If this feature becomes important in the future, we'll open a new ticket and we'll generate our own etherscan. I have an easy solution in mind for it.

packages/sanes-chrome-extension/src/extension/background/model/persona/config/configurationfile.ts
```diff
@@ -24,9 +24,14 @@ export interface ChainNames {
   readonly [key: string]: string;
 }
 
+export interface BlockExplorers {
+  readonly [key: string]: string;
```
webmaster128: The value type is `string | null`

apanizo: Good catch.

apanizo:
@webmaster128 what I had in mind, is even easier, we can set it up a CouchDb instance (10 lines), which it has out of the box a web interface, and in the scrapper we can add each transaction to the coudb db, and we can show the info `localholst:1234/_index/tx/0xasdfg...asd`

But this can wait, the important thing I wanted to have is at least one chain generating links, for checking the feature works. We have this due the fake url, so I am happy. The other stuff can wait. 

Thanks.

PR #292, opened 2019-06-11
Feature #260 - Missing tests for share-identity route

abefernan:


packages/sanes-chrome-extension/src/routes/share-identity/index.dom.spec.ts
```diff
@@ -73,5 +72,5 @@ describe('DOM > Feature > Share Identity', (): void => {
     await sleep(2000);
     //TODO: Check here that share request rejection has been reject successfuly and permanent
```
apanizo: Remove this unnecessary TODO

PR #294, opened 2019-06-11
Feature #260 - Missing tests for tx-request route

abefernan:


packages/sanes-chrome-extension/src/routes/tx-request/index.dom.spec.ts
```diff
@@ -12,44 +14,41 @@ import {
...
+  }, 60000);
 
   it('should accept incoming request and redirect to account status view', async (): Promise<void> => {
     const inputs = TestUtils.scryRenderedDOMComponentsWithTag(identityDOM, 'button');
 
     expect(inputs.length).toBe(2);
 
     const acceptButton = inputs[0];
-
-    TestUtils.act(() => {
```
apanizo: Can you confirm me those tests do not generate any warning in console?

packages/sanes-chrome-extension/src/routes/tx-request/index.dom.spec.ts
```diff
@@ -70,5 +69,5 @@ describe('DOM > Feature > Transaction Request', (): void => {
     await sleep(2000);
     //TODO: Check here that share request rejection has been reject successfuly and permanent
```
apanizo: This comment should be removed because the whitelist sender is well tested in another place.

packages/sanes-chrome-extension/src/routes/tx-request/index.dom.spec.ts
abefernan: Yes, they do not generate any warning

PR #295, opened 2019-06-11
Feature #260 - Missing tests for welcome route

abefernan:


packages/sanes-chrome-extension/src/utils/test/drawer.ts
```diff
@@ -0,0 +1,33 @@
+import { DRAWER_HTML_ID } from 'medulas-react-components/lib/components/Drawer';
+import TestUtils from 'react-dom/test-utils';
+import { click } from './dom';
+import { findRenderedDOMComponentWithId } from './reactElemFinder';
+
+let recoveryPhraseLink: Element;
+let requestsLink: Element;
+
+const initDrawer = async (dom: React.Component): Promise<void> => {
```
apanizo: I'd change two things here:
1. instead of `dom: React.Component` I would say: `drawerComponent: React.Component`.
2. The location of this file should be in `src/routes/account/test/` because we use the drawer only in this place.

packages/sanes-chrome-extension/src/utils/test/persona.ts
```diff
@@ -0,0 +1,52 @@
+import { Store } from 'redux';
```
apanizo: The location of this file should be: `src/extension/background/model/persona/test` for modularity. I do not want to make huge a general folder.

packages/sanes-chrome-extension/src/utils/test/drawer.ts
abefernan: I don't understand the first one, the parameter is the whole dom of the view.

Issue #247, opened 2019-06-03
Login route UI


kmw101:
The account management logic is now in the extension. The demo wallet log-in page needs to be changed as follows:
- [x] Update redux store with extension state support (PR #272);
- [x] Remove the “Don’t have an IOV wallet” text and hyperlink;
- [x] Remove the Password text and input box.Remove the “Forgotten password” link;

kmw101:
Correct if the account is logged in then go to the /balance page.

albertandrejev:
@apanizo is there any way to check extension status: not logged in, profile doesn'tt exists.

albertandrejev:
I mean is there direct way or this is only possible by analysing behaviour of the extension's answers on `sendMessage` request?

apanizo:
So far the only option is to analyse the extension's answers. But we can open a feature request on the extension, besides `getIdentities` and `signAndPost` a third one like `status` for knowing on which state persona is.

@kmw101 can you give a thought and schedule it for after MVP?

kmw101:
`status` would imply a number of possible `states`. Would a simple `is_connected` be better? 

Agree, should come after MvP.

albertandrejev:
Yes, there is at least two possible `states` I can imagine: logged in, profile created. What else?

kmw101:
Thanks, I will add an issue to the extension and once that is done raise an issue here.

kmw101:
Following the discussion this morning I have amended the issue description.

PR #312, opened 2019-06-13
Feature #286 - Migrate PageColumn component

### LINKED ISSUE #286: Login route: Migrate PageColumn
Migrate `PageMenu`, `PageColumn` and `PageMenuColumn` components from `wallet-demo`.

albertandrejev:
**Description**
This PR will migrate PageColumn component. This component is required for #247.

It will close #286. 

See storybook snapshot:
Screenshot 2019-06-17 16 46 31


packages/medulas-react-components/src/components/Block/index.tsx
```diff
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import Box from '@material-ui/core/Box';
 import { SizingBreakpoint } from '../Grid';
+import classNames from 'classnames';
+import { Theme, makeStyles } from '@material-ui/core';
 
 export type Unit = 0 | 1 | 2 | 3 | 4;
```
apanizo: Can you do a quick check if we have available TS props?

packages/medulas-react-components/src/components/Block/index.tsx
```diff
@@ -75,15 +77,26 @@ interface Props {
...
 
-const Block = ({ children, display = 'block', ...restProps }: Props): JSX.Element => {
+const useStyles = makeStyles((theme: Theme) => ({
+  scroll: {
+    overflowY: 'scroll',
+    overflowX: 'hidden',
+  },
+}));
+
+const Block = ({ children, display = 'block', scroll, className, ...restProps }: Props): JSX.Element => {
```
apanizo: To me, a Block is a layout container where it grows as the content grows. If we need to support scroll I would suggest creating a new component. Call it, BoxScroll or something like that. 😉

packages/medulas-react-components/src/components/Button/index.tsx
```diff
@@ -1,8 +1,30 @@
...
+
+  return (
+    <MuiButton variant={muiVariant} {...restProps}>
+      {spinner && (
+        <Block marginRight={2}>
+          <CircularProgress size={22} color="inherit" />
+        </Block>
+      )}
+      {children}
+      {variant === 'continue' && <ArrowForwardIcon fontSize="small" />}
```
apanizo: Like it.

packages/medulas-react-components/src/pages/PageColumn/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+}: Props): JSX.Element => {
+  const classes = useStyles();
+  const theme = useTheme<Theme>();
+
+  const { form, handleSubmit, submitting, invalid, validating } = useForm({
+    onSubmit,
+  });
+
+  return (
+    <Block
```
apanizo: This component is complex, I believe we can simplify it. Some hints:
1. The login page is the only view which has an image on the left, so we can remove the `leftMenu()` and code hre the content.
2. Material ui has a [container](https://material-ui.com/components/container/)  component which can work for the content, you can specify margins automatically.

If you do not have time and you are blocked, just code the first one and create a new ticket for the second one.

packages/medulas-react-components/src/utils/storybook/index.tsx
```diff
@@ -9,27 +9,27 @@ interface Props {
...
+      margin: '0',
+      padding: '0',
+      bottom: '0',
+      top: '0',
+      left: '0',
+      right: '0',
       overflowX: 'hidden',
-      fontFamily: "'Muli', sans-serif",
-      margin: 0,
-      backgroundColor: theme.palette.background.default,
```
apanizo: Why removing this as well?

packages/medulas-react-components/package.json
```diff
@@ -18,7 +18,9 @@
     "@material-ui/core": "4.0.2",
     "@material-ui/icons": "4.0.1",
     "@material-ui/styles": "4.0.2",
+    "@types/lodash": "^4.14.134",
     "add": "^2.0.6",
+    "lodash": "^4.17.11",
```
apanizo: Why lodash? This is a huge library and I prefer to do not include it. Where are you using it?

albertandrejev: hmm, you are right, there is no lodash dependencies.

packages/medulas-react-components/src/utils/storybook/index.tsx
albertandrejev: I just get it from `bierzo-wallet` storybook config

packages/medulas-react-components/src/components/Button/index.tsx
```diff
@@ -1,8 +1,30 @@
-import * as React from 'react';
+import React from 'react';
+import { Omit } from '@material-ui/core';
```
abefernan: Maybe use Typescript 3.5's Omit.

Issue #310, opened 2019-06-13
Login page: Logic


abefernan:
Part of Feature #296.

Implement the expected behaviour of the Login view, involving interaction with the Browser Extension.

kmw101:
link to #296

webmaster128:
Closing since this is in Done

PR #326, opened 2019-06-18
Feature #310 - Login Logic

abefernan:
This PR adds logic to the Login view that makes sure that the extension is installed and the user logged in before giving access to the Dashboard.

### DIFF
--- a/packages/sil-voting-app/src/communication/identities/index.ts
+++ b/packages/sil-voting-app/src/communication/identities/index.ts
@@ -1,0 +1,56 @@
+/*global chrome*/
+import { isPublicIdentity, PublicIdentity } from '@iov/bcp';
+import { TransactionEncoder } from '@iov/core';
+import { isJsonRpcErrorResponse, JsonRpcRequest, makeJsonRpcId, parseJsonRpcResponse2 } from '@iov/jsonrpc';
+import { extensionId } from '..';
+
+export const generateGetIdentitiesRequest = (): JsonRpcRequest => ({
+  jsonrpc: '2.0',
+  id: makeJsonRpcId(),
+  method: 'getIdentities',
+  params: {
+    reason: TransactionEncoder.toJson('I would like to know who you are on Ethereum'),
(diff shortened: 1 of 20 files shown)


packages/sil-voting-app/src/communication/identities/index.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  return data.every(isPublicIdentity);
+};
+
+type GetIdentitiesResponse = ReadonlyArray<PublicIdentity> | undefined;
+
+export const sendGetIdentitiesRequest = async (): Promise<GetIdentitiesResponse> => {
+  const request = generateGetIdentitiesRequest();
+
```
apanizo: Check again my PR you can have problems if you do not check if the variable chrome is available, this might happen if the browser does not have any extension.

packages/sil-voting-app/src/routes/login/index.tsx
```diff
@@ -39,7 +66,9 @@ const Login = (): JSX.Element => {
         <Typography variant="h6">IOV Voting Dashboard</Typography>
       </Block>
       <Block display="flex">
-        <Button className={classes.button}>LOG IN</Button>
+        <Button className={classes.button} onClick={onLogin}>
+          LOG IN
```
apanizo: I know is kind of big solution for a small thing, but I recommend to use Form here, just because is quite easy to disable the button while the onSubmit callback is being executed.

I know, it is weird because you are going to use a form without any input, but it works fast and in a clean way. Check Login view of bierzo-wallet.

packages/sil-voting-app/src/communication/identities/index.ts
abefernan: Addressed in 051d4c71f5ccc759afe99f19c20aeba886220888.

packages/sil-voting-app/src/routes/login/index.tsx
abefernan: Addressed in 1378a3097ebf1cf14e1366d884ada037b92cc7f7.

PR #331, opened 2019-06-24
Feature #319 - Dashboard Header

### LINKED ISSUE #319: Dashboard structure: Top bar
Part of Feature #279.

Implement the layout and the styling for the Dashboard's Top bar.

abefernan:
**Description**
This PR adds the Header component to the Dashboard route of the Voting App.

Closes #319.

**Screenshots**
imagen


packages/sil-voting-app/src/components/Header.tsx
```diff
@@ -0,0 +1,33 @@
...
+    },
+  },
+}));
+
+const Header = (): JSX.Element => {
+  const classes = useStyles();
+
+  return (
+    <Block width="100%" minHeight="78px" display="flex" alignItems="center">
+      <Block width="20%" display="flex" alignItems="center" justifyContent="center" className={classes.logo}>
```
apanizo: Strange way to do the layout. I am not saying it's incorrect, or bad code, just saying I am bit surprised. If parent's root Block uses "flex", then I expect its children to have automatic width regulated by `flexGrow`. 
If for any reason you need specific width, like in your case, I expected to use flexBasis property. Something like:
```
<Block width="100%" display="flex">
   <Block flexBasis="200">This is the left menu</Block>
   <Block flexGrow={1}>This is the main content</Block>
</Block>
```

Taken from css-tricks: 
> flex-basis -> This defines the default size of an element **before the remaining space is distributed**

So, you can assume flexGrow=1 will take all the space available.

packages/sil-voting-app/src/components/Header.tsx
```diff
@@ -0,0 +1,33 @@
...
+        <CircleImage alt="Logo" icon={icon} dia="48px" circleColor="#fff" />
```
apanizo: Strange way to do the layout. I am not saying it's incorrect, or bad code, just saying I am bit surprised. If parent's root Block uses "flex", then I expect its children to have automatic width regulated by `flexGrow`. 
If for any reason you need specific width, like in your case, I expected to use flexBasis property. Something like:
```
<Block width="100%" display="flex">
   <Block flexBasis="200">This is the left menu</Block>
   <Block flexGrow={1}>This is the main content</Block>
</Block>
```

Taken from css-tricks: 
> flex-basis -> This defines the default size of an element **before the remaining space is distributed**

So, you can assume flexGrow=1 will take all the space available.

abefernan: I tried with that, but flexGrow=0.2 and flexGrow=0.8 was not giving the same results as width=20% and width=80%. Therefore, it was not aligning with the drawer and the list of the next PR.

I will tinker with flexGrow it a bit more.

apanizo: Note my solution is a mix of flexBasis and flexGrow (which by the way only accept positive integers, so if you want to use only fleGrow % solution with the rule 20 / 80, you should use 2, 8 respectively).

But again I think the left menu should be fixed to some pixels, and therefore use the `flexBasis && flexGrow` solution.

packages/sil-voting-app/src/components/Header.tsx
```diff
@@ -0,0 +1,34 @@
...
+import Block from 'medulas-react-components/lib/components/Block';
+import CircleImage from 'medulas-react-components/lib/components/Image/CircleImage';
+import Typography from 'medulas-react-components/lib/components/Typography';
+import React from 'react';
+import icon from '../assets/iov-logo.svg';
+
+const useStyles = makeStyles((theme: Theme) => ({
+  logo: {
+    flexBasis: '205px',
+    '& > div + h5': {
```
apanizo: What is this?

packages/sil-voting-app/src/components/Header.tsx
```diff
@@ -0,0 +1,34 @@
...
+    },
+  },
+}));
+
+const Header = (): JSX.Element => {
+  const classes = useStyles();
+
+  return (
+    <Block width="100%" minHeight="78px" display="flex" alignItems="center">
+      <Block display="flex" alignItems="center" justifyContent="center" className={classes.logo}>
```
apanizo: Why not using flexBasis directly here?

abefernan: Because Block does not have the `flexBasis` property.

packages/sil-voting-app/src/components/Header.tsx
abefernan: It's a way to specify the gaps in the container element instead of on each element, but in this case I can apply it directly to the `h5` to make it more readable.

apanizo: Yes, it is better if you assign directly a new rule directly to those elements, rather than using selectors on parents elements.

Does it complicate your solution?

packages/sil-voting-app/src/components/Header.tsx
apanizo: Interesting, I thought it was a lack of typing, but the actual implementation does not support it. Despite it is not the purist solution, I think is safe enough if you just declare `width="205px"` (or the size you want to apply to the left column) on the left element. Let me know how this work.

packages/sil-voting-app/src/components/Header.tsx
abefernan: It's okay, I wrapped Typography in a Block with marginLeft.

packages/sil-voting-app/src/components/Header.tsx
abefernan: `width="205px"` has some wrong behaviour when the screen shrinks a bit. However, `minWidth="205px"` works well. There is also no longer need for `flexGrow` in this case since it does not do anything.

PR #332, opened 2019-06-24
Feature #321 - Dashboard main view

### LINKED ISSUE #321: Dashboard structure: Main view
Part of Feature #279.

Implement the layout and the styling for the Dashboard's Main view.

abefernan:
**Description**
This PR adds the layout for the filter on the left side of the page and a dummy skeleton of the future list of proposals.

Closes #321;

**Screenshots**
imagen

### DIFF
--- a/packages/sil-voting-app/src/routes/dashboard/index.tsx
+++ b/packages/sil-voting-app/src/routes/dashboard/index.tsx
@@ -1,16 +1,18 @@
 import Block from 'medulas-react-components/lib/components/Block';
 import Hairline from 'medulas-react-components/lib/components/Hairline';
-import Typography from 'medulas-react-components/lib/components/Typography';
 import React from 'react';
 import Header from '../../components/Header';
+import AsideFilter from './components/AsideFilter';
+import ProposalsList from './components/ProposalsList';
 
 const Dashboard = (): JSX.Element => {
   return (
     <Block width="100%" maxWidth="1024px" height="auto" display="flex" flexDirection="column" margin="0 auto">
       <Header />
       <Hairline />
-      <Block minWidth="100%" minHeight="100%">
-        <Typography variant="h6">DASHBOARD VIEW</Typography>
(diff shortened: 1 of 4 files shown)


packages/sil-voting-app/src/routes/dashboard/index.tsx
```diff
@@ -1,16 +1,18 @@
...
 import Hairline from 'medulas-react-components/lib/components/Hairline';
-import Typography from 'medulas-react-components/lib/components/Typography';
 import React from 'react';
 import Header from '../../components/Header';
+import AsideFilter from './components/AsideFilter';
+import ProposalsList from './components/ProposalsList';
 
 const Dashboard = (): JSX.Element => {
   return (
     <Block width="100%" maxWidth="1024px" height="auto" display="flex" flexDirection="column" margin="0 auto">
```
albertandrejev: Why don't you want to use `Grid` component instead of `Block` in this case?

packages/sil-voting-app/src/routes/dashboard/index.tsx
```diff
@@ -1,16 +1,18 @@
...
       <Header />
       <Hairline />
-      <Block minWidth="100%" minHeight="100%">
-        <Typography variant="h6">DASHBOARD VIEW</Typography>
+      <Block minWidth="100%" display="flex">
```
albertandrejev: Same comment as for upper `Block`.

packages/sil-voting-app/src/routes/dashboard/index.tsx
abefernan: `Grid` doesn't have a `margin` property so I can't horizontal align it.

albertandrejev: Yes but you can use nested component for this. But I thought a little about it and think `Grid` component idea is not suitable in this case. Sorry for confusing.

Issue #320, opened 2019-06-17
Dashboard structure: List of Proposals


abefernan:
Part of Feature #279.

Implement the layout, the styling, and the testing for the Dashboard's List of proposals. The Proposal component should act as a ListItem.

PR #333, opened 2019-06-24
Feature #320 - Dashboard Proposal List

abefernan:
**Description**
This PR implements the layout and the "layout logic" for the list of proposals of the Dashboard.

- The Proposal component itself is pretty complex and therefore presents a lot of conditional rendering. Because of that, it has been divided in the following components:
  - Title
  - CreationData
  - Description
  - Status
- A randomizer utility has been implemented so that the layout could be tested against every type of instance of a proposal. However, in order to simplify the storybook snapshot matching, it has been replaced by simple hand-made proposals in production code.
- The [Zeplin design](https://app.zeplin.io/project/5cb65f5c34967632b48fcb12/dashboard) has been followed in order to organize the layout.

Part of #320.

**Screenshots**
imagen


packages/sil-voting-app/src/routes/dashboard/utils/randomProposal.ts
```diff
@@ -0,0 +1,96 @@
+import { ProposalProps, VoteResult } from '../components/Proposal';
```
apanizo: Where do you use those utilities? I think you should place them inside: `dashboard/utils/test/randomProposal.ts`

packages/sil-voting-app/src/routes/dashboard/utils/randomProposal.ts
```diff
@@ -0,0 +1,96 @@
...
+const getStatus = (
+  expiryDate: Date,
+  vote: 'Invalid' | 'Yes' | 'No' | 'Abstain',
+): 'Active' | 'Submitted' | 'Ended' => {
+  if (new Date() > expiryDate) return 'Ended';
+  if (vote !== 'Invalid') return 'Submitted';
+  return 'Active';
+};
+
+export const randomProps = (): ProposalProps => {
```
apanizo: Rename it to `aRandomProposal`

packages/sil-voting-app/src/routes/dashboard/components/Proposal.tsx
```diff
@@ -0,0 +1,191 @@
...
+          <Block marginLeft={2}>
+            <Typography variant="body1">Proposal ID: {props.id}</Typography>
+          </Block>
+          <Block marginLeft={2}>
+            <Typography variant="body1">
+              Created on {props.creationDate.toLocaleDateString('es-ES')}
+            </Typography>
+          </Block>
+        </Block>
+        {props.description.length < DESC_MAX_LENGTH && (
```
apanizo: From line 81 to 105 should be extracted into:
`src/routes/dashboard/components/Proposal/Description.tsx` note Proposal.tsx is now -> `dashboard/components/Proposal/intex.tsx`

packages/sil-voting-app/src/routes/dashboard/components/Proposal.tsx
```diff
@@ -0,0 +1,191 @@
...
+              <Typography inline link onClick={onClick} variant="body1" weight="semibold">
+                Read less
+              </Typography>
+            </Collapse>
+          </React.Fragment>
+        )}
+        <Block display="flex" alignItems="center">
+          {props.status !== 'Ended' && (
+            <Typography variant="body1" weight="semibold">
+              Expires on {props.expiryDate.toLocaleDateString('es-ES')}
```
apanizo: Why `toLocaleDateString` and not just `toString`?

packages/sil-voting-app/src/routes/dashboard/components/Proposal.tsx
```diff
@@ -0,0 +1,191 @@
...
+            </Typography>
+          )}
+          {props.status === 'Ended' && (
```
apanizo: From line 106 to 168 (status conditions) should be refactored into a separate component Proposal/Status.tsx

apanizo:
Also between all different elements in the Proposal, include a bit of marginBottom, if you see the attached screenshot, the component seems to be really compressed. Good job!

packages/sil-voting-app/src/routes/dashboard/components/Proposal.tsx
abefernan: `toLocaleDateString` produces: `27/3/2005`
`toString` produces: `Tue Jan 20 2009 10:22:17 GMT+0100 (hora estándar de Europa central)`

packages/sil-voting-app/src/routes/dashboard/components/AsideFilter.tsx
```diff
@@ -1,33 +1,41 @@
 import { Drawer, List, ListItem, ListItemText } from '@material-ui/core';
```
apanizo: Here I can see the AsideFilter, where is the "header" I assume you have coded it in a separate PageLayout (or something like that), right?

packages/sil-voting-app/src/routes/dashboard/components/Proposal/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+          <Description description={props.description} />
+        </Block>
+        <Status
+          expiryDate={props.expiryDate}
+          quorum={props.quorum}
+          result={props.result}
+          status={props.status}
+        />
+      </Block>
+      <Block minWidth="100px" margin={2} display="flex" flexDirection="column">
```
apanizo: I still think you can extract this Block into <VoteActions ... /> component.

packages/sil-voting-app/src/routes/dashboard/components/Proposal/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+  readonly result: VoteResult;
+  readonly vote: 'Invalid' | 'Yes' | 'No' | 'Abstain';
+  readonly status: 'Active' | 'Submitted' | 'Ended';
+}
+
+const Proposal = (props: ProposalProps): JSX.Element => {
+  return (
+    <Block width="100%" display="flex" alignItems="center">
+      <Block flexGrow={1} margin={2}>
+        <Block display="flex" alignItems="center" marginBottom={1}>
```
apanizo: Extract this Block to <Title />

packages/sil-voting-app/src/routes/dashboard/components/Proposal/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+          <Block marginLeft={2}>
+            <Typography variant="body1">Proposal ID: {props.id}</Typography>
+          </Block>
+          <Block marginLeft={2}>
+            <Typography variant="body1">
+              Created on {props.creationDate.toLocaleDateString('es-ES')}
+            </Typography>
+          </Block>
+        </Block>
+        <Block marginBottom={1}>
```
apanizo: Extract this to <Description />

packages/sil-voting-app/src/routes/dashboard/components/AsideFilter.tsx
abefernan: The header is in `src/components/Header.tsx` as you suggested in a frontend weekly call.

packages/sil-voting-app/src/routes/dashboard/components/Proposal/index.tsx
```diff
@@ -0,0 +1,48 @@
...
+  readonly quorum: number;
+  readonly threshold: number;
+  readonly result: VoteResult;
+  readonly vote: 'Invalid' | 'Yes' | 'No' | 'Abstain';
+  readonly status: 'Active' | 'Submitted' | 'Ended';
+}
+
+const Proposal = (props: ProposalProps): JSX.Element => {
+  return (
+    <Block width="100%" display="flex" alignItems="center">
```
apanizo: Good job here.

packages/sil-voting-app/src/routes/dashboard/components/Proposal/Description.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+interface Props {
+  readonly description: string;
+}
+
+const Description = (props: Props): JSX.Element => {
+  const [expanded, setExpanded] = useState(false);
+
+  const onClick = (): void => {
+    setExpanded(!expanded);
```
apanizo: This is not correct, you can have race conditions here. You should update expanded based on the previous state:
```
setExpanded(prevExpanded => !prevExpanded)
```

packages/sil-voting-app/src/routes/dashboard/components/Proposal/Description.tsx
```diff
@@ -0,0 +1,51 @@
...
+  };
+
+  return (
+    <Block marginBottom={1}>
+      {props.description.length < DESC_MAX_LENGTH && (
```
apanizo: Seeing this component does not have a big hierarchy pending from it I would do it more readable:

```
const smallDescription = props.description.length < DESC_MAX_LENGTH
if (smallDescription) {
  return <Block marginBottom={1}><Typography variant="body1">{props.description}</Typography></Block>
}

if (!expanded) {
   return (
       <Block marginBottom={1}>
          <Typography inline variant="body1">
              {elipsify(props.description, DESC_MAX_LENGTH)}{' '}
          </Typography>
          <Typography inline link onClick={onClick} variant="body1" weight="semibold">
              Read more
          </Typography>
      </Block>
    )
}

return (
       <Block marginBottom={1}>
          <Collapse in={expanded} timeout="auto" unmountOnExit>
            <Typography inline variant="body1">
              {props.description}{' '}
            </Typography>
            <Typography inline link onClick={onClick} variant="body1" weight="semibold">
              Read less
            </Typography>
          </Collapse>
      </Block>
    )
```

In this way, we can understand what the component returns based on its internal state in less than one second.

packages/sil-voting-app/src/routes/dashboard/components/Proposal/Status.tsx
```diff
@@ -0,0 +1,106 @@
...
+  if (maxVotes === 0) vote = 'Abstain';
+
+  const percentage = Math.floor((maxVotes / getTotalVotes(result)) * 100);
+
+  return { vote, percentage };
+};
+
+const Status = (props: Props): JSX.Element => {
+  return (
+    <React.Fragment>
```
apanizo: Split this component into `<StatusEnded />` and  `<StatusPending />` then, in the Proposal/index.tsx introduce the logic for rendering one or the other.

Right now is complex to follow.

packages/sil-voting-app/src/routes/dashboard/components/Proposal/Description.tsx
abefernan: I agree with the simplification. However, this code makes the `Collapse` animation disappear. The rendering is immediate when clicking "Read more" or "Read less".

apanizo: Ahhh, I see. Good point. In this case I suggest:
```
return (
    <Block marginBottom={1}>
        { !expanded && <ReadMore .../> }
         <Collapse in={expanded} timeout="auto" unmountOnExit>
            <Typography inline variant="body1">
              {props.description}{' '}
            </Typography>
            <Typography inline link onClick={onClick} variant="body1" weight="semibold">
              Read less
            </Typography>
         </Collapse>
    </Block>
```

And `<ReadMore />` component should be inside the same class, meaning Description.tsx, there is no need to create a separate component, you can inline it as a variable.

packages/sil-voting-app/src/routes/dashboard/components/Proposal/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <Block flexGrow={1} margin={2}>
+        <Title title={props.title} status={props.status} />
+        <CreationData author={props.author} id={props.id} creationDate={props.creationDate} />
+        <Description description={props.description} />
+
+        {props.status !== 'Ended' && <StatusPending expiryDate={props.expiryDate} />}
+
+        {props.status === 'Ended' && (
```
apanizo: I'd recommend:
```
const hasEnded = props.status === 'Ended';
const showExpireDate = !hasEnded;
...
{ showExpiryDate && ...}
{ hadEnded && ...}
```

packages/sil-voting-app/src/routes/dashboard/components/Proposal/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const Proposal = (props: ProposalProps): JSX.Element => {
+  return (
+    <Block width="100%" display="flex" alignItems="center">
+      <Block flexGrow={1} margin={2}>
+        <Title title={props.title} status={props.status} />
+        <CreationData author={props.author} id={props.id} creationDate={props.creationDate} />
+        <Description description={props.description} />
+
+        {props.status !== 'Ended' && <StatusPending expiryDate={props.expiryDate} />}
```
apanizo: I'd recommend:
```
const hasEnded = props.status === 'Ended';
const showExpireDate = !hasEnded;
...
{ showExpiryDate && ...}
{ hadEnded && ...}
```

packages/sil-voting-app/src/routes/dashboard/components/Proposal/StatusPending.tsx
```diff
@@ -0,0 +1,38 @@
...
+        <Block marginLeft={2}>
+          <Block display="flex" alignItems="center">
+            <Img src={deleteIcon} alt="Delete Icon" height="16px" />
+            <Typography variant="body1" weight="semibold">
+              Delete
+            </Typography>
+          </Block>
+        </Block>
+      </Block>
+      <Block padding={1} bgcolor="#d8d8d8" borderRadius="16px">
```
apanizo: Do not use constants directly here, you have two options:
1. Use useTheme from material-ui and access a color prop from there.
2. Create theme/css.ts file where you place the constants with a readable name like BOX_BACKGROUND

Generally speaking is better to use 1, because if you change the theme, automatically you have changed the color, but if you are using a very specific one which is not related with the theme, feel free to use second option.

packages/sil-voting-app/src/routes/dashboard/components/Proposal/StatusEnded.tsx
```diff
@@ -0,0 +1,77 @@
...
+            </Typography>
+          </Block>
+          <Block marginLeft={2}>
+            <Typography variant="body1" weight="semibold">
+              Result: {getWinner(props.result).vote}
+            </Typography>
+          </Block>
+        </React.Fragment>
+      </Block>
+      <Block width="100%" border="1px solid #979797" borderRadius="16px">
```
apanizo: Same here

apanizo: In this case you can use useStyles hook (check sources for example)

packages/sil-voting-app/src/routes/dashboard/components/Proposal/StatusEnded.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <Block
+          padding={1}
+          bgcolor="#d8d8d8"
```
apanizo: No constants, please.

packages/sil-voting-app/src/routes/dashboard/components/Proposal/StatusPending.tsx
abefernan: I used approach 2 because the two colors that I used are not defined in the theme.

packages/sil-voting-app/src/routes/dashboard/components/Proposal/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const hasEnded = props.status === 'Ended';
+  const showExpiryDate = !hasEnded;
+
+  return (
+    <Block width="100%" display="flex" alignItems="center">
+      <Block flexGrow={1} margin={2}>
+        <Title title={props.title} status={props.status} />
+        <CreationData author={props.author} id={props.id} creationDate={props.creationDate} />
+        <Description description={props.description} />
+
```
apanizo: No blank line space, please

packages/sil-voting-app/src/routes/dashboard/components/Proposal/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+        {showExpiryDate && <StatusPending expiryDate={props.expiryDate} />}
+
```
apanizo: No blank line space, please

Issue #305, opened 2019-06-13
Allow different transaction types to be signed


webmaster128:
Right now `Persona.processTransaction` can only handle `SendTransaction`s. The same is true for `ProcessedTx` and  the UI.

We will soon need the ability to use other transaction types as well, starting with `RegisterUsernameTx`.

This ticket is about
1. generalizing transaction handling such that different types will work
2. show that 1. works by adding support for `RegisterUsernameTx`

PR #339, opened 2019-06-26
Feature – Prepare ProcessedTx handling to support multiple transaction types

webmaster128:
Part 1 of https://github.com/iov-one/ponferrada/issues/305

Missing

- [x] Support multiple transactions in tx-request component

### DIFF
--- a/packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
+++ b/packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
@@ -47,14 +47,11 @@ export const PersonaProvider = ({ children, persona }: Props): JSX.Element => {
 
     chrome.runtime.onMessage.addListener((msg, sender, _sendResponse) => {
       const sameTarget = sender.id === chrome.runtime.id;
-      const msgToPersonaProvider = msg.action === MessageToForegroundAction.TransactionsChanged;
-      if (!sameTarget || !isMessageToForeground(msg) || !msgToPersonaProvider) {
+      if (!sameTarget || !isMessageToForeground(msg)) {
         // Only handle messages from background script
         return;
       }
 
-      // TODO Refactor to do not hard check everything, if the message is not for here
-      // just ignore it. Change the above checks as well.
       switch (msg.action) {
         case MessageToForegroundAction.TransactionsChanged:
           if (!Array.isArray(msg.data)) {
@@ -63,7 +60,7 @@ export const PersonaProvider = ({ children, persona }: Props): JSX.Element => {
           setTxs(msg.data);
           break;
         default:
-          throw new Error('Unknown action');
+        // this listener is not responsible for the given message, ignore message
       }
     });
   }, []);
--- a/packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
+++ b/packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
@@ -47,9 +47,8 @@ export const PersonaProvider = ({ children, persona }: Props): JSX.Element => {
 
     chrome.runtime.onMessage.addListener((msg, sender, _sendResponse) => {
       const sameTarget = sender.id === chrome.runtime.id;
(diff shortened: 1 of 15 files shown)


packages/sanes-chrome-extension/src/context/PersonaProvider.tsx
```diff
@@ -47,24 +47,29 @@ export const PersonaProvider = ({ children, persona }: Props): JSX.Element => {
...
+              }
+            })
+            .catch(error => {
+              // eslint-disable-next-line no-console
+              console.error(error);
+            });
           break;
         default:
-          throw new Error('Unknown action');
+        // this listener is not responsible for the given message, ignore message
```
albertandrejev: In this case do we need `default` at all?

webmaster128: Good point. You are right, it is not strictly needed. But I found it useful to use it as a place for the comment. I think it is always good to be explicit about all possible cases. Before, we threw an Error but now we explicitly ignore other messages.

packages/sanes-chrome-extension/src/routes/tx-request/components/ShowRequest/index.tsx
```diff
@@ -14,33 +15,43 @@ interface Props {
...
-    <Block marginBottom={2} />
-  </PageLayout>
-);
+const Layout = ({ sender, tx, creator, onAcceptRequest, showRejectView }: Props): JSX.Element => {
+  let req: JSX.Element;
+
+  if (isSendTransaction(tx)) {
+    req = <ReqSendTransaction tx={tx} creator={creator} />;
+  } else {
+    throw new Error('Received transaction type that cannot be displayed');
```
albertandrejev: This will make extension to crash if we will not implement "[Error Boundaries](https://reactjs.org/blog/2017/07/26/error-handling-in-react-16.html#introducing-error-boundaries)" for the extension.

webmaster128: Crashing is perfect in this case. Incoming transactions with unsupported type are already rejected by the signing server: https://github.com/iov-one/ponferrada/blob/5fd61f2/packages/sanes-chrome-extension/src/extension/background/model/signingServer/index.ts#L67-L69 If we receive an unsupported transaction type in this part of the code, this is a programming error.

albertandrejev: So we need to implement "Error Boundaries" to receive reports to Sentry about this. @kmw101 what do you think?

webmaster128: "Error Boundaries" are an interesting concept. Thanks for the link. However, I tend to disagree with

> A JavaScript error in a part of the UI shouldn’t break the whole app.

If there is an unhandled exception, this means no developer thought about this error case before. Ignoring such an exception and continuing to run the app can lead to invalid app state and undetected errors. Letting unhandled exceptions crash the application increase chances that the programming error is detected during development or when running tests.

I think we should just crash the application for now and add Sentry later on as an additional way to detect and understand errors that happen on the user's machine and did not occur during development.

albertandrejev: Anyway we need to create issue for this case. Let's wait what @kmw101 will tell about this.

kmw101: I tend to agree with the point about letting unhandled exceptions crash, I prefer that to something silently failing and nobody knows or can find out what happened.

Let's keep this simple and if we have time later publish to Sentry.

webmaster128: Added an issue: https://github.com/iov-one/ponferrada/issues/347

PR #348, opened 2019-07-01
Feature #342 - Add T&C links

### LINKED ISSUE #342: Add T&C links
The Browser Extension needs to have terms and conditions of use. 

Please add a link to the account status page to 

https://support.iov.one/hc/en-us

With the text Terms and Conditions of use.

albertandrejev:
**Description**
This PR will add T&C link to the Account state view. It will close #342.

**Before**
toc-before

**After**
toc-after


packages/sanes-chrome-extension/src/routes/account/index.dom.spec.ts
```diff
@@ -112,7 +112,8 @@ withChainsDescribe('DOM > Feature > Account Status', () => {
 
     // Check for the link
     const links = TestUtils.scryRenderedDOMComponentsWithTag(accountStatusDom, 'a');
-    expect(links.length).toBe(1);
+    //Amount of links calculation: 1 link for the blockExplorerUrl + 1 link for T&C
+    expect(links.length).toBe(2);
```
webmaster128: Can't we change the test in a way to better search the link we look for? This test is about

> generates a link inside transaction box for an ethereum transaction

so we should search in the transaction box only and ignore other parts of the screen. Checking the T&C could be a different test that needs no transaction logic.

packages/sanes-chrome-extension/src/routes/account/index.tsx
```diff
@@ -82,6 +83,13 @@ const AccountView = (): JSX.Element => {
         <Block>
           <ListTxs title="Transactions" txs={personaProvider.txs} />
         </Block>
+        <Block marginTop={2}>
+          <Link to="https://support.iov.one/hc/en-us">
```
webmaster128: What about storing this link in the cnfig file (especially since we expect changes)?

albertandrejev: I do not think that we will change it more then once also it will be required to update extension in store in both cases (if we will hardcode URL or even if we will add it to the config file).

packages/sanes-chrome-extension/src/routes/account/index.dom.spec.ts
albertandrejev: Not required any more because "terms" has been moved to drawer menu

packages/sanes-chrome-extension/src/routes/account/assets/requests.svg
```diff
@@ -1,3 +1,3 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="18" height="16" fill="none" viewBox="0 0 18 16">
-    <path fill="#31E6C9" d="M1.714 0C1.246 0 .8.165.482.482.165.795 0 1.242 0 1.714v1.143a.57.57 0 0 0 .571.572h1.715v8.428c0 .942.696 1.857 1.714 1.857h3.728A5.148 5.148 0 0 0 12 16a5.15 5.15 0 0 0 5.143-5.143c0-2.232-1.438-4.116-3.429-4.83V1.714c0-.473-.165-.92-.482-1.232-.299-.299-.714-.46-1.156-.473A.393.393 0 0 0 12 0H1.714zm0 1.143h8.84c-.103.24-.268.45-.268.714v.429H1.143v-.572c0-.218.062-.339.147-.424.085-.085.21-.147.424-.147zm10.286 0c.214 0 .34.062.424.147.085.085.147.206.147.424v4.058c-.191-.022-.375-.058-.571-.058-1.21 0-2.313.438-3.188 1.143h-4.24V8h3.191a5.34 5.34 0 0 0-.593 1.143H4.57v1.143h2.344c-.022.192-.058.375-.058.571 0 .603.12 1.174.313 1.714H4c-.353 0-.571-.276-.571-.714V3.43h7.428a.57.57 0 0 0 .572-.572v-1c0-.437.218-.714.571-.714zM4.571 4.57v1.143h5.715V4.571H4.57zM12 6.857c2.214 0 4 1.786 4 4s-1.786 4-4 4-4-1.786-4-4 1.786-4 4-4zM11.429 8v2.286H9.143v1.143h2.286v2.285h1.142V11.43h2.286v-1.143h-2.286V8H11.43z"/>
+<svg xmlns="http://www.w3.org/2000/svg" width="15" height="16" fill="none" viewBox="0 0 15 16">
```
webmaster128: Why is this new icon 15x16 instead of 18x16 (like the old one and like terms.svg)?

packages/sanes-chrome-extension/src/routes/account/index.tsx
```diff
@@ -14,7 +14,9 @@ import { createAccount } from '../../utils/chrome';
 import { ACCOUNT_STATUS_ROUTE, RECOVERY_PHRASE_ROUTE, REQUEST_ROUTE } from '../paths';
 import recoveryPhrase from './assets/recoveryPhrase.svg';
 import requests from './assets/requests.svg';
+import terms from './assets/terms.svg';
 import ListTxs from './components/ListTxs';
+import Link from 'medulas-react-components/lib/components/Link';
```
webmaster128: Is this new import used?

packages/sanes-chrome-extension/src/routes/account/assets/requests.svg
webmaster128: Hmm, icon width seems to be inconsistent. In https://github.com/iov-one/ponferrada/pull/352 we have 16x16. Is this a problem? Are they horizontally centered in some box?

albertandrejev: No, I just replaced old icon with new one, because old icon was used for T&C menu item (like it was in `wallet-demo`.

packages/sanes-chrome-extension/src/routes/account/index.tsx
albertandrejev: You are right, it was not used.

packages/sanes-chrome-extension/src/routes/account/assets/requests.svg
albertandrejev: Icon sizes are the same as in the `wallet-demo`

PR #354, opened 2019-07-02
Feature #353 - Add T&C to the extension signup process

### LINKED ISSUE #353: Add a check box on sign up for T&Cs
As a user I need to tick I have read and agree with the T&C on the registration page.

- [x] check box
- [x] link to T&C with "I have read and agree the T&C"
- [x] Disable the register button until the box has been ticked.

albertandrejev:
**Description**
This PR will add T&C checkbox and link to the extension signup process. Closes #353.

**Before**
new-account-before

**After**
new-account-after

### DIFF
--- a/packages/medulas-react-components/src/components/forms/CheckboxField/index.tsx
+++ b/packages/medulas-react-components/src/components/forms/CheckboxField/index.tsx
@@ -3,9 +3,15 @@ import { useField } from 'react-final-form-hooks';
 import * as React from 'react';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import Checkbox, { CheckboxProps } from '@material-ui/core/Checkbox';
-import { Omit } from '@material-ui/core';
+import { Omit, FormControl, FormHelperText, makeStyles, Theme } from '@material-ui/core';
 import CheckBoxOutlineBlankIcon from '@material-ui/icons/CheckBoxOutlineBlank';
 import CheckBoxIcon from '@material-ui/icons/CheckBox';
+
+const useStyles = makeStyles<Theme>({
+  root: {
+    marginTop: 0,
+  },
+});
 
 interface Props extends Omit<CheckboxProps, 'form'> {
   readonly fieldName: string;
@@ -17,8 +23,10 @@ interface Props extends Omit<CheckboxProps, 'form'> {
 }
 
 const CheckboxField = ({ fieldName, form, initial, onChangeCallback, label }: Props): JSX.Element => {
-  const { input } = useField(fieldName, form);
+  const classes = useStyles();
+  const { input, meta } = useField(fieldName, form);
 
+  const error = meta.error && (meta.touched || !meta.pristine);
   const { name, onChange, value, ...restInput } = input;
   const inputProps = { name, ...restInput };
 
@@ -49,7 +57,12 @@ const CheckboxField = ({ fieldName, form, initial, onChangeCallback, label }: Pr
     />
   );
 
-  return <FormControlLabel control={control} label={label} />;
+  return (
+    <FormControl error={error}>
+      <FormControlLabel control={control} label={label} />
+      {error && <FormHelperText className={classes.root}>{meta.error}</FormHelperText>}
+    </FormControl>
+  );
 };
 
 export default CheckboxField;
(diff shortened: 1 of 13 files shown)


packages/sanes-chrome-extension/src/routes/requests/index.stories.tsx
```diff
@@ -26,8 +27,8 @@ const intialRequests: Request[] = [
       senderUrl: 'Sender 2',
       requestedIdentities: [],
     },
-    accept: jest.fn(),
```
albertandrejev: Usage of `jest.fn()` breaks storybook.

packages/sanes-chrome-extension/src/routes/signup/test/operateSignup.ts
```diff
@@ -82,13 +98,13 @@ export const submitNewAccount = async (
   accountName: string,
   password: string,
 ): Promise<void> => {
-  const inputs = TestUtils.scryRenderedDOMComponentsWithTag(newAccountDom, 'input');
-  expect(inputs.length).toBe(3);
```
albertandrejev: This check is not required because we are checking amount of fields in another test.

packages/sanes-chrome-extension/src/routes/signup/index.dom.spec.ts
```diff
@@ -59,30 +61,32 @@ describe('DOM > Feature > Signup', () => {
...
+      [accountNameInput, passwordInput, passwordConfirmInput, termsAcceptField] = newAccountInputs;
       newAccountForm = getNewAccountForm(signupDom);
       buttons = TestUtils.scryRenderedDOMComponentsWithTag(signupDom, 'button');
       [backButton, continueButton] = buttons;
     });
 
-    it('has three inputs', () => {
-      expect(newAccountInputs.length).toBe(3);
+    it('has four inputs', () => {
+      expect(newAccountInputs.length).toBe(4);
```
albertandrejev: We are checking for amount of inputs here.

packages/medulas-react-components/src/components/forms/CheckboxField/index.tsx
```diff
@@ -3,22 +3,30 @@ import { useField } from 'react-final-form-hooks';
 import * as React from 'react';
 import FormControlLabel from '@material-ui/core/FormControlLabel';
 import Checkbox, { CheckboxProps } from '@material-ui/core/Checkbox';
-import { Omit } from '@material-ui/core';
+import { Omit, FormControl, FormHelperText, makeStyles, Theme } from '@material-ui/core';
```
apanizo: You can remove this Omit and just use the default one which comes in TS 3.5

packages/sanes-chrome-extension/src/routes/requests/index.stories.tsx
apanizo: 🤔 I can figure out the reason but the question is... why suddenly the usage of `jest.fn()` breaks storybook and not before? We created this class long time ago.

Any idea?

packages/sanes-chrome-extension/src/routes/signup/components/NewAccountForm.tsx
```diff
@@ -10,10 +10,15 @@ import Form, {
...
 import { SIGNUP_ROUTE } from '../../paths';
+import CheckboxField from 'medulas-react-components/lib/components/forms/CheckboxField';
+import Typography from 'medulas-react-components/lib/components/Typography';
+import Link from 'medulas-react-components/lib/components/Link';
 
 export const ACCOUNT_NAME_FIELD = 'accountNameField';
 export const PASSWORD_FIELD = 'passwordInputField';
 export const PASSWORD_CONFIRM_FIELD = 'passwordConfirmInputField';
+export const TERMS_ACCEPT_FIELD = 'termsAcceptCheckboxField';
+export const TERMS_URL = 'https://support.iov.one/hc/en-us';
```
apanizo: I'd say we should put this on `src/routes/paths`, maybe near soon in the future, we create our own view/path.

packages/sanes-chrome-extension/src/routes/requests/index.stories.tsx
albertandrejev: Storybook continue working but some of the stories become unavailable, especially in `Extension` part.

PR #357, opened 2019-07-03
Feature #355 – Add footer to drawer with T&C link in it

### LINKED ISSUE #355: Convert Terms & Conditions button into link
As discussed in https://github.com/iov-one/ponferrada/pull/352#issuecomment-507692187

webmaster128:
Closes #355

**Before:**

352-after

**After:**

Bildschirmfoto 2019-07-02 um 16 02 29


packages/sanes-chrome-extension/src/routes/account/index.tsx
```diff
@@ -81,15 +81,20 @@ const AccountView = (): JSX.Element => {
...
-      icon: termsIcon,
-      text: 'Terms & Conditions',
-      action: () => window.open('https://support.iov.one/hc/en-us', '_blank'),
-    },
   ];
 
+  const footer = (
+    <Block marginTop={4} textAlign="center">
+      <Block marginBottom={1}>
+        <Typography variant="subtitle2" inline>
```
albertandrejev: We usually uses `Typography` as child of the `Link` component, also we have `link` prop for `Typography`.

webmaster128: I changed this because in the storybook, the `<a>` and `<h6>` end up with different colors.

Bildschirmfoto 2019-07-03 um 16 18 47

However, this seems to be a storybook issue only. Will update for consistency with https://github.com/iov-one/ponferrada/blob/d3da632/packages/sanes-chrome-extension/src/routes/login/components/LoginControls.tsx

albertandrejev: you can use `<Typography .... inline link color="primary">....</Typography>`

webmaster128: This is what is done in LoginControls, where I saw the color issue above.

But I found the issue: I open the storybook in Firefox and the extension in Chrome. Firefox uses the `<a>` color for the line (default link blue), Chrome uses the `<h6>` color for the line.

One more thing understood :)

packages/medulas-react-components/src/components/Drawer/index.tsx
```diff
@@ -134,23 +135,26 @@ function PersistentDrawerRight({ children, items, elevation = 0 }: Props): JSX.E
...
+                {item.icon && (
+                  <ListItemIcon>
+                    <Image src={item.icon} alt="Menu icon" />
+                  </ListItemIcon>
+                )}
+                <ListItemText primary={item.text} onClick={item.action} />
+              </ListItem>
+            ))}
+          </List>
+          {footer && <footer>{footer}</footer>}
```
apanizo: Conceptually this is not entirely correct, footer normally is used to specify a region of webapp, normally at the bottom of the view. In this case, Drawer is a component which has a vertical space, where, in our case, we place links (with an icon or without it) inside it. Seeing the screenshots I do not see the connection with a footer.

From a user perspective, I would not be happy if I use a Drawer and it has a... footer. 

Anyway do not change it, it works, and it works fine.

packages/medulas-react-components/src/components/Link/index.tsx
```diff
@@ -8,8 +8,14 @@ interface Props extends LinkProps {
 const Link = ({ children, to, ...rest }: Props): JSX.Element => {
   if (typeof to === 'string') {
     if (/^((mailto:)|((https?:)?\/\/))/.test(to)) {
+      // We use href and target in the <a> tag for styling and hover efects.
+      // The actual click event is handled via JavaScript to allow testing those clicks
+      const clickHandler = (event: React.MouseEvent): void => {
```
apanizo: This is interesting, I perfectly remember to be able to test <a> links in jest simulating the click properly, why do we need this?

webmaster128: Because a click on an `<a href` does not emit a JavaScript `click` event, which can be mocked and tested. The only thing you can test is if the page changed to the desired url. But in this case, the URL is opened in a different window we have no access to.

I'd be very happy to revert and simplify if you show me a way how to test it. I searched the web for half an our or so and all the results test by checking `window.location.href`, which does not change when opening in a new window.

packages/medulas-react-components/src/components/Drawer/index.tsx
webmaster128: Hmm, a drawer is the whole element that slides into the screen, which gave it the name:
https://en.wikipedia.org/wiki/Drawer_(furniture)#/media/File:Drawer.agr.jpg

The material design guide describes headers inside of Drawers, but not explicitely footers. So even if not the most typical thing, I don't see what is wrong with having a footer analogue to a header. Regarding the HTML `<footer>` element, I don't care much which element is used. I think it is pretty useful for element selections in tests but it could be changed easily.

packages/medulas-react-components/src/components/Link/index.tsx
apanizo: Which is the faulty test? I need to understand the context and what you are trying to test.

webmaster128: 'redirects to the Terms and Conditions page'

packages/medulas-react-components/src/components/Drawer/index.tsx
apanizo: The material design guidelines say a header is allowed if for any reason the drawer's location collides with a real header in the app. In this header region you can place your logo for example.

But it does not say you can use the <header> tag. In fact, if you use them you are committing SEO penalty, and lower page rating.

Same applies to the footer, there is absolutely no reference about it, the closest reference is [this](https://www.dropbox.com/s/zleo52p992zn1ep/Screenshot%202019-07-04%2013.45.01.png), and it talks navigation bar.

I do not want you to expend time to justify your decisions, I was off, so I was not able to review this PR, just wanted to leave a comment just in case we need to do something else here, to remember it, and refactor. Just that. 

Good job on the FE code btw.

Issue #336, opened 2019-06-26
Introduce tickers in store


apanizo:
- [x] Added harcoded tokens to store
- [x] Include chains config in project
- [x] Inject properly available tokens

This ticket will introduce all tickers from supported chains inside redux, for making the information available in the app (especially balance and send payment route).

The structure will follow:
```
{  tickers: 
      "BOV": {
           chainId: '...'
           ticker: { fractionalDigits, tokenTicker, tokenName}
     }
     ....
} 
```

webmaster128:
Closing since done

PR #358, opened 2019-07-04
Feature #336 Introduce tokens in store

apanizo:
**Description**
This is an initial PR for adding tokens of chains to redux store. 

**Note for developers**
The next PR will introduce the chains config for adding all tokens from networks we support, and finally, add them after login.


packages/bierzo-wallet/src/store/extension/actions.ts
```diff
@@ -4,6 +4,6 @@ export const setExtensionStateAction = (
   connected: boolean,
   installed: boolean,
 ): SetExtensionStateActionType => ({
-  type: '@@extension/SET_EXTENSION_STATE',
```
albertandrejev: My idea was to use prefix to describe action relation, so `@@extension/SET_EXTENSION_STATE` means that this action relates to `extensions` state. I think that `EXTENSION part in the `SET_EXTENSION_STATE` name is redundant in this case and it is better to call this action something like `@@extension/SET_STATE`.

packages/bierzo-wallet/src/store/tokens/actions.ts
```diff
@@ -0,0 +1,24 @@
...
+  for (const chainToken of chainTokens) {
+    const ticker = chainToken.tokenTicker as string;
+    tokens[ticker] = { chainId, token: chainToken };
+  }
+
+  return tokens;
+}
+
+export const addTickersAction = (tokens: { [key: string]: BwToken }): AddTickerActionType => ({
+  type: '@@bw/ADD_TOKENS',
```
albertandrejev: So if we will accept my notation this action should be named `@@tokens/ADD_TOKENS` etc.

abefernan: But following your `@@extension/SET_STATE` suggestion, shouldn't it be `@@tokens/ADD`?

albertandrejev: Yes, you are right.

packages/bierzo-wallet/src/store/extension/actions.ts
apanizo: //cc @abefernan 

I see, normally I use redux to describe the use storage of the whole app, that is why I use @@bw (bierzo-wallet).

I totally understand your point and to be honest I do not mind to adapt myself to your way of doing it. 

I need to create a final PR where I call this action after login. In that PR I will update the code based on your remarks. So feel free to approve this PR. :wink:

PR #359, opened 2019-07-04
Feature - Add chains config

apanizo:
**Description**
This is the second step for adding all tokens to store. Now we get connections from all supported chains.

**Note for developers**
The final step is to use everything inside the application, which is going to happen in a separate PR.


packages/bierzo-wallet/src/store/tokens/actions.ts
```diff
@@ -1,18 +1,45 @@
...
+  return null;
+}
+
+export async function getTokens(): Promise<{ [ticker: string]: BwToken }> {
+  const config = await loadConfig();
   const tokens: { [ticker: string]: BwToken } = {};
+  const chains = config.chains;
+
+  for (const chain of chains) {
+    const connection = await createConnectionFor(chain.chainSpec);
```
webmaster128: The `connection` created here needs to be disconnected by calling `connection.disconnect()` when done. It would be better to keep the connects alive and reuse them later

packages/bierzo-wallet/src/store/tokens/actions.ts
```diff
@@ -1,18 +1,45 @@
...
+
+  if (spec.codecType === 'bns') {
+    return BnsConnection.establish(url);
+  }
+
+  if (spec.codecType === 'lsk') {
+    return LiskConnection.establish(url);
+  }
+
+  return null;
```
webmaster128: This case indicates a programming error (got `spec` we do not support), so throwing an exception is better than returning null because the caller code does not need to deal with this case.

packages/bierzo-wallet/src/store/tokens/actions.ts
```diff
@@ -1,18 +1,45 @@
...
+
+export async function getTokens(): Promise<{ [ticker: string]: BwToken }> {
+  const config = await loadConfig();
   const tokens: { [ticker: string]: BwToken } = {};
+  const chains = config.chains;
+
+  for (const chain of chains) {
+    const connection = await createConnectionFor(chain.chainSpec);
+    if (!connection) {
+      continue;
```
webmaster128: At this point, a programming error is silently ignored (as commented in createConnectionFor)

packages/bierzo-wallet/src/logic/connection.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  if (spec.codecType === 'lsk') {
+    return getLiskConnection(url);
+  }
+
+  throw new Error('Chain spec not supported');
+}
+
+export async function disconnect(): Promise<void> {
+  const config = await loadConfig();
```
apanizo: @webmaster128 where in the code we should call this? As far as I understand we rely on WS connection, and if the browser is closed, automatically the connection is cleared.

Is this related to some kind of possible logout action?

webmaster128: Right. At the moment you don't need to call it at all. If we lazy load the connections and keep them alive forever, we are good. Maybe we have a good reason to disconnect and reconnect later on, but at the moment I don't see one.

The problem with the previous code was that every call of `getTokens()` created new connections which were not disconnected, leading to 3, 6, 9, 12, … open connections. This issue does not exist anymore.

PR #360, opened 2019-07-04
Feature #349 - Username registration request component

### LINKED ISSUE #349: Implement ReqRegisterUsernameTx React component
We now have multiple transaction types and need one component per type. 

A stub ReqRegisterUsernameTx already exists. It needs to be updated to display all the important data of the transaction.

albertandrejev:
**Description**
This PR will add username registration request transaction view to `tx-request` route of the extension. Closes #349.

**Before**
N/A

**After**
username-reg

### DIFF
--- a/packages/sanes-chrome-extension/src/routes/tx-request/components/ShowRequest/ReqRegisterUsernameTx.tsx
+++ b/packages/sanes-chrome-extension/src/routes/tx-request/components/ShowRequest/ReqRegisterUsernameTx.tsx
@@ -1,22 +1,11 @@
 import { RegisterUsernameTx } from '@iov/bns';
 import { List, ListItem, ListItemText } from 'medulas-react-components/lib/components/List';
 import * as React from 'react';
-import { makeStyles } from '@material-ui/core';
-import { amountToString } from '../../../../utils/balances';
 import Typography from 'medulas-react-components/lib/components/Typography';
 import Block from 'medulas-react-components/lib/components/Block';
+import TransactionFee, { useTxListItemStyles, txListItemSecondaryProps } from './TransactionFee';
 
 export const REQ_REGISTER_USERNAME = 'req-register-username-tx';
-
-const useStyles = makeStyles({
-  root: {
-    margin: 0,
-  },
-});
-
-const secondaryProps = {
-  noWrap: true,
-};
 
 interface Props {
   readonly creator: string;
@@ -24,22 +13,9 @@ interface Props {
 }
 
 const ReqRegisterUsernameTx = ({ tx }: Props): JSX.Element => {
-  const listItemClasses = useStyles();
+  const listItemClasses = useTxListItemStyles();
 
-  const txFee =
-    tx.fee && tx.fee.tokens ? (
-      <React.Fragment>
-        <Block marginTop={1} />
-        <Typography variant="body1" inline>
-          Fee:{' '}
-        </Typography>
-        <Typography variant="body1" inline>
-          {amountToString(tx.fee.tokens)}
-        </Typography>
-      </React.Fragment>
-    ) : (
-      undefined
-    );
+  const txFee = tx.fee ? <TransactionFee fee={tx.fee} /> : undefined;
 
   return (
     <React.Fragment>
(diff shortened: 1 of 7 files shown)


packages/sanes-chrome-extension/src/routes/tx-request/components/ShowRequest/ReqRegisterUsernameTx.tsx
```diff
@@ -1,16 +1,74 @@
...
+        <Typography variant="body1" inline>
+          Fee:{' '}
+        </Typography>
+        <Typography variant="body1" inline>
+          {amountToString(tx.fee.tokens)}
+        </Typography>
+      </React.Fragment>
+    ) : (
+      undefined
+    );
```
webmaster128: Nice. Can we deduplicate this code block to use th same code in `ReqRegisterUsernameTx` and `ReqSendTransaction`?

packages/sanes-chrome-extension/src/routes/tx-request/components/ShowRequest/ReqRegisterUsernameTx.tsx
```diff
@@ -1,16 +1,74 @@
...
+        username registration request.
+      </Typography>
+      <Block marginTop={1} />
+      <Typography variant="body1">Addresses:</Typography>
+      <List>
+        {tx.addresses.map(address => (
+          <ListItem key={address.address}>
+            <ListItemText
+              classes={listItemClasses}
+              primary={address.chainId}
```
webmaster128: @apanizo already created a chain ID -> chain name mapping (configured in config files). Can we easily use this here, showing the chain name? If that is too complicated, feel free to postpone for after the MvP

packages/sanes-chrome-extension/src/routes/tx-request/components/ShowRequest/ReqRegisterUsernameTx.tsx
albertandrejev: I used different code for transaction fee in username as opposed to the send transaction. Because I used `List` component here to show addresses only. Take a look at screenshot. Anyway I can use same approach for send transaction too.

webmaster128: That would be very useful, since very soon we need to be able to create large amounts of those components quickly. The more we can deduplicate, the better.

packages/sanes-chrome-extension/src/routes/tx-request/components/ShowRequest/ReqRegisterUsernameTx.tsx
albertandrejev: @apanizo can you provide any code example how to use this in react component.

packages/sanes-chrome-extension/src/routes/tx-request/components/ShowRequest/ReqRegisterUsernameTx.tsx
webmaster128: Okay, now that the code is very similar for RegisterUsername and Send, can we have a separate `TransationFee` component, that just takes a `fee?: Fee` property, where type `Fee` is from `@iov/bcp`?

Such that we just write

```
<TransationFee fee={tx.fee} />
```

in this file and in ReqSendTransaction.tsx.

In the case that `tx.fee` is undefined, this can just be an empty compnent. No need to optimize for this case since it will not happen in reality

packages/sanes-chrome-extension/src/routes/tx-request/components/ShowRequest/ReqRegisterUsernameTx.tsx
```diff
@@ -1,16 +1,72 @@
...
+  const txFee =
+    tx.fee && tx.fee.tokens ? (
+      <React.Fragment>
+        <Block marginTop={1} />
+        <Typography variant="body1" inline>
+          Fee:{' '}
+        </Typography>
+        <Typography variant="body1" inline>
+          {amountToString(tx.fee.tokens)}
+        </Typography>
```
webmaster128: Why two `<Typography>` elements? What about

```tsx
        <Typography variant="body1" inline>
          {'Fee: ' + amountToString(tx.fee.tokens)}
        </Typography>
```

albertandrejev: Yes, you are right, my mistake. Also `inline` attribute is not required in this case.

PR #365, opened 2019-07-08
Feature #337 Introduce balances in store

### LINKED ISSUE #337: Introduce balances in store
This ticket will introduce all positive balances from supported chains inside redux, for making the information available in the app (especially balance and send payment route).

The structure will follow:
```
{  
    balances: {
        "IOV": "100000000",
         ....
    }
} 
```

apanizo:
**Description**
In this PR we store in redux balances. If blockchain accounts have 0 balance they are not injected, only positive balances are shown.

For a quick understanding check the unit tests.

This close #337

### DIFF
--- a/packages/bierzo-wallet/src/routes/login/index.tsx
+++ b/packages/bierzo-wallet/src/routes/login/index.tsx
@@ -5,6 +5,7 @@ import * as React from 'react';
 import * as ReactRedux from 'react-redux';
 
 import { history } from '..';
+import { addBalancesAction, getBalances } from '../../store/balances';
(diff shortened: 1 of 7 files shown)


packages/bierzo-wallet/src/routes/login/index.tsx
```diff
@@ -34,6 +36,10 @@ const Login = (): JSX.Element => {
     const chainTokens = await getTokens();
     dispatch(addTickersAction(chainTokens));
```
apanizo: Here I get the keys from the store and I use them for getting positive account balances.

Issue #276, opened 2019-06-10
Balance page : Migrate balance page from wallet-demo


albertandrejev:
Migrate balance page from wallet-demo project and adapt page layout and redux configuration to ponferrado monorepo requirements.

- [x] Migrate PageMenu layout component (#366)
- [x] Migrate layout (#371)
- [x] Adapt layout to the ponferrado requirements (#371)
- [x] Create view storybook (#371)
- [ ] Add DOM and E2E tests (#382)
- [ ] Connect with redux (#381)

PR #366, opened 2019-07-08
Feature #276 - Migrate PageMenu component

albertandrejev:
**Description**
This PR will add PageMenu component to `bierzo-wallet`. It will implement "Migrate PageMenu layout component" part of the ticket #276.

page-menu-component

### DIFF
--- a/packages/bierzo-wallet/src/components/PageMenu/index.tsx
+++ b/packages/bierzo-wallet/src/components/PageMenu/index.tsx
@@ -8,12 +8,11 @@ import { RouteComponentProps, withRouter } from 'react-router';
 import Header from '../Header';
 
 interface Props extends RouteComponentProps {
-  readonly renderProps?: () => React.ReactNode;
   readonly children?: React.ReactNode;
   readonly padding?: boolean;
 }
 
-const PageMenu = ({ children, padding = true, renderProps, location }: Props): JSX.Element => {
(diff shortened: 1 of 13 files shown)


packages/bierzo-wallet/src/components/PageMenu/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+const PageMenu = ({ children, padding = true, renderProps, location }: Props): JSX.Element => {
+  const theme = useTheme<Theme>();
+
+  return (
+    <Block display="flex" flexDirection="column" bgcolor={theme.palette.background.default}>
+      <Header path={location.pathname} />
+      <ErrorBoundary type="inner">
+        <Block display="flex" flexDirection="column" padding={padding ? 4 : undefined}>
+          {renderProps !== undefined ? renderProps() : children}
```
apanizo: Why do we need renderProps and children now? Why not just children?

packages/bierzo-wallet/src/theme/globalStyles.ts
```diff
@@ -20,6 +21,7 @@ export const globalStyles = makeStyles({
       padding: '0',
       fontFamily: '"Muli", sans-serif',
       boxSizing: 'border-box',
+      backgroundColor: backgroundColor,
```
apanizo: you can use the new syntax:
```
{
    ....
    backgroundColor,
}
```

PR #373, opened 2019-07-10
Fix #369 load config at runtime

### LINKED ISSUE #369: Don't rely on NODE_ENV for choosing configuration
We need to be able to switch between those configuations:

* `development.json`: connect to local blockchains
* `staging.json`: connect to Testnets
* `production.json`: connect to Mainnets

In 2e2 tests, `development.json` should be used.

Right now, the code relies on [NODE_ENV](https://facebook.github.io/create-react-app/docs/adding-custom-environment-variables) which is always "production" when `react-scripts build` is used. The value does not change at runtime anymore, such that `production.json` is always used.

This affects e2e tests because they use the build from:

```
app.use(express.static(require('path').join(__dirname, '/../../../build')));
```

apanizo:
**Description**
In this PR we change the way we load configuration in bierzo-wallet project. Now we load it at build time using a custom node environment var.

The problem was for executing the e2e tests we had to compile the application creating its build, and for doing it, react-scripts used by default the `production` value, and therefore we injected the production configuration in the build.

Now, we use a separate and unique node_env making it possible to generate the build with our chosen config (dev, staging and production).

It closes #369 

**Note for developers**
Using require when load config takes advantage of webpack optimization just including the desired config file 🎉


apanizo:
Just seen your message right now 😅. Good job!

PR #376, opened 2019-07-11
Feature #320 - Dashboard proposal list testing

### LINKED ISSUE #320: Dashboard structure: List of Proposals

abefernan:
This PR adds a DOM test suite for the Dashboard route of the Voting App. It also adds some HTML ids to production code to make testing easier.

Closes #320.

### DIFF
--- a/packages/sil-voting-app/src/routes/dashboard/index.dom.spec.ts
+++ b/packages/sil-voting-app/src/routes/dashboard/index.dom.spec.ts
@@ -1,0 +1,87 @@
+import { Store } from 'redux';
+
+import { aNewStore } from '../../store';
+import { RootState } from '../../store/reducers';
+import { click } from '../../utils/test/dom';
+import {
+  checkActiveProposal,
+  checkEndedProposal,
+  checkVoteButtons,
+  getAsideFilter,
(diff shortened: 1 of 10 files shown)


packages/sil-voting-app/src/routes/dashboard/index.dom.spec.ts
```diff
@@ -0,0 +1,87 @@
+import { Store } from 'redux';
```
apanizo: Testing speaking you are doing correctly here. I mean, you are testing everything that should appear in the view, including one of each type of proposals.

What I am not really convinced is the way of proposals are created. In a separate ticket I recommend to create the proposal reducer, and then, in those tests taking advantage of the proposals creator you have created, dispatch them to redux, and then, check how the proposals view gets them and render exactly the same.

The ideal workflow is:
```
1. You create your test having in mind you need to push into views
2. Start with the reducer (sometimes you need to unit test  small clever function outside react)
3. Create the views
4. Check it all
```

PR #377, opened 2019-07-15
Use es6 style imports for config

webmaster128:


webmaster128:
This by the way adds type safety, ensuring that the content of the .json files matches the `Config` type.

webmaster128:
> I thought you confirmed using the import ... made webpack to import the three versions of config files.

It turned out that the 3 imports were included before because webpack did not optimize the branches in the `const configuration = (): Config => {` function using the old variable. I am surprised how clever the webpack optimizer is.

apanizo:
I told you. Not only webpack allowing this clever tree shaking, but the super optimized create-react-app -> react-scripts config which are the real magicians here.

PR #381, opened 2019-07-15
Feature #276 - Connect Balance route with redux

albertandrejev:
**Description**
This PR will connect Balance route component with redux store to show real balance data. It will complete "Connect with redux" step of the issue #276.

**With redux store connected**
balance-view-redux

### DIFF
--- a/packages/bierzo-wallet/src/routes/balance/components/index.tsx
+++ b/packages/bierzo-wallet/src/routes/balance/components/index.tsx
@@ -1,4 +1,3 @@
-import { Amount } from '@iov/bcp';
 import { makeStyles, Theme } from '@material-ui/core';
 import { useTheme } from '@material-ui/styles';
 import Block from 'medulas-react-components/lib/components/Block';
@@ -9,13 +8,14 @@ import Tooltip from 'medulas-react-components/lib/components/Tooltip';
 import Typography from 'medulas-react-components/lib/components/Typography';
 import React from 'react';
 
+import { BalanceState } from '../../../store/balances';
 import { amountToString, trimAmount } from '../../../utils/balances';
 import receive from '../assets/transactionReceive.svg';
 import send from '../assets/transactionSend.svg';
 
 interface Props {
   readonly name: string | undefined;
-  readonly tokens: ReadonlyArray<Amount>;
+  readonly balance: BalanceState;
   readonly onSendPayment: () => void;
   readonly onReceivePayment: () => void;
 }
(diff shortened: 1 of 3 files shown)


packages/bierzo-wallet/src/routes/balance/components/index.tsx
```diff
@@ -9,13 +8,14 @@ import Tooltip from 'medulas-react-components/lib/components/Tooltip';
...
 
+import { BalanceState } from '../../../store/balances';
 import { amountToString, trimAmount } from '../../../utils/balances';
 import receive from '../assets/transactionReceive.svg';
 import send from '../assets/transactionSend.svg';
 
 interface Props {
   readonly name: string | undefined;
-  readonly tokens: ReadonlyArray<Amount>;
+  readonly balance: BalanceState;
```
apanizo: Do we want this? I know they are the same type, but maybe we can have the redux type and do not change the balance attr here in this component. 

I suggest to write the type instead of reusing the redux one. What do you think? In this way, we keep separated components.

packages/bierzo-wallet/src/routes/login/index.e2e.spec.ts
```diff
@@ -42,69 +42,70 @@ withChainsDescribe(
...
-      await getBackgroundPage(browser);
-      await page.bringToFront();
-      await sleep(1000);
-
-      await page.click('button');
-      await sleep(500);
-
-      await checkLoginMessage(page);
-    }, 45000);
+    withChainsDescribe(
```
apanizo: What has happened here? Just format changes?

albertandrejev: Please read comment at https://github.com/iov-one/ponferrada/pull/381/files/e53539aa4986c0f86137964ede5dbeb36d45f7a0#diff-27099a180ad72ea232102aadb726e78fR110

packages/bierzo-wallet/src/routes/login/index.e2e.spec.ts
```diff
@@ -42,69 +42,70 @@ withChainsDescribe(
...
+          await page.click('button');
+          await sleep(500);
+
+          await checkLoginMessage(page);
+        }, 45000);
+      },
+    );
 
     it('shows install IOV extension message', async (): Promise<void> => {
       browser = await launchBrowser(0, false);
```
albertandrejev: Because we are launching browser here we should close previous instance of browser, so I was required to refactor tests suite a little and use different blocks for launching browser.

packages/bierzo-wallet/src/routes/balance/components/index.tsx
albertandrejev: Sorry do not understand your suggestion. Do you want me to create something like:
```
export interface BalanceType {
  [token: string]: Amount;
}
```
And use it as prop type? Or do you want to me cast (convert) `BalanceState` to `ReadonlyArray<Amount>` ?

apanizo: Just use: `tokens: [token: string]: Amount;` :)

PR #385, opened 2019-07-16
Feature #384 - Add account's balances subscriptions

### LINKED ISSUE #384: Account's balances subscriptions
We need to implement account subscriptions in our stack for being aware of txs changes, and therefore make easier to receive tokens in order to use the wallet (sending payments, buy NFT’s names…)

apanizo:
**Description**
In this PR I include the subscription for blockchain account balances. Now each time we receive an event we call getBalances function and we update the redux store. 

For achieving it I had to change the http connection in bns chain to ws.

It closes #384 

**Note for developers**
The way I have used for testing functionality is calling faucet twice, forcing accounts to receive events.

### DIFF
--- a/packages/bierzo-wallet/src/logic/connection.unit.spec.ts
+++ b/packages/bierzo-wallet/src/logic/connection.unit.spec.ts
@@ -21,6 +21,7 @@ withChainsDescribe('Logic :: connection', () => {
     await getConnectionFor(firstChain.chainSpec);
 
     expect(ethereumConnectionSpy).toHaveBeenCalledTimes(1);
+    await disconnect();
   });
 
   it('reset connections correctly when disconnecting', async () => {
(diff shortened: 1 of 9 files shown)


packages/bierzo-wallet/src/logic/connection.unit.spec.ts
```diff
@@ -17,6 +17,7 @@ withChainsDescribe('Logic :: connection', () => {
     await getConnectionFor(firstChain.chainSpec);
 
     expect(ethereumConnectionSpy).toHaveBeenCalledTimes(1);
+    await disconnect();
```
webmaster128: Why not move this in `afterEach` (disconnect once oer file) or `afterAll` (disconnect once per spec) like in other files? Makes it harder to forget when new tests come in

packages/bierzo-wallet/src/logic/subscriptions.ts
```diff
@@ -0,0 +1,47 @@
...
+      },
+    });
+    balanceSubscriptions.push(subscription);
+  }
+
+  // subscribe to transactions
+  // const transactionsStream = connection.liveTx({ sentFromOrTo: address });
+}
+
+export function unsubscribeBalances(): void {
```
webmaster128: 👍

packages/bierzo-wallet/src/logic/subscriptions.unit.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    // Trick for forcing account to receive balance events updates
+    await drinkFaucetIfNeeded(keys, chainTokens);
+
+    // Give some time to open request to be finished
+    await sleep(5000);
+
+    expect(balanceSpy).toHaveBeenCalledTimes(5);
+
+    unsubscribeBalances();
+    await disconnect();
```
webmaster128: Why not move this in `afterEach` (disconnect once oer file) or `afterAll` (disconnect once per spec) like in other files? Makes it harder to forget when new tests come in

PR #386, opened 2019-07-16
Feature #378 - Dashboard proposal redux

### LINKED ISSUE #378: Dashboard structure: Redux integration for Proposals
Closes #279.

As suggested by @apanizo [in a comment](https://github.com/iov-one/ponferrada/pull/376#discussion_r303307028):

Implement Redux reducer to get and create proposals for production and testing.

abefernan:
This PR integrates the Proposals with Redux to better manage its state.
It uses statically declared proposals since the integration with the Governance API is not done yet.

Closes #378.


packages/sil-voting-app/src/routes/dashboard/components/ProposalsList.tsx
```diff
@@ -1,72 +1,40 @@
...
-    threshold: 20,
-    result: { yes: 15, no: 4, abstain: 6 },
-    vote: 'Abstain',
-    status: 'Ended',
-  },
-];
+const renderProposals = (storedProposals: ProposalsState): JSX.Element[] => {
+  const proposals = Object.values(storedProposals).map(state => {
+    return state.proposal;
+  });
```
albertandrejev: You can simplify this `Object.values(storedProposals).map(state => state.proposal);`

packages/sil-voting-app/src/routes/dashboard/components/ProposalsList.tsx
```diff
@@ -1,72 +1,40 @@
...
 
-const ProposalsList = (): JSX.Element => {
-  const proposals = props.map((_, index) => {
+  return proposals.map((_, index) => {
```
albertandrejev: Why are you running map twice? Also I would like to recommend you to move this method to separate component and call it as React component (see comment below).

packages/sil-voting-app/src/routes/dashboard/components/ProposalsList.tsx
```diff
@@ -1,72 +1,40 @@
...
+};
+
+const ProposalsList = (): JSX.Element => {
+  const storedProposals = ReactRedux.useSelector((state: RootState) => state.proposals);
+  const renderedProposals = renderProposals(storedProposals);
 
   return (
     <Block id={PROPOSALS_HTML_ID} flexGrow={1}>
-      {proposals}
+      {renderedProposals}
```
albertandrejev: You can create component to render proposals and call it like `{storedProposals.map((proposal, index) => <RenderedProposal proposal={proposal} index={index} />)}`

packages/sil-voting-app/src/routes/dashboard/components/ProposalsList.tsx
albertandrejev: Whole this method can be rewritten to separate component.

packages/sil-voting-app/src/routes/dashboard/components/ProposalsList.tsx
abefernan: The first `map` is to convert from `SilProposal` to `ProposalProps`, the second `map` is to have an `index` that lets me do conditional rendering and add `Hairline`s between the rendered proposals.

packages/sil-voting-app/src/routes/dashboard/components/ProposalsList.tsx
abefernan: I can't do `storedProposals.map()` because it's not an array nor is it iterable, so I've done `proposals.map()` where `proposals` would be `const proposals = Object.values(storedProposals).map(state => state.proposal);`

packages/sil-voting-app/src/routes/dashboard/components/Proposal/RenderedProposal.tsx
```diff
@@ -0,0 +1,24 @@
...
+import React from 'react';
+
+import Proposal, { ProposalProps } from '.';
+
+interface Props {
+  readonly proposal: ProposalProps;
+  readonly index: number;
+}
+
+const RenderedProposal = (props: Props): JSX.Element => {
```
albertandrejev: We are usually using object destructuring for props properties.

packages/sil-voting-app/src/routes/dashboard/components/Proposal/RenderedProposal.tsx
```diff
@@ -0,0 +1,24 @@
...
+  if (props.index === 0) {
+    return <React.Fragment key={props.index}>{Proposal(props.proposal)}</React.Fragment>;
```
albertandrejev: You can set `key` for `<RenderedProposal key={index}/>` instead of passing it to the component itself. Also using array `index` as `key` propery is not good idea. https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

packages/sil-voting-app/src/routes/dashboard/components/Proposal/RenderedProposal.tsx
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  return (
+    <React.Fragment key={props.index}>
```
albertandrejev: Same here.

packages/sil-voting-app/src/routes/dashboard/components/Proposal/RenderedProposal.tsx
```diff
@@ -0,0 +1,24 @@
...
+      <Hairline />
+      {Proposal(props.proposal)}
```
albertandrejev: Why don't you want to use it as component? You can pass `props.proposal` as `<Proposal {...props.proposal} />`.

packages/sil-voting-app/src/routes/dashboard/components/Proposal/RenderedProposal.tsx
abefernan: Thanks for the advice, I changed every component to follow that approach.

packages/sil-voting-app/src/routes/dashboard/components/Proposal/RenderedProposal.tsx
abefernan: I was aware, totally forgot to change it when I had `proposal.id` in scope, thank you.

packages/sil-voting-app/src/routes/dashboard/components/Proposal/RenderedProposal.tsx
abefernan: Must be a remnant of some old code, I've just changed it.

packages/sil-voting-app/src/store/reducers/proposals/actions.ts
```diff
@@ -0,0 +1,64 @@
+import { AddProposalsActionType, SilProposal } from './reducer';
```
apanizo: Just one comment, I recommend to use:
```
src/store/index <-- where we connect reducers and prepare redux
src/store/proposals <-- for each type of "thing" have its own folder.
src/store/extension
```

PR #390, opened 2019-07-18
Feature – Use natural identities response type

webmaster128:
~Based on #389. Only check last commit.~

--------------

The type `GetIdentitiesResponse` is now a list of identities to better represent the [return type of the RPC method getIdentities](https://github.com/iov-one/iov-core/blob/master/docs/out-of-process-rpc.md#methods).

The logic of the grouping is unchanged. It was just made explicit in the `groupIdentitiesByChain` method. This prepares the expansion to multiple identities per chainId but does not change the current redux logic.

### DIFF
--- a/packages/bierzo-wallet/src/communication/identities/index.ts
+++ b/packages/bierzo-wallet/src/communication/identities/index.ts
@@ -24,7 +24,7 @@ function isArrayOfIdentity(data: any): data is ReadonlyArray<Identity> {
 
 /**
  * The response of a "getIdentities" RPC call to the browser extension.
- * `undefined` represents the case that the website cound not connect to the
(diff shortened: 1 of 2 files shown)


packages/bierzo-wallet/src/communication/identities/index.ts
```diff
@@ -22,14 +22,19 @@ function isArrayOfIdentity(data: any): data is ReadonlyArray<Identity> {
   return data.every(isIdentity);
 }
 
-type GetIdentitiesResponse = { [chain: string]: Identity } | undefined;
+/**
+ * The response of a "getIdentities" RPC call to the browser extension.
+ * `undefined` represents the case that the website cound not connect to the
```
apanizo: could

PR #391, opened 2019-07-18
Feature - Use ws connection

apanizo:
**Description**
In this PR we introduce the usage of WS for allowing in a different step drink from faucet. This PR has simplified the drink faucet process, only calling `credit` on specified tokens.

### DIFF
--- a/packages/bierzo-wallet/src/config/index.ts
+++ b/packages/bierzo-wallet/src/config/index.ts
@@ -26,7 +26,7 @@ export interface ChainSpec {
 
 export interface FaucetSpec {
   readonly uri: string;
-  readonly token: string;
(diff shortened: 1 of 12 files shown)


packages/bierzo-wallet/src/config/index.ts
```diff
@@ -26,7 +26,7 @@ export interface ChainSpec {
 
 export interface FaucetSpec {
   readonly uri: string;
-  readonly token: string;
+  readonly token: ReadonlyArray<string>;
```
webmaster128: please rename to plural `tokens` is you want to support a list here. However, right now one token per faucet should be sufficient

PR #402, opened 2019-07-18
Feature – Remove obsolete transaction creator address

webmaster128:
What was the transaction creator address before is now `SendTransaction.sender`.


apanizo:
Can you give a bit of more context, as far as I remember we thought it would be good to know who is creating the request,  right?

webmaster128:
> as far as I remember we thought it would be good to know who is creating the request, right?

Right now there are two things mixed up:
1. The sender of a SendTransaction (i.e. where the money comes from)
2. The address of the creator of the transaction

There are cases where those are not the same, for example when tokens are sent that belong to a multi-sig contract. Then 1. is the address of the contract and the transaction creator in 2. is one of the people that own the multisig contract and sign transactions.

You are definitely right that is makes sense the show 2. as well. However, this needs a proper design and should work the same way for all transaction types. The UI only allowed to display 1. so far.

This change simplifies things for now and leaves the task of showing the transaction creator open.

apanizo:
Ah! Ok, so we remove the feature because it needs more work for showing properly information. Sounds good.

PR #403, opened 2019-07-19
Feature #252 - Show human readable address on balance route.

### LINKED ISSUE #252: Balance Page : Show human readable address, if available
Detect if the wallet user has a human readable name. If not display a reminder notification “Get your human readable address” with a link to the create username page.

The logic for this can be found in the bootSequence in wallet-demo

albertandrejev:
**Description**
This PR will add support of BNS name to the balance route component. It will close #252.

**Without name**
balance-without-name

**With name**
balance-need-funds

### DIFF
--- a/packages/bierzo-wallet/src/routes/balance/components/index.tsx
+++ b/packages/bierzo-wallet/src/routes/balance/components/index.tsx
@@ -13,7 +13,7 @@ import receive from '../assets/transactionReceive.svg';
 import send from '../assets/transactionSend.svg';
 
 interface Props {
-  readonly name: string | undefined;
+  readonly iovAddress: string | undefined;
   readonly tokens: { [token: string]: Amount };
   readonly onSendPayment: () => void;
   readonly onReceivePayment: () => void;
@@ -55,7 +55,7 @@ const Card = ({ id, text, logo, onAction }: CardProps): JSX.Element => {
   );
 };
 
-const BalanceLayout = ({ name, tokens, onSendPayment, onReceivePayment }: Props): JSX.Element => {
+const BalanceLayout = ({ iovAddress, tokens, onSendPayment, onReceivePayment }: Props): JSX.Element => {
   const hasTokens = tokens && Object.keys(tokens).length;
   const theme = useTheme<Theme>();
 
@@ -72,7 +72,7 @@ const BalanceLayout = ({ name, tokens, onSendPayment, onReceivePayment }: Props)
       <Block bgcolor={theme.palette.background.paper} height="unset" width={450}>
         <Block padding={4} display="flex" flexDirection="column">
           <Typography variant="h5" align="center" weight="light">
-            {name ? name : '--'}
+            {iovAddress ? iovAddress : 'Get your human readable address.'}
           </Typography>
           <Hairline space={4} />
           <Typography variant="subtitle2" align="center">
--- a/packages/bierzo-wallet/src/routes/balance/components/index.tsx
+++ b/packages/bierzo-wallet/src/routes/balance/components/index.tsx
@@ -72,7 +72,7 @@ const BalanceLayout = ({ iovAddress, tokens, onSendPayment, onReceivePayment }: 
       <Block bgcolor={theme.palette.background.paper} height="unset" width={450}>
         <Block padding={4} display="flex" flexDirection="column">
           <Typography variant="h5" align="center" weight="light">
-            {iovAddress ? iovAddress : 'Get your human readable address.'}
+            {iovAddress ? iovAddress : 'No human readable address registered.'}
           </Typography>
           <Hairline space={4} />
           <Typography variant="subtitle2" align="center">
--- a/packages/bierzo-wallet/src/routes/balance/index.dom.spec.ts
+++ b/packages/bierzo-wallet/src/routes/balance/index.dom.spec.ts
@@ -9,7 +9,7 @@ import { UsernamesState } from '../../store/usernames';
 import { expectRoute } from '../../utils/test/dom';
 import { findRenderedDOMComponentWithId } from '../../utils/test/reactElemFinder';
 import { PAYMENT_ROUTE, RECEIVE_FROM_IOV_USER } from '../paths';
-import { getIOVUsername, getNoFundsMessage } from './test/operateBalances';
+import { getIovUsername, getNoFundsMessage } from './test/operateBalances';
 import { travelToBalance } from './test/travelToBalance';
 
 const balancesAmount: DeepPartial<BalanceState> = {
@@ -84,7 +84,7 @@ describe('The /balance route', () => {
     });
 
     it('should show bns username', async () => {
-      const noUsernameMessage = getIOVUsername(TestUtils.scryRenderedDOMComponentsWithTag(balanceDom, 'h5'));
+      const noUsernameMessage = getIovUsername(TestUtils.scryRenderedDOMComponentsWithTag(balanceDom, 'h5'));
 
       expect(noUsernameMessage).toBe('albert*iov');
     });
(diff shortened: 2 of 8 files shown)


packages/bierzo-wallet/src/routes/balance/components/index.tsx
```diff
@@ -72,7 +72,7 @@ const BalanceLayout = ({ name, tokens, onSendPayment, onReceivePayment }: Props)
       <Block bgcolor={theme.palette.background.paper} height="unset" width={450}>
         <Block padding={4} display="flex" flexDirection="column">
           <Typography variant="h5" align="center" weight="light">
-            {name ? name : '--'}
+            {iovAddress ? iovAddress : 'Get your human readable address.'}
```
apanizo: Can you check with @kmw101 and @webmaster128 if they give thumbs up for the 'Get your human-readable address' text?

packages/bierzo-wallet/src/routes/balance/index.dom.spec.ts
```diff
@@ -24,6 +25,15 @@ const balancesAmount: DeepPartial<BalanceState> = {
   },
 };
 
+const bnsChaingId = 'local-bns-devnet';
```
apanizo: typo, should be -> bnsChainId

packages/bierzo-wallet/src/routes/balance/index.e2e.spec.ts
```diff
@@ -55,4 +56,10 @@ withChainsDescribe('E2E > Balance route', (): void => {
     expect(secondBalance).toBe('10 CASH');
     expect(thirdBalance).toBe('10 ETH');
   }, 45000);
+
+  fit('should contain message to get username', async (): Promise<void> => {
```
apanizo: fit?

packages/bierzo-wallet/src/store/usernames/selectors.ts
```diff
@@ -0,0 +1,5 @@
+import { RootState } from '../reducers';
+import { BwUsername } from '.';
+
+export const getUsername = (state: RootState): BwUsername | undefined =>
+  Object.values(state.usernames).find(username => username);
```
apanizo: What is the reasoning behind this? Get the first username? Why? Are we assuming they are going to be the same for all addresses?

Can you add a comment with some explanations?

packages/bierzo-wallet/src/routes/balance/index.e2e.spec.ts
apanizo: Also, we do not have an e2e integration test for showing a real human readable address, right?

albertandrejev: Yes we don't, but how to show it in e2e test? I wanted to ask this too.

packages/bierzo-wallet/src/store/usernames/selectors.ts
albertandrejev: 1. This is not possible to show different usernames according to current wallet spec.
2. I saw this code snippet in `wallet-demo`: 
```
const bnsConn = value.connection as BnsConnection; dispatch(setBnsChainId(bnsConn.chainId()));
```
which shows me that `bnsConn.chainId()` contains bns chain id.
And I decided as we are checking for Bns in [/usernames/actions.ts#L17](https://github.com/iov-one/ponferrada/blob/85e6ed75847d4282684e5d7f4b63706166c3f755/packages/bierzo-wallet/src/store/usernames/actions.ts#L17)  and according to the whole of this method logic then I can use first record from `usernames` state as username of the current logged in user.

Correct me if I am wrong. @webmaster128.

packages/bierzo-wallet/src/routes/balance/components/index.tsx
apanizo: 👌

packages/bierzo-wallet/src/routes/balance/index.e2e.spec.ts
apanizo: In this case, let @webmaster128 know we need to have a ticket for having an e2e test where:

1. We signup in the app using the extension
2. We buy a human-readable address
3. We travel to the account balance view
4. We see the human-readable address just bought. 

Also, do the same and assure if we just reload, the page, we do login and the human-readable address is there by default.

packages/bierzo-wallet/src/store/usernames/selectors.ts
webmaster128: Logic is okay for now. Multiple usernames are an edge case we can take care of later on.

But please update code a little make explicit this is a "first" getter. By the way, the simpler way to implement "first§ in a type-safe way is `.find(() => true)`.

```ts
export const getUsername = (state: RootState): BwUsername | undefined => {
  const firstUsername = Object.values(state.usernames).find(() => true);
  return firstUsername;
}
  
```

packages/bierzo-wallet/src/routes/balance/index.e2e.spec.ts
webmaster128: @albertandrejev @apanizo if you want to extract work into a separate ticket, please create that on your own. There is no need for me to think about every single tickets that is created, especially when you are the experts of the particular job.

packages/bierzo-wallet/src/store/usernames/selectors.ts
apanizo: > This is not possible to show different usernames according to current wallet spec.

I am not saying you need to support multiple usernames (one by chain), but if we just take the first one, due a product restriction, we should comment it. For avoid confusion.

If you take a look into the usernames reducer you will see an array of usernames, one per key. In the wallet demo we had the same username for all chains, but unless I am wrong we could have one per chain.

My point is, seeing the username is tied to a chain id, having a `getUsername` selector without having specified somehow the chainId is kind of incorrect. I suggest:
```
const getUsernameByChainId = createSelector(
  state => Object.values(state.usernames),
  (_, chainId) => chainId,
  (usernames, chainId) => usernmaes.find(username => username.chainId === chainId)
)

....
// in react component
const username = useSelector(state =>
    getUsernameByChainId(state, 'bnsChainId')
  ))
```
👀 👀 👀 
For now, I suggest to hardcode the bnsChainId, but contact @webmaster128 and ask **him to create a ticket** for introducing in the extension reducer an action for spec the bns chain id as separate property:
```
export interface ExtensionState {
  readonly connected: boolean;
  readonly installed: boolean;
  readonly keys: { [chain: string]: string };
  readonly bnsChainId: string;
}
```
Or if you think this is messing up things, just create `src/store/chains` reducer and place there the chainIds of supported chains.

packages/bierzo-wallet/src/routes/balance/index.e2e.spec.ts
apanizo: oh! Thanks for the clarification, thought everything had to go through you. Maybe we should involve Martin for validating them?

webmaster128: I assume everything is valid by default. If I find something that makes no sense, we can discuss and invalidate/adjust/close on demand. You guys know what you are doing. I just take care of prioritization.

packages/bierzo-wallet/src/store/usernames/selectors.ts
apanizo: @webmaster128 can you validate the reasoning here and then, @albertandrejev create a ticket pitching Martin for validating it?

packages/bierzo-wallet/src/routes/balance/index.e2e.spec.ts
albertandrejev: https://github.com/iov-one/ponferrada/issues/405

packages/bierzo-wallet/src/store/usernames/selectors.ts
webmaster128: > My point is, seeing the username is tied to a chain id, having a getUsername selector without having specified somehow the chainId is kind of incorrect.

Usernames only exist ion the BNS chain. What we want to show is all usernames that the user owns (typically 0 or 1). I will go though the code that creates this data today and try to clarify the situation. Albert's use makes sense to me.

albertandrejev: Yes, this was my first idea is to hardcode chain id.

albertandrejev: @apanizo #408.

PR #411, opened 2019-07-22
Feature #253 - Make signAndPost request from Payment route

### LINKED ISSUE #253: Send Payment : Route for continue
On clicking the <continue> button the browser extension is called to sign the transaction.

- [x] Send payment (#411)
- [x] Storybook (#411)
- [ ] DOM and E2E tests (#417)

albertandrejev:
**Description**
This PR will implement payment logic for Payment route. It will implement "Send payment" and "Storybook" parts of this issue #253.

### DIFF
--- a/packages/bierzo-wallet/src/communication/signAndPost/index.ts
+++ b/packages/bierzo-wallet/src/communication/signAndPost/index.ts
@@ -1,9 +1,10 @@
 /*global chrome*/
 import {
   Address,
+  Amount,
+  ChainId,
   Identity,
   SendTransaction,
-  TokenTicker,
   TransactionId,
   UnsignedTransaction,
   WithCreator,
@@ -13,27 +14,33 @@ import { isJsonRpcErrorResponse, JsonRpcRequest, makeJsonRpcId, parseJsonRpcResp
 import { TransactionEncoder } from '@iov/multichain';
 
 import { extensionId } from '..';
-import { ChainConfig } from '../../config';
-import { getConnectionFor } from '../../logic/connection';
+import { getCodecForChainId } from '../../logic/codec';
+import { getConnectionForChainId } from '../../logic/connection';
 
-async function withChainFee<T extends UnsignedTransaction>(chain: ChainConfig, transaction: T): Promise<T> {
-  const connection = await getConnectionFor(chain.chainSpec);
+async function withChainFee<T extends UnsignedTransaction>(chainId: ChainId, transaction: T): Promise<T> {
+  const connection = await getConnectionForChainId(chainId);
   const withFee = await connection.withDefaultFee(transaction);
   return withFee;
 }
 
-/*const generateSignAndPostRequest = async (chain: ChainConfig, creator: Identity): Promise<JsonRpcRequest> => {
-  const transactionWithFee: SendTransaction & WithCreator = await withChainFee(chain, {
+const generateSignAndPostRequest = async (
+  chainId: ChainId,
+  creator: Identity,
+  recipient: Address,
+  amount: Amount,
+  memo: string | undefined,
+): Promise<JsonRpcRequest> => {
+  const codec = await getCodecForChainId(chainId);
+
+  const transactionWithFee: SendTransaction & WithCreator = await withChainFee(chainId, {
     kind: 'bcp/send',
-    recipient: '0x0000000000000000000000000000000000000000' as Address,
-    creator: creator,
-    sender: ethereumCodec.identityToAddress(creator),
-    amount: {
-      quantity: '1234000000000000000',
-      fractionalDigits: 18,
-      tokenTicker: 'ETH' as TokenTicker,
-    },
+    recipient,
+    creator,
+    sender: codec.identityToAddress(creator),
+    amount: amount,
+    memo: memo ? memo : '',
   });
+  const tx = TransactionEncoder.toJson(transactionWithFee);
 
   return {
     jsonrpc: '2.0',
@@ -41,13 +48,20 @@ async function withChainFee<T extends UnsignedTransaction>(chain: ChainConfig, t
     method: 'signAndPost',
     params: {
       reason: TransactionEncoder.toJson('I would like you to sign this request'),
-      transaction: TransactionEncoder.toJson(transactionWithFee),
+      transaction: tx,
     },
   };
 };
 
-export const sendSignAndPostRequest = async (creator: Identity): Promise<TransactionId | null> => {
-  const request = await generateSignAndPostRequest(creator);
+export const sendSignAndPostRequest = async (
+  chainId: ChainId,
+  creator: Identity,
+  recipient: Address,
+  amount: Amount,
+  memo: string | undefined,
+): Promise<TransactionId | null> => {
+  const request = await generateSignAndPostRequest(chainId, creator, recipient, amount, memo);
+  console.log(request);
 
   return new Promise((resolve, reject) => {
     chrome.runtime.sendMessage(extensionId, request, response => {
@@ -72,4 +86,3 @@ export const sendSignAndPostRequest = async (creator: Identity): Promise<Transac
     });
   });
 };
-*/
--- a/packages/bierzo-wallet/src/communication/signAndPost/index.ts
+++ b/packages/bierzo-wallet/src/communication/signAndPost/index.ts
@@ -9,7 +9,6 @@ import {
   UnsignedTransaction,
   WithCreator,
 } from '@iov/bcp';
-import { ethereumCodec, EthereumConnection } from '@iov/ethereum';
 import { isJsonRpcErrorResponse, JsonRpcRequest, makeJsonRpcId, parseJsonRpcResponse2 } from '@iov/jsonrpc';
 import { TransactionEncoder } from '@iov/multichain';
 
--- a/packages/bierzo-wallet/src/communication/signAndPost/index.ts
+++ b/packages/bierzo-wallet/src/communication/signAndPost/index.ts
@@ -13,16 +13,17 @@ import { isJsonRpcErrorResponse, JsonRpcRequest, makeJsonRpcId, parseJsonRpcResp
 import { TransactionEncoder } from '@iov/multichain';
 
 import { extensionId } from '..';
+import { ChainConfig } from '../../config';
+import { getConnectionFor } from '../../logic/connection';
 
-async function withEthereumFee<T extends UnsignedTransaction>(transaction: T): Promise<T> {
-  const connection = await EthereumConnection.establish('http://localhost:8545', {});
+async function withChainFee<T extends UnsignedTransaction>(chain: ChainConfig, transaction: T): Promise<T> {
+  const connection = await getConnectionFor(chain.chainSpec);
   const withFee = await connection.withDefaultFee(transaction);
-  connection.disconnect();
   return withFee;
 }
 
-const generateSignAndPostRequest = async (creator: Identity): Promise<JsonRpcRequest> => {
-  const transactionWithFee: SendTransaction & WithCreator = await withEthereumFee({
+/*const generateSignAndPostRequest = async (chain: ChainConfig, creator: Identity): Promise<JsonRpcRequest> => {
+  const transactionWithFee: SendTransaction & WithCreator = await withChainFee(chain, {
     kind: 'bcp/send',
     recipient: '0x0000000000000000000000000000000000000000' as Address,
     creator: creator,
@@ -71,3 +72,4 @@ export const sendSignAndPostRequest = async (creator: Identity): Promise<Transac
     });
   });
 };
+*/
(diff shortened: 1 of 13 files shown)


packages/bierzo-wallet/src/routes/balance/components/index.tsx
```diff
@@ -13,7 +13,7 @@ import receive from '../assets/transactionReceive.svg';
 import send from '../assets/transactionSend.svg';
 
 interface Props {
-  readonly iovAddress: string | undefined;
+  readonly iovAddress?: string;
```
webmaster128: Why this change? `string | undefined` is more strict than an optional `string` since it requires setting the argument. Before

```tsx
        <Layout
          tokens={BALANCE}
          onReceivePayment={action('onReceivePayment')}
          onSendPayment={action('onSendPayment')}
        />
```

was an error (someone forgot to set `iovAddress`) but

```tsx
        <Layout
          iovAddress={undefined}
          tokens={BALANCE}
          onReceivePayment={action('onReceivePayment')}
          onSendPayment={action('onSendPayment')}
        />
```

was okay. With this change, the `iovAddress` property can be omitted. Since we always want to specify `iovAddress`, even if there is no address, the old code was better for that case.

webmaster128: Nevermind, now I see. It makes the story code easier.

albertandrejev:
Due to a big size of this PR I will implement tests in next PR.

packages/bierzo-wallet/src/communication/signAndPost/index.ts
```diff
@@ -1,52 +1,66 @@
...
-    amount: {
-      quantity: '1234000000000000000',
-      fractionalDigits: 18,
-      tokenTicker: 'ETH' as TokenTicker,
-    },
+    recipient,
+    creator,
+    sender: codec.identityToAddress(creator),
+    amount: amount,
+    memo: memo ? memo : '',
```
webmaster128: Why this case distinction? You can pass `string | undefined` into `memo` directly

packages/bierzo-wallet/src/logic/account.ts
```diff
@@ -0,0 +1,30 @@
+import { Address, ChainId } from '@iov/bcp';
+import { BnsConnection } from '@iov/bns';
+
+import { getConfig } from '../config';
+import { getConnectionFor, isBnsSpec } from './connection';
+
+// getAddressByName returns the address associated with the name, or undefined if not registered
+// the name should not have the "*iov" suffix
```
webmaster128: Can you use doc comments? (starting with `/**` ending with `/*`. Then you get this documentation in the IDE when you mouseover symbols

packages/bierzo-wallet/src/logic/account.ts
```diff
@@ -0,0 +1,30 @@
...
+  for (const chain of chains) {
+    if (!isBnsSpec(chain.chainSpec)) {
+      continue;
+    }
+
+    const connection = (await getConnectionFor(chain.chainSpec)) as BnsConnection;
+    const usernames = await connection.getUsernames({ username });
+    if (usernames.length !== 1) return undefined;
+
+    const bnsUsername = usernames[0].addresses.find(addr => addr.chainId === chainId);
```
webmaster128: `bnsUsername` -> `chainAddressPair`

packages/bierzo-wallet/src/logic/account.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    return bnsUsername ? bnsUsername.address : undefined;
+  }
+}
```
webmaster128: Nice implementation 👍 You are an IOV-Core pro soon

packages/bierzo-wallet/src/routes/payment/index.tsx
```diff
@@ -1,87 +1,85 @@
...
+    let recipient: string | undefined = formValues[ADDRESS_FIELD];
+    if (isIov(recipient)) {
+      recipient = await getAddressByName(recipient.replace(/\*iov$/, ''), chainId);
+    }
 
-const onSubmit = (): void => {};
+    if (!recipient) {
+      toast.show('IOV username was not found', ToastVariant.ERROR);
+      return;
+    }
```
webmaster128: This code is very confusing because `recipient` holds the user input as well as the native blockchain address. Could you split that in two variables where the user input can be of type `string | undefined` and the native address of type `Address | undefined`?

packages/bierzo-wallet/src/routes/payment/index.tsx
```diff
@@ -1,87 +1,85 @@
...
 
-const Payment = ({ location }: RouteComponentProps): JSX.Element => {
-  const classes = useStyles();
+    const plainPubkey = pubKeys[chainId];
+    if (!plainPubkey) {
+      toast.show('Public key was not found', ToastVariant.ERROR);
```
webmaster128: "None of your identities can send on this chain"

packages/bierzo-wallet/src/routes/payment/index.tsx
```diff
@@ -1,87 +1,85 @@
...
-  const { form, handleSubmit, invalid, pristine, submitting } = useForm({
-    onSubmit,
-  });
+    const identity: Identity = TransactionEncoder.fromJson(JSON.parse(plainPubkey));
+
+    try {
+      const transactionId = await sendSignAndPostRequest(
+        chainId,
+        identity,
+        formValues[ADDRESS_FIELD] as Address,
```
webmaster128: Don't you want the `recipient` address you resolved above?

packages/bierzo-wallet/src/logic/account.ts
```diff
@@ -0,0 +1,30 @@
+import { Address, ChainId } from '@iov/bcp';
+import { BnsConnection } from '@iov/bns';
+
+import { getConfig } from '../config';
+import { getConnectionFor, isBnsSpec } from './connection';
+
+// getAddressByName returns the address associated with the name, or undefined if not registered
+// the name should not have the "*iov" suffix
+export async function getAddressByName(username: string, chainId: ChainId): Promise<Address | undefined> {
```
apanizo: The existing usernames info are already in the redux store, I think this code should be a selector because we are duplicating effort. Also having centralized the info in one place, making redux the single point of truth for views make the app stronger.

Something like:
```
const getAddressByNameSelector = createSelector(
  state => state.usernames,
  (_, chainId) => chainId,
  (usernames, chainId) => {
       if (!usernames) {
          throw new Error(...);
       }

       const username = usernames.addresses.find(name => name.chainId === chainId)

       if (!username) {
          return undefined;
       }

       return username.address;
}):

....

// In a React component
const address = useSelector(state => getAddressByNameSelector(state, chainId));
```

webmaster128: > The existing usernames info are already in the redux store

Those are different usernames. In the redux store we have the names owned by the current user (i.e. the *sender*). But `getAddressByName` makes a lookup in the BNS for the *recipient*.

apanizo: Ahh! Good point. I understand, and in my PR probably I will need to use this method as well.

In this case, the name of the function mislead me, @albertandrejev can you just ignore my comment and rename the function to `lookupRecipientAddressByName`.

webmaster128: Good idea 👆

packages/bierzo-wallet/src/communication/signAndPost/index.ts
albertandrejev: Unfortunately not :( I got 'Cannot encode type to JSON' if I tried to use undefined `memo`. It happens during [this call](https://github.com/iov-one/ponferrada/blob/a961be138e79875158391ed779da55988c9c82f3/packages/bierzo-wallet/src/communication/signAndPost/index.ts#L42).

webmaster128: Ah okay, this is a problem that needs fixing. Thanks for letting know. I will find a solution on the IOV-Core side. For now, your workaround is fine.

packages/bierzo-wallet/src/routes/payment/index.tsx
```diff
@@ -1,87 +1,90 @@
...
+        formValues[ADDRESS_FIELD].replace(/\*iov$/, ''),
+        chainId,
+      );
+    } else {
+      recipient = formValues[ADDRESS_FIELD] as Address;
+    }
 
-const onSubmit = (): void => {};
+    if (!recipient) {
+      toast.show('IOV username was not found', ToastVariant.ERROR);
```
webmaster128: This error message belong into the `if (isIov(formValues[ADDRESS_FIELD])) {` block. When the user entered a native address, the text makes no sense

albertandrejev: In case of native address this text will not appear, because `recipient` will not be `undefined`

webmaster128: Ah, I see. You can express this fact with code and type system more clearly:

```ts
    let recipient: Address;
    if (isIov(formValues[ADDRESS_FIELD])) {
       const addressForName = await lookupRecipientAddressByName(
        formValues[ADDRESS_FIELD].replace(/\*iov$/, ''),
        chainId,
      );
      if (!addressForName) {
        toast.show('IOV username was not found', ToastVariant.ERROR);
        return;
      }
      recipient = addressForName;
    } else {
      recipient = formValues[ADDRESS_FIELD] as Address;
    }
```

albertandrejev: To be honest I do not like nested `if`'s :)

webmaster128: That's a matter if taste, of course. But nested `if`s allow you to write code that matches the problem that you are facing here. I don't think nested `if`s are a problem as long as you don't nest 5 levels deep

packages/bierzo-wallet/src/routes/payment/index.tsx
```diff
@@ -1,87 +1,90 @@
-import { Theme } from '@material-ui/core';
-import { makeStyles } from '@material-ui/styles';
-import Block from 'medulas-react-components/lib/components/Block';
-import Button from 'medulas-react-components/lib/components/Button';
-import Form, { useForm } from 'medulas-react-components/lib/components/forms/Form';
+import { Address, Identity, TokenTicker } from '@iov/bcp';
+import { TransactionEncoder } from '@iov/core';
```
webmaster128: `from '@iov/multichain'`

packages/bierzo-wallet/src/logic/account.ts
```diff
@@ -0,0 +1,34 @@
...
+  for (const chain of chains) {
+    if (!isBnsSpec(chain.chainSpec)) {
+      continue;
+    }
+
+    const connection = (await getConnectionFor(chain.chainSpec)) as BnsConnection;
+    const usernames = await connection.getUsernames({ username });
+    if (usernames.length !== 1) return undefined;
+
+    const chainAddressPair = usernames[0].addresses.find(addr => addr.chainId === chainId);
```
webmaster128: `usernames[0].addresses` -> `usernames[0].targets`

packages/bierzo-wallet/src/routes/payment/index.tsx
```diff
@@ -1,87 +1,90 @@
...
-import { RouteComponentProps, withRouter } from 'react-router';
+import * as ReactRedux from 'react-redux';
 
-import Header from '../../components/Header';
-import CurrencyToSend from './components/CurrencyToSend';
-import ReceiverAddress from './components/ReceiverAddress';
-import TextNote from './components/TextNote';
+import { history } from '..';
+import { sendSignAndPostRequest } from '../../communication/signAndPost';
+//import { sendSignAndPostRequest } from '../../communication/signAndPost';
```
webmaster128: Remove outdated comment

packages/bierzo-wallet/src/logic/account.ts
```diff
@@ -0,0 +1,34 @@
...
+      continue;
+    }
+
+    const connection = (await getConnectionFor(chain.chainSpec)) as BnsConnection;
+    const usernames = await connection.getUsernames({ username });
+    if (usernames.length !== 1) return undefined;
+
+    const chainAddressPair = usernames[0].targets.find(addr => addr.chainId === chainId);
+
+    return chainAddressPair ? chainAddressPair.targets : undefined;
```
webmaster128: the property in the `chainAddressPair` did not change (see compile error in CI). `chainAddressPair` is of type

```ts
export interface ChainAddressPair {
  readonly chainId: ChainId;
  readonly address: Address;
}
```

The change in line 26 is correct

webmaster128:
@apanizo I'm merging this since all major remarks are implemented. Let's assume the recact stuff is good (UI works as expected from the user's perspective) and if we find something fishy, we fix later. We need to ensure the diffs of our work are small. 🚀

PR #413, opened 2019-07-23
Feature – Upgrade IOV-Core to 0.16

### LINKED ISSUE #393: Fix transaction signing
As described in https://medium.com/@simonwarta/limitations-of-the-instanceof-operator-f4bcdbe7a400, the `instanceof` operator used by IOV-Core'r `isUnsignedTransaction` does not work reliably after pulling objects from background script into UI. To make this work again, transaction need to be in a format that can trivially be copied.

### LINKED ISSUE #394: Upgrade IOV-Core to 0.16

webmaster128:
Closes #393
Closes #394

Important changes

- the package @iov/core was renamed to @iov/multichain
- Usernames that are sent to and read from IOV-Core now contain the `*iov` suffix automatically


packages/bierzo-wallet/src/store/balances/index.unit.spec.ts
```diff
@@ -45,11 +45,9 @@ withChainsDescribe('Tokens reducer', () => {
...
     const balances = store.getState().balances;
-    const ethBalance = {
-      fractionalDigits: 18,
-      quantity: '100000000000000000000',
-      tokenTicker: 'ETH',
-    };
-    expect(balances).toEqual({ ETH: ethBalance });
+    expect(balances.ETH).toBeDefined();
+    expect(balances.ETH.fractionalDigits).toEqual(18);
+    expect(balances.ETH.quantity).toMatch(/^[0-9]{10,22}$/);
```
albertandrejev: Quantity can by any number from 10 to 22 symbols length? I think it is much better to test it against string at least for readability.

webmaster128: The problem was that some other tests send tokens from this account, such that the quantity changes from time to time. Instead of one fixed value we just require _some large number_ now

PR #414, opened 2019-07-23
Feature #397 Transactions view

### LINKED ISSUE #397: Implement /transactions page
`/transactions` should list all transactions of the current user's identities

apanizo:
**Description**
This PR implements the Transaction view, taking advantage of the redux connection.

It closes #397 

Screenshot 2019-07-25 09 36 32


packages/bierzo-wallet/src/routes/transactions/components/sorting.ts
```diff
@@ -0,0 +1,45 @@
...
+export type TxsOrder = 'Date';
+
+export const filterTxsBy = (
+  txs: ReadonlyArray<ProcessedTx>,
+  rowsPerPage: number,
+  pageNumber: number,
+  orderBy: TxsOrder,
+  order: SortOrder,
+): ReadonlyArray<ProcessedTx> => {
+  const orderedTxs = txs.slice(0).sort((a: ProcessedTx | undefined, b: ProcessedTx | undefined) => {
```
webmaster128: - When `txs` is of type `ReadonlyArray<ProcessedTx>`, why are the elements `a` and `b` in the compare function of type `ProcessedTx | undefined`?
- `slice(0)` is there to make a mutable copy, right?

packages/bierzo-wallet/src/routes/transactions/components/sorting.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (!a || !b) {
+      return 0;
+    }
+
+    if (orderBy === TX_DATE_COLUMN) {
+      return (a.time < b.time ? -1 : a.time > b.time ? 1 : 0) * order;
+    }
+
+    return 0;
```
webmaster128: By default elements are equal? This does not make sense. Better avoid calling `.sort()` when you don't want to sort

packages/bierzo-wallet/src/routes/transactions/components/DownloadCSV/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    width: 'auto',
+    justifyContent: 'left',
+    textTransform: 'inherit',
+  },
+  sizeSmall: {
+    height: `${theme.spacing(4)}px`,
+    padding: '0 !important',
+  },
+  text: {
+    padding: `0px ${theme.spacing(1.5)}px`,
```
albertandrejev: Nice, I thought that it should be possible to provide non integer (real) values but didn't tried.

packages/bierzo-wallet/src/routes/transactions/components/DownloadCSV/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+      >
+        <CircleImage
+          icon={download}
+          circleColor={theme.palette.primary.main}
+          alt="Download"
+          dia={diameter}
+          width={16}
+          height={16}
+        />
+        <Typography variant="subtitle2" weight="regular" className={classes.text}>
```
albertandrejev: Why don't you want to wrap it into Block with padding?

packages/bierzo-wallet/src/routes/transactions/components/DownloadCSV/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    backgroundColor: 'white',
+    color: theme.palette.text.primary,
+    padding: 0,
+    boxShadow: 'none',
+    border: `1px solid ${getBorderColor(theme)}`,
+    '&:hover': {
+      background: theme.palette.background.default,
+    },
+  },
+  panel: {
```
albertandrejev: If you will remove this style and implement it directly on `Block` component and if you will remove `text` style as well  then you can remove line `const fabClasses = {` and provide classes object directly to `Fab` component.

packages/bierzo-wallet/src/routes/transactions/components/NoTransactions/index.tsx
```diff
@@ -0,0 +1,38 @@
+import { makeStyles } from '@material-ui/core';
+import Block from 'medulas-react-components/lib/components/Block';
+import Image from 'medulas-react-components/lib/components/Image';
+import Typography from 'medulas-react-components/lib/components/Typography';
+import * as React from 'react';
+
+import noTransactions from '../../assets/noTransactions.svg';
+
+const useStyles = makeStyles({
```
albertandrejev: this style can be implemented directly on `Block` component.

packages/bierzo-wallet/src/routes/transactions/components/TxTable/TxDetails/index.tsx
```diff
@@ -0,0 +1,59 @@
+import { makeStyles } from '@material-ui/core';
+import Block from 'medulas-react-components/lib/components/Block';
+import Typography from 'medulas-react-components/lib/components/Typography';
+import * as React from 'react';
+
+import { ProcessedTx } from '../../../../../store/notifications';
+import { getAddressPrefix } from '../rowTxBuilder';
+
+const useStyles = makeStyles({
+  details: {
```
albertandrejev: Can be implemented on `Block`

packages/bierzo-wallet/src/routes/transactions/components/TxTable/TxTableHeader/utils/TxColumn.tsx
```diff
@@ -0,0 +1,36 @@
+import { makeStyles } from '@material-ui/core';
+import classNames from 'classnames';
+import Block from 'medulas-react-components/lib/components/Block';
+import Typography from 'medulas-react-components/lib/components/Typography';
+import * as React from 'react';
+
+const useStyles = makeStyles({
+  header: {
```
albertandrejev: Can be moved to `Block` but we need to figure it out what to do with `flexBasis`

packages/bierzo-wallet/src/routes/transactions/components/TxTable/TxTableHeader/utils/TxSortableColumn.tsx
```diff
@@ -0,0 +1,54 @@
...
+import Typography from 'medulas-react-components/lib/components/Typography';
+import * as React from 'react';
+
+import sortDown from '../../../../assets/sortDown.svg';
+import sortDownActive from '../../../../assets/sortDownActive.svg';
+import sortUp from '../../../../assets/sortUp.svg';
+import sortUpActive from '../../../../assets/sortUpActive.svg';
+import { calculateOppositeOrder, ORDER_ASC, ORDER_DESC, SortingStateProps } from '../../../sorting';
+
+const useStyles = makeStyles({
```
albertandrejev: same here

packages/bierzo-wallet/src/routes/transactions/components/TxTable/TxTableRow/index.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+import { getBorderColor } from '../../../../../theme/css';
+import { amountToNumber } from '../../../../../utils/balances';
+import { getDate, getTime } from '../../../../../utils/date';
+import dropdownArrow from '../../../assets/dropdownArrow.svg';
+import dropdownArrowClose from '../../../assets/dropdownArrowClose.svg';
+import { DEFAULT_ADDRESS, getAddressPrefix, getTypeIcon, TxTableRowProps } from '../rowTxBuilder';
+import TxDetails from '../TxDetails';
+
+const useStyles = makeStyles({
```
albertandrejev: same here

packages/bierzo-wallet/src/routes/transactions/components/TxTable/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+    flexDirection: 'column',
+    backgroundColor: 'white',
+    borderRadius: 4,
+    boxShadow: `0 0 20px 0 ${getShadowColor()}`,
+  },
+  column: {
+    display: 'flex',
+    flexDirection: 'column',
+  },
+}));
```
albertandrejev: Some of the styles above can be moved to `Block` but not all like `borderRadius`, `boxShadow`.

packages/bierzo-wallet/src/routes/transactions/components/sorting.ts
apanizo: The `sort` method expects a function which returns a number, I have checked and this return never will be reached  but left the default just in case to tell to the algorithm: "Ey, just leave the order of those two elements in the same position as they are"

Is there any problem with this approach which I am missing?

packages/bierzo-wallet/src/routes/transactions/components/DownloadCSV/index.tsx
apanizo: And what about classes.panel's backgroundColor? This is not accepted as Block's prop.

albertandrejev: there is `bgcolor` prop available for `Block`

packages/bierzo-wallet/src/routes/transactions/components/sorting.ts
webmaster128: Yes, there is a difference between
1. Telling the sort algorithm all elements are equal (by returning 0)
2. All elements should be returned in the same order
Most sort algorithms will return a list that is entirely mixed when all elements are equal. This is also the JS standard. V8 implements a sort algorithm that is _stable_, i.e. if two elements are equal they remain in the original order. But you cannot assume you have this property.

packages/bierzo-wallet/src/routes/transactions/components/DownloadCSV/index.tsx
albertandrejev: You can use `bgcolor` props to set background color fro `Block`

packages/bierzo-wallet/src/routes/transactions/components/DownloadCSV/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+export interface DownloadCSVProps {
+  readonly onDownloadCSV: () => void;
+}
+
+const DownloadCSV = ({ onDownloadCSV }: DownloadCSVProps): JSX.Element => {
+  const theme = useTheme<Theme>();
+  const classes = useStyles();
+
+  const fabClasses = {
```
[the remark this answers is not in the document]
albertandrejev: If you will get rid of `panel` style you can use classes directly as `fabClasses`

packages/bierzo-wallet/src/routes/transactions/components/TxTable/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+import TxTableFooter from './TxTableFooter';
+import TxTableHeader from './TxTableHeader';
+import TxTableRow from './TxTableRow';
+
+const useStyles = makeStyles((theme: Theme) => ({
+  inner: {
+    flexBasis: 'auto',
+  },
+  panel: {
+    backgroundColor: 'white',
```
albertandrejev: `bgcolor` prop for `Block`

packages/bierzo-wallet/src/routes/transactions/components/sorting.ts
apanizo: > by " all elements"

you mean the two elements which are being compared, right?

🤔 so, what should I do? throw an exception? or just leave as it is for safety? Any other idea?

webmaster128: > you mean the two elements which are being compared, right?

Yes, if the two compared elements stay in the original order, the whole list styley in the original order.

I suggest not running the sort when you want to keep the original order:

```ts
  const orderedTxs = txs.slice(0); // make a mutable copy
  
  if (orderBy === TX_DATE_COLUMN) {
    orderedTxs.sort((a: ProcessedTx, b: ProcessedTx) => {
      return (a.time < b.time ? -1 : a.time > b.time ? 1 : 0) * order;
    }
  }
```

apanizo: Done!

PR #416, opened 2019-07-24
Feature – Upgrade IOV-Core to 0.16.0-alpha.2

webmaster128:
1. Fixes the `string | undefined` bug discussed in #411
2. @iov/multichain is not needed anymore by the web apps


packages/sanes-chrome-extension/src/extension/background/model/signingServer/index.server.unit.spec.ts
```diff
@@ -1,4 +1,5 @@
-import { JsonRpcSigningServer, SignAndPostAuthorization, TransactionEncoder } from '@iov/multichain';
+import { TransactionEncoder } from '@iov/encoding';
+import { JsonRpcSigningServer, SignAndPostAuthorization } from '@iov/multichain';
```
albertandrejev: You said that
> @iov/multichain is not needed anymore by the web apps

webmaster128: This is code of the browser extension. In the extension, @iov/multichain is still used

PR #429, opened 2019-07-30
Feature – Speedup wait for faucet

webmaster128:
* Faucets run in parallel now (one job per chain)
* Deduplicate logic of `whenOnNavigatedToRoute` and `whenOnNavigatedToE2eRoute`
* Replace 18s sleep with balance list observer
* Replace magic number with `nonBalanceH6Elements` constant
* Reduce retry interval from 500ms to 400ms

### DIFF
--- a/packages/bierzo-wallet/src/utils/test/navigation.ts
+++ b/packages/bierzo-wallet/src/utils/test/navigation.ts
@@ -1,38 +1,34 @@
 import { Page } from 'puppeteer';
 
 const retryInterval = 500;
-const defaultTimeout = 18000;
+
+/**
+ * Calls callback until it returns true or timeout is reachend.
+ *
+ * @param callback a callback function that is called in a loop. It should be cheap to execute.
+ * @param timeout timeout in milliseconds. Defaults to 10 seconds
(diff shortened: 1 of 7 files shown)


webmaster128:
Fixed, should be green now

packages/bierzo-wallet/src/utils/test/navigation.ts
```diff
@@ -1,38 +1,34 @@
...
-      const actualRoute = window.location.pathname;
-      if (actualRoute === desiredRoute) {
+/**
+ * Calls callback until it returns true or timeout is reachend.
+ *
+ * @param callback a callback function that is called in a loop. It should be cheap to execute.
+ * @param timeout timeout in milliseconds. Defaults to 10 seconds
+ * @returns a promise that resolves when the callback returned true and is rejected when timeout is reached.
+ */
+export function whenTrue(callback: () => boolean | Promise<boolean>, timeout: number = 10000): Promise<void> {
```
albertandrejev: Nice!

webmaster128: :) thanks

PR #434, opened 2019-07-30
Feature – Fix transactions subscriptions

webmaster128:
Closes #412

PR #441, opened 2019-07-31
Feature – Connect to Lovenet

### LINKED ISSUE #433: Switch to Lovenet 😍

webmaster128:
Base on #434

Closes #433


packages/bierzo-wallet/src/logic/transactions.unit.spec.ts
```diff
@@ -31,19 +31,19 @@ withChainsDescribe('Logic :: transaction subscriptions', () => {
...
-    await sleep(5000);
+    // Wait for events to be processed
+    await sleep(1000);
 
-    // 2 (cash) + 2 (bash) + 2 eth
-    expect(txsSpy).toHaveBeenCalledTimes(6);
+    // Got one incoming transaction for BASH, CASH, ETH
+    expect(txsSpy).toHaveBeenCalledTimes(3);
+    const transactions = txsSpy.mock.calls.map(call => call[0]);
+    expect(new Set(transactions.map(tx => tx.amount.tokenTicker))).toEqual(new Set(['BASH', 'CASH', 'ETH']));
```
albertandrejev: Why can't we simply check against array and not against `Set`?

webmaster128: Because we don't know the oder in which the transaction events are coming in. The arrays `["BASH", "CASH", "ETH"]` and `["ETH", "BASH", "CASH"]` are different. But when converted to a set, the elements have no order anymore, such that `Set(["BASH", "CASH", "ETH"])` and `Set(["ETH", "BASH", "CASH"])` are equal.

PR #443, opened 2019-07-31
Feature - Support different txs types

### LINKED ISSUE #442: Refactor to support SendTransaction as one type
The bierzo-wallet app should support different kinds of transactions. Right now the architecture has been built around the SendTransaction type. So the first step is to decouple it allowing add more types in an easier way, without compromising the performance and keeping in control complexity.

apanizo:
**Description**
The first step for adding a general component for all txs which are not Send ones is to prepare the architecture for adding them. Those changes include:
- Change in the reducer the type, from `processedTx` to `parsedTx<T>`, where parsedTx assures to have `time` and `kind` attributes on T.
- Create a factory for giving a parsedTx get the BwTransaction component which as methods for getting its react representation and also the CSV representation.
- Update the transactions's view adding rows as an array of react elements which they are calculated upfront when parsedTxs are obtained from redux store.

It closes #442


packages/bierzo-wallet/src/logic/transactions/BwTransaction.ts
```diff
@@ -0,0 +1,14 @@
+import { Address, BlockchainConnection, ConfirmedTransaction, LightTransaction } from '@iov/bcp';
+import { ReadonlyDate } from 'readonly-date';
+
+export type ParsedTx<T> = Pick<LightTransaction, 'kind'> & { time: ReadonlyDate } & T;
+
+export abstract class BwTransaction<T extends LightTransaction, K> {
```
webmaster128: This class is not a "transaction" but a parser. So it should be called something like `BwParser`

packages/bierzo-wallet/src/logic/transactions/types/BwSendTransaction/index.tsx
```diff
@@ -0,0 +1,47 @@
+import { Address, BlockchainConnection, ConfirmedTransaction, SendTransaction } from '@iov/bcp';
+import * as React from 'react';
+
+import { BwTransaction } from '../../BwTransaction';
+import SendTransactionComponent, { BwSendTransactionProps } from './ui';
+
+export class BwSendTransaction extends BwTransaction<SendTransaction, BwSendTransactionProps> {
```
webmaster128: Also not a transaction but a parser. Suggestion: `BwSendParser`

packages/bierzo-wallet/src/logic/transactions/index.ts
```diff
@@ -2,49 +2,50 @@ import {
...
-    received,
-    recipient: payload.recipient,
-    sender: payload.sender,
-    success: true,
-  };
-};
+    return tx;
+  }
+
+  return {} as any;
```
webmaster128: Here we need a good type for unsupported transactions.

packages/bierzo-wallet/src/logic/transactions/types/BwSendTransaction/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  public csvRepresentation(tx: BwSendTransactionProps): string {
+    const parties = `"${tx.id}";"${tx.recipient}";"${tx.sender}";`;
+    const payment = `"${tx.amount.quantity}";"${tx.amount.fractionalDigits}";"${tx.amount.tokenTicker}";`;
+    const date = `"${tx.time.toISOString()}";`;
+    const status = `"${tx.received}";"${tx.success}";"${tx.err}";"${tx.memo}"`;
+
+    const txRow = `${parties}${payment}${date}${status}`;
+
+    return txRow;
```
webmaster128: Can you create an array of cells then then use `cells.join(';')` for readability?

packages/bierzo-wallet/src/routes/transactions/components/TxTable/rowTxBuilder.ts
```diff
@@ -6,16 +6,8 @@ import toAddress from '../../assets/toAddress.svg';
...
-  readonly tx: ProcessedTx;
-}
-
-export interface TxTableState {
-  readonly phoneHook: HTMLDivElement | null;
-}
-
 export interface TxTableProps extends SortingStateProps {
-  readonly txs: ReadonlyArray<ProcessedTx>;
+  readonly rows: JSX.Element[];
```
webmaster128: `readonly rows: readonly JSX.Element[];`

packages/bierzo-wallet/src/logic/transactions/index.ts
webmaster128: Converted to an exception, okay. This is also good.

packages/bierzo-wallet/package.json
```diff
@@ -37,7 +37,7 @@
   "scripts": {
     "lint": "tsc --noEmit && eslint -c .eslintrc.js --max-warnings 0 'src/**/*.ts{,x}'",
     "format": "prettier --write --loglevel warn './src/**/*.{ts,tsx,json,md,css}'",
-    "start-developmen": "REACT_APP_CONFIG=development react-scripts start",
```
webmaster128: Sorry about this one. Is fixed on master now

apanizo: This stuff happens, do not worry 😉

packages/bierzo-wallet/src/logic/transactions/types/BwSendTransaction/index.tsx
apanizo: It is already done:
```
function onDownloadCSV(): void {
    // TODO UPDATE HEADER WHEN OTHER TX TYPES ARE ADDED
    const csvHeader =
      '"ID";"Recipient";"Sender";"Quantity";"Fractional Digits";"Token Ticker";"Time";"Received";"Success";"Error";"Note"';
    const csvBody = orderedTxs.map((tx: ParsedTx<{}>) => BwTransactionFactory.getCsvRepresentation(tx));
    ....
}
```

Or... what do you mean?

webmaster128: I mean the function where I commented:

Like this:

```ts
  public csvRepresentation(tx: BwSendTransactionProps): string {
    const parties = [`"${tx.id}"`, `"${tx.recipient}"`, `"${tx.sender}"`];
    const payment = [`"${tx.amount.quantity}"`, `"${tx.amount.fractionalDigits}"`, `"${tx.amount.tokenTicker}"`];
    const date = [`"${tx.time.toISOString()}"`];
    const status = [`"${tx.received}"`, `"${tx.success}"`, `"${tx.err}"`, `"${tx.memo}"`];

    const txRow = [...parties, ...payment, ...date, ...status];

    return txRow.join(';');
  }
```

packages/bierzo-wallet/src/logic/transactions/types/BwTransactionFactory.ts
```diff
@@ -0,0 +1,48 @@
...
+      throw new Error('Not supported error txs for now');
+    }
+
+    if (!isConfirmedTransaction(trans)) {
+      throw new Error('Confirmed transaction expected');
+    }
+
+    const { transaction: payload } = trans;
+    if (isSendTransaction(payload)) {
+      return (new BwSendParser() as unknown) as BwTransaction<T, K>;
```
apanizo: @webmaster128 can you guess a solution for sorting this ugly cast out to `unknown`?

webmaster128: `getBwTransactionFrom` should not be generic as the argument is always of type `ConfirmedTransaction<LightTransaction>`. Then you can do, which avoids all the casts:

```ts
  public static getBwTransactionFrom(
    trans: ConfirmedTransaction<LightTransaction> | FailedTransaction,
  ): BwTransaction<LightTransaction, ProcessedTx> {
```

PR #444, opened 2019-07-31
Fix #439 - Float number validator

### LINKED ISSUE #439: Number placeholder format is invalid
Placeholder suggests using comma

Bildschirmfoto 2019-07-31 um 06 49 28

Comma-natation not allowed

Bildschirmfoto 2019-07-31 um 06 49 40

Since the entire app is English, I suggest requiring the English number format as well (i.e. decimal deparator must be `.`).

albertandrejev:
**Description**
This PR will fix float number validator and will allow to use float numbers with comma and point as decimal separator. It will close #439.

**Before**
number-comma-error

**After**
payment-number-placeholder

### DIFF
--- a/packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
+++ b/packages/bierzo-wallet/src/routes/payment/components/CurrencyToSend/index.tsx
@@ -89,7 +89,7 @@ const CurrencyToSend = ({ form, onTokenSelectionChanged }: Props): JSX.Element =
                 name={QUANTITY_FIELD}
                 form={form}
                 validate={validator}
-                placeholder="0,00"
+                placeholder="0.00"
                 fullWidth
                 margin="none"
               />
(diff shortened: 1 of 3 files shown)


albertandrejev:
We discussed this behaviour some time ago with @ethanfrey and @apanizo for wallet-demo and decided to use [both number formats](https://github.com/iov-one/wallet-demo/pull/276). But if you insist I can change this.

webmaster128:
If we want to support both number formats, we should do the conversion in the input field such that the user has the chance to review the input.

E.g.

1. User types-in `1,000`
2. It gets automatically converted to `1.000`
3. Field validation runs on `1.000`, i.e. validation code does not need to change

This is nice and convenient, but more work.

albertandrejev:
For example in my bank account I can use both separators without any conversion on UI.

webmaster128:
> For example in my bank account I can use both separators without any conversion on UI.

But we can do better. Accepting both input types can easily lead to sending an amount that the user did not intend to send.

albertandrejev:
So how do you think it is better to implement? Just replace placeholder? or replace comma with period during user typing (not sure that this is easy in `react-final-form-hooks`)?

webmaster128:
> So how do you think it is better to implement? Just replace placeholder? or replace comma with period during user typing (not sure that this is easy in `react-final-form-hooks`)?

I would do this in two steps:
1. Use `.` only and update placeholder (now)
2. Accept and auto-replace user inputs for convenience (post-MVP)

PR #449, opened 2019-07-31
Feature - Remove generics on parsed tx

apanizo:
In this PR I remove, as an optional step the generics on the ParsedTx. The idea behind this is to take advantage of how TS deals with inheritance in types.

If you think is a better way to do it, just approve this PR. :)

### DIFF
--- a/packages/bierzo-wallet/src/logic/transactions/types/BwParser.ts
+++ b/packages/bierzo-wallet/src/logic/transactions/types/BwParser.ts
@@ -7,14 +7,14 @@ import {
 } from '@iov/bcp';
 import { ReadonlyDate } from 'readonly-date';
 
-export type ParsedTx<T> = Pick<LightTransaction, 'kind'> & { time: ReadonlyDate } & T;
+export type ParsedTx = Pick<LightTransaction, 'kind'> & { time: ReadonlyDate };
 
 export abstract class BwParser<K> {
   abstract async parse(
     conn: BlockchainConnection,
     transaction: ConfirmedTransaction<LightTransaction> | FailedTransaction,
     currentUserAddress: Address,
-  ): Promise<ParsedTx<K>>;
-  abstract graphicalRepresentation(tx: ParsedTx<K>): JSX.Element;
-  abstract csvRepresentation(tx: ParsedTx<K>): string;
+  ): Promise<ParsedTx>;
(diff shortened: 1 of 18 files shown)


packages/bierzo-wallet/src/logic/transactions/types/BwParser.ts
```diff
@@ -7,14 +7,14 @@ import {
 } from '@iov/bcp';
 import { ReadonlyDate } from 'readonly-date';
 
-export type ParsedTx<T> = Pick<LightTransaction, 'kind'> & { time: ReadonlyDate } & T;
+export type ParsedTx = Pick<LightTransaction, 'kind'> & { time: ReadonlyDate };
```
webmaster128: With this change, `ParsedTx` is not a transaction anymore and things like `<TxItem key={(item as ProcessedTx).id} item={item as ProcessedTx}` are wrong since the required fields are missing in the type.

What about using

```ts
export type ParsedTx = Pick<LightTransaction, 'kind'> & ProcessedTx;
```

and then you work towards making ProcessedTx flexible enough to support different transaction types like here: https://github.com/iov-one/ponferrada/blob/c5e06b6/packages/sanes-chrome-extension/src/extension/background/model/persona/index.ts#L21-L41

apanizo: I think this is the model to follow, and it has been inspired based on your comment in my other PR. I thought for doing inherital checking TS needed to know the common attributes and also the specific ones of a type specific by a generic `ParsedTx<K>`.

In this way, `const sendTx ->  ParsedTx<ProcessedTx>`, and `const voteTx -> ParsedTx<VoteTx>`, where ParsedTx had the necessary information for creating a BwParsed object and have access to custom implementations doing the good practice of coding against interfaces without violating the Dependency Inversion Principle.

But TS is clever enough to work nicely with ParsedTx and then call the factory (we need a factory because we have many different types of txs) for retrieving the correct implementation and use it.

Your example: 
```
<TxItem key={(item as ProcessedTx).id} item={item as ProcessedTx}
```

This is a hack for keeping the Header happy, remember I am just focused for the moment on the transactions view, but I will change the header pretty soon.

From my point of view, removing the generic is the best way to follow. Code is much simpler now, and again, I did it based on one of your comments.

I created this PR for showing changes, but I should have implemented them directly in the other one.

So, unless you have a very strong opposition I'd rather merge it.

PR #453, opened 2019-08-01
Feature – Upgrade IOV-Core to 0.16.0

webmaster128:


packages/bierzo-wallet/src/utils/test/e2e.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-console */
```
albertandrejev: We also can override [eslint rules](https://eslint.org/docs/user-guide/configuring#example-configuration) for our tests.

webmaster128: Yeah, right. In this case I followed the same pattern that we have so far. Let's think about updating this later on.

PR #462, opened 2019-08-01
Feature - Adapt header to support different TXs

### LINKED ISSUE #461: Adapt Header to support different TX types
Following the same approach as we did for transactions in the Transactions view, the header should be ready to handle different kinds of transactions.

apanizo:
In this PR we adapt the Header component to handle different kinds of transactions.

It closes #461 

Screenshot 2019-08-01 17 24 59


packages/bierzo-wallet/src/logic/transactions/types/BwParserFactory.ts
```diff
@@ -20,6 +20,14 @@ export class BwParserFactory {
     throw new Error('Not supporting generic components yet');
   }
 
+  public static getHeaderRepresentation(tx: any, lastOne: boolean): JSX.Element {
+    if (isSendTransaction({ kind: tx.kind })) {
```
webmaster128: This (like in `getReactComponent` and `getCsvRepresentation` as well) makes assumptions on the internal behaviour of `isSendTransaction`. `isSendTransaction` can easily change to check other fields as well.

Better check:

```ts
if (tx.kind === "bcp/send") {
```

in all cases in this file

packages/bierzo-wallet/src/logic/transactions/types/BwParserFactory.ts
```diff
@@ -9,7 +9,7 @@ import {
 
 import { BwParser, ParsedTx } from '../types/BwParser';
 import { BwSendParser } from './BwSendTransaction';
-import { BwSendProps } from './BwSendTransaction/ui';
+import { BwSendProps } from './BwSendTransaction';
```
webmaster128: two imports from the same file

PR #463, opened 2019-08-01
Feature - Add unknown row component

### LINKED ISSUE #404: /transactions: Create component to show Unsupported transaction type
The transactions list in /transactions should show every transaction that belongs to the user. This can easily be 20 different transaction types in the near future. Since we cannot have components for all of them at the moment, we should have a component that represents _everything else_.

Data to be displayed:
* Transaction ID with link to block explorer
* The `LightTransaction.kind` string
* Success or error

apanizo:
**Description**
This PR creates the react implementation of the Unknown TX for its variant in the header and in the transactions view.

Note the transaction Id is going to be always shown (the id is obtained in the subscription, and therefore I had to mock it in storybook as a hardcoded value)

It closes #404 

Screenshot 2019-08-02 12 40 46
Screenshot 2019-08-02 12 45 04

### DIFF
--- a/packages/bierzo-wallet/src/logic/transactions/types/BwParserFactory.ts
+++ b/packages/bierzo-wallet/src/logic/transactions/types/BwParserFactory.ts
@@ -7,33 +7,38 @@ import {
   LightTransaction,
 } from '@iov/bcp';
 
+import { ProcessedSendTransaction } from '../../../store/notifications';
 import { BwParser, ProcessedTx } from '../types/BwParser';
-import { BwSendParser, BwSendProps } from './BwSendTransaction';
-import { BwUnknownProps, BwUnkownParser } from './BwUnkownTransaction';
+import { BwSendParser } from './BwSendTransaction';
+import { BwUnkownParser } from './BwUnkownTransaction';
+
+function isProcessedSendTransaction(tx: ProcessedTx): tx is ProcessedSendTransaction {
+  return isSendTransaction(tx.original);
+}
 
 export class BwParserFactory {
-  public static getReactComponent(tx: any): JSX.Element {
-    if (tx.kind === 'bcp/send') {
-      return new BwSendParser().graphicalRepresentation(tx as BwSendProps);
+  public static getReactComponent(tx: ProcessedTx): JSX.Element {
+    if (isProcessedSendTransaction(tx)) {
+      return new BwSendParser().graphicalRepresentation(tx);
     }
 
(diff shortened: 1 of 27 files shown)


packages/bierzo-wallet/src/logic/transactions/types/BwParserFactory.ts
```diff
@@ -20,6 +20,14 @@ export class BwParserFactory {
     throw new Error('Not supporting generic components yet');
   }
 
+  public static getHeaderRepresentation(tx: any, lastOne: boolean): JSX.Element {
+    if (isSendTransaction({ kind: tx.kind })) {
```
webmaster128: Check like discussed in other PR

packages/bierzo-wallet/src/logic/transactions/types/BwUnkownTransaction/index.tsx
```diff
@@ -0,0 +1,45 @@
+import { Address, BlockchainConnection, ConfirmedTransaction, SendTransaction } from '@iov/bcp';
+import * as React from 'react';
+
+import { BwParser, ParsedTx } from '../../types/BwParser';
+import UnkownTransactionHeader from './ui/UnknownTxHeader';
+import UnkownTransactionRow from './ui/UnknownTxRow';
+
+export interface BwUnknownProps extends ParsedTx {
+  readonly id: string;
+  readonly received: boolean;
```
webmaster128: Most unknown transactions do not have a direction like `SendTransaction`s. `RegisterUsernameTx` for example just has

```
  readonly kind: "bns/register_username";
  readonly username: string;
  readonly targets: readonly ChainAddressPair[];
  readonly fee?: Fee;
```

 Do we need this field? Would be easier if we can get rid of it for now

packages/bierzo-wallet/src/logic/transactions/types/BwParserFactory.ts
apanizo: Yep, I need to rebase.

packages/bierzo-wallet/src/logic/transactions/types/BwUnkownTransaction/index.tsx
apanizo: I believe I can remove it, added because it was useful for some render logic, but I will figure out something. Good point. Thanks.

webmaster128: If you need an intermediate step, set it to `false` by default since almost all transactions we list there are signed by the user

PR #464, opened 2019-08-02
Feature #440 - Receive payment view

### LINKED ISSUE #440: Create "Receive payments" screen
Right now, the in the wallet the user has no way to see their native addresses for the use in other tools. "Receive payments" should contain a list with all addresses.

This should be implemented as a table like

| Blockchain | Address |
| ----------| ----------| 
| IOV Lovenet | tiov14mvqq924qdmpuqnqm3sqjs2c5l5g796nhpcc38 |
| Rinkeby | 0xAed800155503761E0260dC60094158A7e88f1753 |
| Lisk Testnet | 2136030286917746303L |

Requires:
* Looing through all available identities
* Derive address using `(await getCodecForChainId(chainId)).identityToAddress(identity)`
* Copying and use the blockchain name getter `getChainName(chainId: ChainId)` from sanes-chrome-extension
* Sort rows by blockchain name

albertandrejev:
**Description**
This PR will show list of available addresses for the profile, It will close #440.

**Before**
N/A

**After**
receive-payment

### DIFF
--- a/packages/bierzo-wallet/src/routes/receivePayments/index.e2e.spec.ts
+++ b/packages/bierzo-wallet/src/routes/receivePayments/index.e2e.spec.ts
@@ -64,7 +64,6 @@ withChainsDescribe('E2E > Receive Payment route', () => {
 
     [chainName, address] = await getAddressRow(page, 5);
     expect(chainName).toBe('Lisk Devnet');
-    expect(address.length).toBe(20);
   }, 35000);
 
   it('should copy address to clipboard and show toast with message', async () => {
--- a/packages/bierzo-wallet/src/routes/receivePayments/index.e2e.spec.ts
+++ b/packages/bierzo-wallet/src/routes/receivePayments/index.e2e.spec.ts
@@ -1,12 +1,20 @@
+import clipboardy from 'clipboardy';
 import express, { Request, Response } from 'express';
 import { Server } from 'http';
 import { Browser, Page } from 'puppeteer';
 
-import { createExtensionPage, createPage, launchBrowser } from '../../utils/test/e2e';
+import {
+  closeBrowser,
+  closeToast,
+  createExtensionPage,
+  createPage,
+  getToastMessage,
+  launchBrowser,
+} from '../../utils/test/e2e';
 import { withChainsDescribe } from '../../utils/test/testExecutor';
 import { waitForAllBalances } from '../balance/test/operateBalances';
 import { travelToBalanceE2E } from '../balance/test/travelToBalance';
-import { getAddressRow } from './test/operateReceivePayment';
+import { copyAddress, getAddressRow } from './test/operateReceivePayment';
 import { travelToReceivePaymentE2E } from './test/travelToReceivePayment';
 
 withChainsDescribe('E2E > Receive Payment route', () => {
@@ -34,7 +42,7 @@ withChainsDescribe('E2E > Receive Payment route', () => {
   }, 60000);
 
   afterEach(async () => {
-    //await closeBrowser(browser);
+    await closeBrowser(browser);
   });
 
   afterAll(() => {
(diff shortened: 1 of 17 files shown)


packages/bierzo-wallet/src/routes/receivePayments/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+        addressesMap.push({
+          chainId: token.chainId,
+          chainName,
+          address,
+        });
+      }
+      addressesMap.sort((a: ChainAddressMap, b: ChainAddressMap) => {
+        if (a.chainName > b.chainName) return 1;
+        else if (a.chainName < b.chainName) return -1;
+        else return 0;
```
webmaster128: [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) is your friend:

```ts
addressesMap.sort((a: ChainAddressMap, b: ChainAddressMap) => a.chainName.localeCompare(b.chainName))
```

packages/bierzo-wallet/src/routes/receivePayments/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+const ReceivePayment = (): JSX.Element => {
+  const tokens = ReactRedux.useSelector((state: RootState) => state.tokens);
+  const pubKeys = ReactRedux.useSelector((state: RootState) => state.extension.keys);
+
+  const [chainAddressMap, setChainAddressMap] = React.useState<ChainAddressMap[]>([]);
+
+  React.useEffect(() => {
+    async function processAddresses(
+      tokens: { [key: string]: BwToken },
+      pubKeys: { [chain: string]: string },
```
webmaster128: You don't need the tokens here. You can just use the pubkeys and loop over the identities. An Identity is always a chainId/pubkey pair:

```ts
for (const plainPubkey of Object.values(pubKeys)) {
  const identity: Identity = TransactionEncoder.fromJson(JSON.parse(plainPubkey));
  const address = (await getCodecForChainId(identity.chainId)).identityToAddress(identity);
  const chainName = getChainName(identity.chainId);
}
```

packages/bierzo-wallet/src/routes/receivePayments/index.e2e.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+    let [chainName, address] = await getAddressRow(page, 1);
+    expect(chainName).toBe('Ganache');
+    expect(address.length).toBe(42);
+
+    [chainName, address] = await getAddressRow(page, 4);
+    expect(chainName).toBe('IOV Devnet');
+    expect(address.length).toBe(43);
+
+    [chainName, address] = await getAddressRow(page, 5);
+    expect(chainName).toBe('Lisk Devnet');
```
webmaster128: Better use a pattern check instead of length for all address types, because
* It checks that address is a string
* Much stricter
* Easy to read (well, this one is probably a bit subjective)
* Works for address types with variable length

with (untested)

* Ethereum: `.toMatch(/^0x[a-fA-F0-9]{40}$/)`
* IOV: `.toMatch(/^tiov1[0-9a-f]{38}$`
* Lisk: `.toMatch(/^[0-9]+L$/)`

packages/bierzo-wallet/src/routes/receivePayments/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+}
+
+const ReceivePayment = (): JSX.Element => {
+  const pubKeys = ReactRedux.useSelector((state: RootState) => state.extension.keys);
+
+  const [chainAddressMap, setChainAddressMap] = React.useState<ChainAddressMap[]>([]);
+
+  React.useEffect(() => {
+    async function processAddresses(pubKeys: { [chain: string]: string }): Promise<void> {
+      if (!pubKeys) return;
```
webmaster128: This check sould not be necessary. `pubKeys` type does not allow a falsy value

albertandrejev: Yes, but check fails because of this.

albertandrejev:
How can I get list of available tickers/address for blockchain used in Travis? Because tests fails

packages/bierzo-wallet/src/routes/receivePayments/index.tsx
albertandrejev: Anyway this was fixed

PR #466, opened 2019-08-05
Feature – Refactor process addresses

webmaster128:
Some minor code adjustments that do not change functionality

### DIFF
--- a/packages/bierzo-wallet/src/routes/receivePayments/index.tsx
+++ b/packages/bierzo-wallet/src/routes/receivePayments/index.tsx
@@ -8,7 +8,7 @@ import { getChainName } from '../../config';
 import { getCodecForChainId } from '../../logic/codec';
 import { RootState } from '../../store/reducers';
 import { BALANCE_ROUTE } from '../paths';
-import Layout, { ChainAddressMap } from './components';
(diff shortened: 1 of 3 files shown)


packages/bierzo-wallet/src/routes/receivePayments/index.tsx
```diff
@@ -17,38 +17,34 @@ function onReturnToBalance(): void {
...
-        }
-
-        const identity: Identity = TransactionEncoder.fromJson(JSON.parse(plainPubkey));
-        const address = (await getCodecForChainId(chainId)).identityToAddress(identity);
-        const chainName = getChainName(chainId);
-        addressesMap.push({
-          chainId: chainId,
-          chainName,
-          address,
+      const identities = Object.values(pubKeys).map(
```
albertandrejev: If you will get me code and remove unnecessary lines you will get it at least 6 lines shorter, without unnecessary `.map` loop.

webmaster128: Yeah, I'm generally not so happy with all thise raw pubkey maps in the codebase. It should be even easier to get the list of identities from redux. But one step at a time

PR #474, opened 2019-08-05
Feature #458 - Load wallet config from web resource

### LINKED ISSUE #458: Make it possible to change configuration without recompiling

albertandrejev:
**Description**
This PR will load wallet configuration from web resource and refactor all occurrences of `getConfig()` method. It will close #458.


packages/bierzo-wallet/src/setupTests.js
```diff
@@ -0,0 +1,4 @@
+export const window = global.window || {};
+
+global.window = window;
+global.config = require('../build/static/config/conf.json');
```
webmaster128: What about always loading the `./config/development.json` from the source repo? Tests will always need this one. This way, the tests can run undependent on the `./build` directory (at least unit and dom tests).

albertandrejev: ok, no problems

webmaster128:
@albertandrejev can you rebase on master and run `yarn install && yarn format` to get the new prettier settings? Thanks!

albertandrejev:
I already did, but will check again.

packages/bierzo-wallet/package.json
```diff
@@ -39,22 +39,25 @@
     "clipboardy": "^2.1.0"
   },
   "scripts": {
+    "use-extension-config-development": "cd ../sanes-chrome-extension && yarn use-config-development && yarn override-config-development && cd ../bierzo-wallet",
```
webmaster128: I think `use-extension-config-development` should only run for e2e tests. The other tests must not touch the extension directory

packages/sanes-chrome-extension/package.json
```diff
@@ -14,7 +14,7 @@
     "build-staging": "yarn use-config-staging && yarn build-with-any-config && yarn make-manifest-staging",
     "build-production": "yarn use-config-production && yarn build-with-any-config && yarn make-manifest-production",
     "build": "yarn build-development",
-    "override-config-development": "cp ./config/development.json ./build/assets/config/conf.json && yarn make-manifest-development",
+    "override-config-development": "mkdir -p ./build/assets/config && cp ./config/development.json ./build/assets/config/conf.json && yarn make-manifest-development",
```
webmaster128: Why this change? `override-config-development` is used to override the config of an existing build directory. When this is called before `yarn build`, this is a bug in the code that calls it.

packages/bierzo-wallet/package.json
webmaster128: Fixing this probably fixes all the problem. Then the additional dependencies is not needed since `use-extension-config-development` is not called during the build but only when running the e2e test.

packages/sanes-chrome-extension/package.json
albertandrejev: You are right I need to use this `override` only during e2e test, so this code will go away. I added this because `bierzo-wallet` was built before `extension`. Will fix now.

PR #475, opened 2019-08-05
Feature – Upgrade final-form

### LINKED ISSUE #473: Upgrade final-form
This causes a lot of compilation errors

webmaster128:
`FieldValidator` is now a generic, so we need to adapt some of our code.

Closes #473


packages/bierzo-wallet/src/routes/payment/components/ReceiverAddress/index.tsx
```diff
@@ -18,8 +18,8 @@ interface Props {
 }
 
 const ReceiverAddress = ({ form, selectedChainCodec }: Props): JSX.Element => {
-  const recipientValidator: FieldValidator = React.useMemo(() => {
-    const validator = (value: string): string | undefined => {
+  const recipientValidator: FieldValidator<string | undefined> = React.useMemo(() => {
```
albertandrejev: You can use `InputValue` here as well, if I understand you correctly.

packages/sanes-chrome-extension/src/routes/restore-account/components/SetMnemonicForm.tsx
```diff
@@ -14,11 +14,9 @@ import { RESTORE_ACCOUNT } from "../../paths";
 
 export const MNEMONIC_FIELD = "mnemonicField";
 
-const englishMnemonicValidator: FieldValidator = (value): string | undefined => {
-  if (typeof value !== "string") throw new Error("Input must be a string");
-
+const englishMnemonicValidator: FieldValidator<string | undefined> = (value): string | undefined => {
```
albertandrejev: `InputValue` can be used here too.

packages/bierzo-wallet/src/routes/payment/components/ReceiverAddress/index.tsx
webmaster128: Right, thanks, will update

packages/sanes-chrome-extension/src/routes/restore-account/components/SetMnemonicForm.tsx
webmaster128: Right, thanks, will update

PR #482, opened 2019-08-06
Feature – Fix and improve config handling

webmaster128:
Before this `react-scripts build` cleared the whole build folder and the config file was lost.


packages/bierzo-wallet/package.json
```diff
@@ -40,23 +40,20 @@
...
-    "test-unit": "yarn use-config-development && CI=true react-scripts test '\\.unit\\.(spec|test)\\.[jt]s[x]?$' --env=jsdom --detectOpenHandles",
-    "test-e2e": "yarn use-config-development && yarn use-extension-config-development && CI=true react-scripts test '\\.e2e\\.(spec|test)\\.[jt]s[x]?$' --env=jsdom --runInBand",
-    "test-single": "yarn use-config-development && yarn use-extension-config-development && CI=true react-scripts test --env=jsdom --detectOpenHandles",
+    "test-dom": "yarn override-config-development && CI=true react-scripts test '\\.dom\\.(spec|test)\\.[jt]s[x]?$' --env=jsdom",
+    "test-unit": "yarn override-config-development && CI=true react-scripts test '\\.unit\\.(spec|test)\\.[jt]s[x]?$' --env=jsdom --detectOpenHandles",
+    "test-e2e": "yarn override-config-development && yarn use-extension-config-development && CI=true react-scripts test '\\.e2e\\.(spec|test)\\.[jt]s[x]?$' --env=jsdom --runInBand",
+    "test-single": "yarn override-config-development && yarn use-extension-config-development && CI=true react-scripts test --env=jsdom --detectOpenHandles",
```
albertandrejev: We do not need to override configs because of:
```
if (process.env.NODE_ENV === "test") {
    return (window as WindowWithConfig).developmentConfig;
  }
```

webmaster128: Exactly. The config file in build/ was not used anyways. However, we need to keep it in the e2e tests that start a webserver serving ./build data

albertandrejev: You can leave it for e2e tests only.

webmaster128: Updated in https://github.com/iov-one/ponferrada/pull/482/commits/c649e4e6ecc5b8f0d3f025cfc03f298b246c1d07

PR #485, opened 2019-08-06
Feature #477 - Request Identity from Extension

### LINKED ISSUE #477: Request identity from extension

abefernan:
**Description**
This PR makes Redux store the identity that the current user has on the BNS chain, and also makes the loading of the BNS Chain come from a Config file.

Closes #477.

### DIFF
--- a/packages/sil-voting-app/src/store/extension/actions.ts
+++ b/packages/sil-voting-app/src/store/extension/actions.ts
@@ -10,7 +10,7 @@ export async function getExtensionStatus(): Promise<ExtensionState> {
     return { installed: false, connected: false, identity: undefined };
   }
 
-  if (Object.keys(identities).length === 0) {
+  if (identities.length === 0) {
     return { installed: true, connected: false, identity: undefined };
   }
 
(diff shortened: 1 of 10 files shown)


packages/sil-voting-app/src/store/extension/actions.ts
```diff
@@ -1,27 +1,33 @@
...
 
 export async function getExtensionStatus(): Promise<ExtensionState> {
   const identities = await sendGetIdentitiesRequest();
 
   if (!identities) {
-    return { installed: false, connected: false };
+    return { installed: false, connected: false, identity: undefined };
   }
 
   if (Object.keys(identities).length === 0) {
```
webmaster128: This is an issue in the old code: `identities` is an array. We should simply use `if (identities.length === 0) {`

packages/sil-voting-app/src/store/extension/reducer.ts
```diff
@@ -6,6 +6,7 @@ import * as actions from "./actions";
 export interface ExtensionState {
   readonly connected: boolean;
   readonly installed: boolean;
+  readonly identity: string | undefined;
```
webmaster128: Exactly!

PR #488, opened 2019-08-06
Feature #249 - Create username page

### LINKED ISSUE #249: Add a Create Username page
In the demo wallet the second step after registration is to create a human readable address.

This view will be adapted but the routing to this page from the sign-up is no longer needed.

- [x] Prepare view and storybook (PR #488)
- [ ] Implement register username request logic (PR #505)
- [ ] Add transaction item component for username registration tx
- [ ] E2E tests

albertandrejev:
**Description**
This PR will add username creation page. It will implement "Prepare view and storybook" part of the #249.

**Before**
N/A

**After**
reg_username


packages/bierzo-wallet/src/routes/setName/components/index.tsx
```diff
@@ -0,0 +1,109 @@
...
+              </Back>
+            </Block>
+          </Block>
+        </Block>
+      </Form>
+    </Block>
+  );
+};
+
+export default SetName;
```
webmaster128: Can you rename the component (and folder) to `RegisterUsername`? "Set name" sounds like you can do that only once. But in theory a user can register as many names as they want

packages/valdueza-storybook/.storybook/config.js
```diff
@@ -20,17 +20,20 @@ addParameters({
...
-// https://github.com/abrcdf1023/egroup-material/commit/a876a4b411fda490a53ee20179acc72fe6514aaa
-addDecorator(story => {
-  const generateClassName = (rule, styleSheet) => `${styleSheet.options.classNamePrefix}-${rule.key}`;
-  return (
-    <StylesProvider generateClassName={generateClassName}>
-      {story()}
-    </StylesProvider>
-  );
-});
+// We need to add decorator for testing only, otherwise we will break visual control of the storybooks
```
webmaster128: 👍

webmaster128: This is now fixed on master: https://github.com/iov-one/ponferrada/pull/491 Thanks for the patch!

.gitignore
```diff
@@ -3,3 +3,4 @@ yarn-error.log
 
 # macOS
 .DS_Store
+.yarnrc
```
webmaster128: You can also check-in this file. Then the registry setting is available for everyone, no matter what the developer's user config is

albertandrejev:
This PR contain view implementation only, logic will be implement in next PR.

packages/bierzo-wallet/src/routes/registerName/components/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+                  form={form}
+                  placeholder="Register IOV username"
+                  fullWidth
+                  margin="none"
+                />
+              </Block>
+            </Block>
+          </Paper>
+
+          <Paper>
```
webmaster128: Can we write both parts into the same `<Paper>`? This makes it clear that the addresses belong to the exact same operation

packages/bierzo-wallet/src/routes/registerName/components/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+              padding={3}
+            >
+              <Avatar classes={avatarClasses}>
+                <FontAwesomeIcon icon={faRegistered} color="#ffffff" />
+              </Avatar>
+              <Block width="100%" marginTop={2} marginBottom={1}>
+                <TextFieldForm
+                  name={SET_USERNAME_FIELD}
+                  form={form}
+                  placeholder="Register IOV username"
```
webmaster128: What about `username*iov` as a placeholder? The user will need to type-in `*iov` as well, so this would make it easier to see what is expected

PR #491, opened 2019-08-06
Alter storybook configuration

willclarktech:
Cherry-picked from https://github.com/iov-one/ponferrada/tree/feature/%23249-create-username


willclarktech:
@webmaster128 I don't know in detail how it's supposed to look, but with this change it certainly looks better than before!

webmaster128:
Perfect. The decorator [created here](https://github.com/iov-one/ponferrada/pull/467) artifically removed the IDs from css classes. This is needed to avoid massive unnecessary diffs in the snapshot testing. If that decorator is disabled, everything works the default way.

PR #494, opened 2019-08-07
Fix - Add babel-loader to storybook

albertandrejev:
**Description**
This PR will add `babel-loader` package to storybook.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,3 @@ yarn-error.log
 
 # macOS
 .DS_Store
-.yarnrc
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ yarn-error.log
 
 # macOS
 .DS_Store
+.yarnrc
--- a/packages/valdueza-storybook/package.json
+++ b/packages/valdueza-storybook/package.json
@@ -17,7 +17,8 @@
     "extends": "react-app"
   },
   "dependencies": {
-    "medulas-react-components": "^0.2.0"
+    "medulas-react-components": "^0.2.0",
+    "babel-loader": "^8.0.6"
   },
   "devDependencies": {
     "@iov/browser-extension-react-scripts": "^0.2.1",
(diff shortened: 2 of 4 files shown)


packages/valdueza-storybook/package.json
```diff
@@ -17,7 +17,8 @@
     "extends": "react-app"
   },
   "dependencies": {
-    "medulas-react-components": "^0.2.0"
+    "medulas-react-components": "^0.2.0",
+    "babel-loader": "^8.0.6"
```
webmaster128: Good! Context: babel-loader was removed as an explicit dependency from the root package.json in https://github.com/iov-one/ponferrada/pull/484/commits because the latest version of react-scripts complained and rightfully noted, that we should not install babel-loader on our own. Instead it is pulled in as a depdendecy from `react-scripts` and `@iov/browser-extension-react-scripts`. But here we need it as a runtime dependency for webpack to find it.

yarn.lock
```diff
@@ -4599,6 +4599,16 @@ babel-loader@8.0.5:
     mkdirp "^0.5.1"
     util.promisify "^1.0.0"
 
+babel-loader@^8.0.6:
+  version "8.0.6"
+  resolved "https://registry.npmjs.org/babel-loader/-/babel-loader-8.0.6.tgz#e33bdb6f362b03f4bb141a0c21ab87c501b70dfb"
```
webmaster128: Can you please use `https://registry.yarnpkg.com` as a registry? When you cannot change your local yarn config, can you add a `.yarnrc` in this project [as described here](https://stackoverflow.com/a/46000761/2013738)?

.gitignore
```diff
@@ -3,3 +3,4 @@ yarn-error.log
 
 # macOS
 .DS_Store
+.yarnrc
```
webmaster128: This file will be removed on the next `git clean -xdf`. Feel free to check it in instead

PR #497, opened 2019-08-07
Feature #478 - Integrate Governor Class

### LINKED ISSUE #478: Integrate Govenor class from @iov/bns-governance

abefernan:
**Description**
This PR adds a `BNS Governance` `Governor` instance to Redux in order to get and create proposals in the future.

The `Identity` stored in Redux has been removed because the only use case for it will be for signing transactions, and @willclarktech pointed out the Tx itself holds its creator's `Identity`.

I tested the creation and method calling of the `Governor` in `store/proposals/actions.ts` -> `getProposals()`. When testing `governor.getProposals()` I noticed that the extension asks for confirmation _twice_ before redirecting to the `/dashboard` route, I don't know if this is intended or an issue.

Closes #478.


webmaster128:
> When testing `governor.getProposals()` I noticed that the extension asks for confirmation _twice_ before redirecting to the `/dashboard` route, I don't know if this is intended or an issue.

Not intended. There should ne only one getIdentities request per session. Could you check if

```
const onLogin = async (): Promise<void> => {
    if (await isExtensionConnected()) {
      await loadProposals();
      history.push(DASHBOARD_ROUTE);
    }
  };
```

is called multiple times?

abefernan:
> ```
> const onLogin = async (): Promise<void> => {
> ```
> 
> is called multiple times?

It's only called once, I checked with `console.count()`.

PR #498, opened 2019-08-07
Feature – Small refinements for extension router

webmaster128:
First steps towards #257. Unfortunately it was not trivial to get the redux store out of the router.

### DIFF
--- a/packages/sanes-chrome-extension/src/utils/test/navigation.ts
+++ b/packages/sanes-chrome-extension/src/utils/test/navigation.ts
@@ -18,25 +18,35 @@ import { REQUEST_FIELD } from "../../routes/requests/components/RequestList";
 import { history } from "../../store/reducers";
 import { createDom } from "../../utils/test/dom";
 
-const MAX_TIMES_EXECUTED = 35;
-const INTERVAL = 500;
+const retryInterval = 400;
+
+/**
+ * Calls callback until it returns true or timeout is reachend.
+ *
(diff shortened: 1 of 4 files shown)


packages/sanes-chrome-extension/src/utils/test/navigation.ts
```diff
@@ -18,25 +18,35 @@ import { REQUEST_FIELD } from "../../routes/requests/components/RequestList";
...
+const retryInterval = 400;
+
+/**
+ * Calls callback until it returns true or timeout is reachend.
+ *
+ * @param callback a callback function that is called in a loop. It should be cheap to execute.
+ * @param timeout timeout in milliseconds. Defaults to 10 seconds
+ * @returns a promise that resolves when the callback returned true and is rejected when timeout is reached.
+ */
+export function whenTrue(callback: () => boolean | Promise<boolean>, timeout: number = 10000): Promise<void> {
```
albertandrejev: Maybe if we were removed constant and passing it as function argument (`timeout`) then let's passing `retryInterval` as argument too?

webmaster128: Yeah, I was thinking about that too. I did not include it since at the moment we don't need flexibility.

I think at some point soon we will deduplicate such low-level logic functions in it's own package that is available for all apps. Then we can make them more flexible and add tests as well. But before that we need Adolfo to find a new package name :)

PR #499, opened 2019-08-07
Deploy BNSD proposals for development

### LINKED ISSUE #495: Introduce Proposals on the Blockchain
In order to test the listing of proposals before implementing its creation, it's needed that the blockchain has some initial proposals to do the testing.

willclarktech:
Closes #495

Just put it together in JS for this initial PoC because we need to decide on how to structure this. E.g. `package.json`, TS infrastructure etc?

Edit: Updated to copy across the `scripts/bnsd` directory from IOV-Core after https://github.com/iov-one/iov-core/pull/1178

### DIFF
--- a/scripts/bnsd/deploy_proposals.js
+++ b/scripts/bnsd/deploy_proposals.js
@@ -1,0 +1,126 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+const { isBlockInfoPending, isBlockInfoSucceeded } = require("@iov/bcp");
+const { bnsCodec, BnsConnection, VoteOption } = require("@iov/bns");
+const { /* committeeIds, guaranteeFundEscrowIds, */ Governor, ProposalType } = require("@iov/bns-governance");
+const { Ed25519HdWallet, HdPaths, UserProfile } = require("@iov/keycontrol");
+
+// Dev admin
+// path: m/44'/234'/0'
+// pubkey: 418f88ff4876d33a3d6e2a17d0fe0e78dc3cb5e4b42c6c156ed1b8bfce5d46d1
+// IOV address: tiov15nuhg3l8ma2mdmcdvgy7hme20v3xy5mkxcezea
+// This account has money in the genesis file (see scripts/bnsd/README.md).
+const adminMnemonic = "degree tackle suggest window test behind mesh extra cover prepare oak script";
+const adminPath = HdPaths.iov(0);
+const committeeId = 2;
+const electionRuleId = 2;
+const bnsdUrl = "ws://localhost:23456";
+const connectionPromise = BnsConnection.establish(bnsdUrl);
+
+function createSignAndPoster(connection, profile) {
+  return async function signAndPost(tx) {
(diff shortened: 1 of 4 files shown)


scripts/bnsd/start.sh
```diff
@@ -25,3 +25,10 @@ sleep 3
 # for debug output
 cat "$LOGS_FILE_TM"
 cat "$LOGS_FILE_APP"
+
+if [ -n "${INIT_PROPOSALS:-}" ]; then
+  sleep 3
+  echo "Deploying proposals..."
+  node "./deploy_proposals.js"
+  echo "Done."
+fi
```
webmaster128: Can we move this into an explicit init.sh please (like in `scripts/ethereum/`)?

scripts/bnsd/genesis_app_state.json
```diff
@@ -89,6 +89,25 @@
...
+          "whole": 100000000000000
+        }
+      ],
+      "arbiter": "cond:gov/rule/0000000000000002",
+      "//destination-name": "Reward Fund",
+      "destination": "cond:gov/rule/0000000000000002",
+      "source": "0000000000000000000000000000000000000000",
+      "timeout": "2999-12-31T00:00:00Z"
+    }
+  ],
```
webmaster128: Is it possible to sync this genesis with IOV-Core in a way that we can copy the entire `./bnsd/` folder from IOV-Core to the ponferrada repo? This has been the process so far and it saves a lot of time.

scripts/bnsd/deploy_proposals.js
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }
+}
+
+main()
+  .catch(console.error)
+  .finally(async () => {
+    (await connectionPromise).disconnect();
+    process.exit(0);
+  });
```
webmaster128: Since we don't have `ts-node` installed yet in this repo, I think it is good to have this pure .js

scripts/bnsd/genesis_app_state.json
willclarktech: You mean close this PR and open an equivalent one in IOV-Core instead?

webmaster128: I don't mind in which way the sync is achieved. I would just be very happy if at the end of this process both repos have BNS governance tests that pass using the same genesis file.

willclarktech: Oh, well all I did in this commit was copy and paste the genesis file over: https://github.com/iov-one/ponferrada/pull/499/commits/27eed130e500322aca2dca0f90797fdad61e35d4

I think it's easiest if we keep IOV-Core as the single source of truth though, and just periodically copy and paste the whole directory.

willclarktech: https://github.com/iov-one/iov-core/pull/1178

webmaster128: In this case we are on the same page already, perfect

PR #503, opened 2019-08-07
Feature – Refator incoming/outgoing flags in wallet

### LINKED ISSUE #496: Use consistent incoming/outgoing flags for SendTransactions
In extension and wallet, the interpretation of incoming/outgoing (also called `received`) is inconsistent. The data model should be changed in a way that

```ts
export interface ProcessedSendTransaction extends ProcessedTx<SendTransaction> {
  readonly incoming: boolean; // recipient == current user
  readonly outgoing: boolean; // sender == current user
}
```

This way the the data structure supports send transactions that are both incoming and outgoing much better and it is up to the user interface to interprete all possible states.

webmaster128:
Closes #496

**Before**

before

**After**

after

### DIFF
--- a/packages/bierzo-wallet/src/components/PageMenu/index.stories.tsx
+++ b/packages/bierzo-wallet/src/components/PageMenu/index.stories.tsx
@@ -7,6 +7,7 @@ import { DeepPartial } from "redux";
 
 import { ProcessedSendTransaction } from "../../store/notifications";
 import { RootState } from "../../store/reducers";
+import { stringToAmount } from "../../utils/balances";
 import DecoratedStorybook, { WALLET_ROOT } from "../../utils/storybook";
 import PageMenu from "./index";
 
@@ -62,8 +63,22 @@ const faultTx: ProcessedSendTransaction = {
   outgoing: true,
 };
 
+const incomingAndOutgoingSendTransaction: ProcessedSendTransaction = {
+  time: new ReadonlyDate("2019-12-24T04:35:03.763Z"),
+  id: "EDBBA9C7C558A60E09A589C2263CF5DDC7B25ED014E3EF5959C6B1C8E6DBAD4E" as TransactionId,
+  original: {
+    kind: "bcp/send",
+    sender: "tiov1xgm95mecmf3vkn7lnszfe9q4uy6nv0pwkr8wc3" as Address,
+    recipient: "tiov1xgm95mecmf3vkn7lnszfe9q4uy6nv0pwkr8wc3" as Address,
+    amount: stringToAmount("7.4", "IOV" as TokenTicker),
+    memo: "Send money to myself for fun",
+  },
+  incoming: true,
+  outgoing: true,
+};
+
 const fullStore = (): DeepPartial<RootState> => {
-  const fullTxs = [faultTx, ...txs];
+  const fullTxs = [faultTx, incomingAndOutgoingSendTransaction, ...txs];
 
   return {
     notifications: {
--- a/packages/bierzo-wallet/src/components/PageMenu/index.stories.tsx
+++ b/packages/bierzo-wallet/src/components/PageMenu/index.stories.tsx
@@ -12,7 +12,7 @@ import PageMenu from "./index";
 
 const txs: readonly ProcessedSendTransaction[] = [
   {
-    received: true,
+    incoming: true,
     time: new ReadonlyDate("2018-12-24T10:51:33.763Z"),
     id: "tx1" as TransactionId,
     original: {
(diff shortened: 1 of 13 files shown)


packages/bierzo-wallet/src/components/PageMenu/index.stories.tsx
```diff
@@ -57,6 +58,8 @@ const faultTx: ProcessedSendTransaction = {
     sender: "tiov1dcg3fat5zrvw00xezzjk3jgedm7pg70y222af3" as Address,
     recipient: "tiov1k898u78hgs36uqw68dg7va5nfkgstu5z0fhz3f" as Address,
   },
+  incoming: false,
+  outgoing: true,
 };
```
willclarktech: What about stories where `incoming` and `outgoing` are both true here as well?

packages/bierzo-wallet/src/logic/transactions/types/BwSendTransaction/index.tsx
```diff
@@ -38,7 +37,7 @@ export class BwSendParser extends BwParser<ProcessedSendTransaction> {
       `"${original.amount.tokenTicker}"`,
     ];
     const date = [`"${tx.time.toISOString()}"`];
-    const status = [`"${tx.received}"`, `"${original.memo}"`];
+    const status = [`"${original.memo}"`];
 
     const txRow = [...parties, ...payment, ...date, ...status];
```
willclarktech: I'm not convinced all these arrays/destructuring make things any clearer.

webmaster128: I fully agree. However, the whole CSV idea needs a big rework since we want to support different kind of transactions in the same table. I'm resolving this as unrelated for now.

webmaster128:
> The idea is that if I'm sending funds to myself I'm going to want to see specifically which addresses the funds were sent from/to?

First idea is to display transactions that are both incoming and outgoing as outgoing (like in the extension).

Displaying both sender and recipient consistently for all kind of send transactions in the details area is slightly unrelated to the core of this PR. But I found the alternating content structure in that box very confusing. It allows the user to verfy which of their addresses was used. This is not something for the front page but should definitely be available in the details view in my opinion.

> It's a bit difficult to recognise visually that the addresses are my own right now.

I agree. I'd prefer to have some kind of _(mine)_ marker for addresses that are mine (both sender and recipient). What do you think?

> Side note: the font showing the addresses/memo is very faint! This can be a serious accessibility issue.

Jap. This design will not live very long anyways and big redesign discussions are comming after launch.

packages/bierzo-wallet/src/components/PageMenu/index.stories.tsx
webmaster128: Added in https://github.com/iov-one/ponferrada/pull/503/commits/0f6e665e8955732173197fe7bbe8357a3bb31caf. There is no storybook update since the transactions in the menu are only rendered when clicking the menu. But it is good to see it when testing manually.

packages/bierzo-wallet/src/routes/transactions/components/sorting.ts
```diff
@@ -6,12 +6,12 @@ export interface SortingStateProps {
...
 
-export const ORDER_ASC = 1;
-export const ORDER_DESC = -1;
 export type SortOrder = 1 | -1;
+export const ORDER_ASC: SortOrder = 1;
+export const ORDER_DESC: SortOrder = -1;
 
-export const TX_AMOUNT_COLUMN = "Amount";
-export const TX_DATE_COLUMN = "Date";
+export const TX_AMOUNT_COLUMN = "Amount" as const;
```
albertandrejev: What is this for (`as const`)?

webmaster128: When you assign a literal to a variable, TypeScript does type widening, i.e. the variable has a wider type than the literal. E.g.

```ts
const a = "abc"; // type of a is string, not "abc"
const b = 123; // type of b is number, not 123
const c = [1, 2, 3]; // type of c is number[], not readonly [1, 2 3]
```

with the `as const` language construct (new in TS3.4) you can stop this widening, e.g

```ts
const a = "abc" as const; // type of a is "abc"
const b = 123 as const; // type of b is 123
const c = [1, 2, 3] as const; // type of c is readonly [1, 2 3]
```

In this case, the type of `TX_DATE_COLUMN` changed from `string` to `"Date"`. Now `TX_DATE_COLUMN` can be assigned to a variable of type `TxsOrder`, which does not accept general `string`s. `TX_AMOUNT_COLUMN` changes to type `"Amount"`, which can be assigned to `readonly name: "Amount" | "Transactions";`.

albertandrejev: thank you

willclarktech:
> I agree. I'd prefer to have some kind of (mine) marker for addresses that are mine (both sender and recipient). What do you think?

Definitely.

PR #504, opened 2019-08-07
Tidy up a couple of spellings

willclarktech:


webmaster128:
I fixed the `elipsify` method of the wallet package in the other PR as well

PR #505, opened 2019-08-07
Feature #249 - Implement register username request

albertandrejev:
**Description**
This PR will register username request logic. It will implement "Implement register username request logic" part of the #249.

### DIFF
--- a/packages/bierzo-wallet/src/communication/signAndPost/index.ts
+++ b/packages/bierzo-wallet/src/communication/signAndPost/index.ts
@@ -78,8 +78,6 @@ export const generateRegisterUsernameTxRequest = async (
   };
 };
 
-//  const request = await generateSendTxRequest(chainId, creator, recipient, amount, memo);
-
 export const sendSignAndPostRequest = async (request: JsonRpcRequest): Promise<TransactionId | null> => {
   const config = await getConfig();
 
--- a/packages/bierzo-wallet/src/communication/signAndPost/index.ts
+++ b/packages/bierzo-wallet/src/communication/signAndPost/index.ts
@@ -15,7 +15,7 @@ import { isJsonRpcErrorResponse, JsonRpcRequest, makeJsonRpcId, parseJsonRpcResp
 
 import { getConfig } from "../../config";
 import { getCodecForChainId } from "../../logic/codec";
-import { getConnectionForBns, getConnectionForChainId } from "../../logic/connection";
+import { getConnectionForChainId } from "../../logic/connection";
 
 async function withChainFee<T extends UnsignedTransaction>(chainId: ChainId, transaction: T): Promise<T> {
   const connection = await getConnectionForChainId(chainId);
--- a/packages/bierzo-wallet/src/communication/signAndPost/index.ts
+++ b/packages/bierzo-wallet/src/communication/signAndPost/index.ts
@@ -9,12 +9,13 @@ import {
   UnsignedTransaction,
   WithCreator,
 } from "@iov/bcp";
+import { ChainAddressPair, RegisterUsernameTx } from "@iov/bns";
 import { TransactionEncoder } from "@iov/encoding";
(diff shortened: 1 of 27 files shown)


packages/bierzo-wallet/src/routes/registerName/index.tsx
```diff
@@ -1,19 +1,66 @@
...
 const RegisterUsername = (): JSX.Element => {
+  const toast = React.useContext(ToastContext);
+  const pubKeys = ReactRedux.useSelector((state: RootState) => state.extension.keys);
+  async function onSubmit(values: object): Promise<void> {
+    const formValues = values as FormValues;
+
+    const username = formValues[SET_USERNAME_FIELD];
+
+    let bnsIdentity: Identity | null = null;
+    const addresses: ChainAddressPair[] = [];
```
webmaster128: Can you create this `const addresses: ChainAddressPair[]` list only once (when registration component is created) and then pass into AddressesTable as a property? This ensures the UI and the sumitted data is always the same, even if we refactor some of the code at some point in time.

`AddressesTable` can have an `addresses: ChainAddressPair[]` property and then get the chain names on demand as part of the UI logic.

webmaster128: Sorting can remain in the `AddressesTable` component, since the order in the transaction does not matter

albertandrejev:
Not tested yet. Will check it today.

packages/bierzo-wallet/src/communication/signAndPost/index.ts
```diff
@@ -22,16 +23,15 @@ async function withChainFee<T extends UnsignedTransaction>(chainId: ChainId, tra
...
-const generateSignAndPostRequest = async (
-  chainId: ChainId,
+export const generateSendTxRequest = async (
   creator: Identity,
   recipient: Address,
   amount: Amount,
   memo: string | undefined,
 ): Promise<JsonRpcRequest> => {
-  const codec = await getCodecForChainId(chainId);
+  const codec = await getCodecForChainId(creator.chainId);
```
webmaster128: Nice simplification! That chain ID of the creator is always the right one

packages/bierzo-wallet/src/routes/registerName/index.tsx
```diff
@@ -1,22 +1,101 @@
...
 function onCancel(): void {
   history.push(BALANCE_ROUTE);
 }
 
+const validate = async (values: object): Promise<object> => {
+  const formValues = values as FormValues;
+  const errors: ValidationError = {};
+
+  const username = formValues[REGISTER_USERNAME_FIELD];
+  if (!username || username.length === 0) {
```
webmaster128: `username.length === 0` is obsolete since the empty string is falsy already:

```
$ node
> !!""
false
```

packages/bierzo-wallet/src/routes/registerName/index.tsx
```diff
@@ -1,22 +1,101 @@
...
+      return;
+    }
+
+    try {
+      const request = await generateRegisterUsernameTxRequest(bnsIdentity, username, addresses);
+      const transactionId = await sendSignAndPostRequest(request);
+      if (transactionId === null) {
+        toast.show("Request rejected", ToastVariant.ERROR);
+      }
+
```
webmaster128: I think we should also redirect to the transaction success page we have for the send transaction. This should be the default flow for all transaction types.

Updating the username is cool and makes sense. But we can also do this in the transaction subscription code (e.g. update username after every RegisterUsernameTx event we get). Then it independent of the UI flow of the user.

packages/bierzo-wallet/src/routes/balance/components/index.tsx
```diff
@@ -55,7 +56,36 @@ const Card = ({ id, text, logo, onAction }: CardProps): JSX.Element => {
...
+      variant="h5"
+      align="center"
+      color="primary"
+      weight="light"
+      inline
+      link
+      onClick={onRegisterUsername}
+    >
+      address.
+    </Typography>
```
webmaster128: This split in two h5 component break tests at the moment

PR #506, opened 2019-08-07
Feature #479 - List Proposals with Governor

### LINKED ISSUE #479: List proposals from blockchain via Governor instance
Make sure identity is not requested twice on login.

abefernan:
**Description**
This PR makes Redux's `getProposals()` method get the proposals from the `Governor` instance.

Closes #479.

### DIFF
--- a/packages/sil-voting-app/src/store/proposals/actions.ts
+++ b/packages/sil-voting-app/src/store/proposals/actions.ts
@@ -1,7 +1,6 @@
 import { Proposal, ProposalStatus, VoteOption } from "@iov/bns";
 import { Governor } from "@iov/bns-governance";
 
-import { getDummyVote } from "./dummyData";
 import { AddProposalsActionType, ProposalsState } from "./reducer";
 
 export async function getProposals(governor: Governor | undefined): Promise<ProposalsState> {
@@ -33,8 +32,15 @@ export async function getProposals(governor: Governor | undefined): Promise<Prop
     return threshold;
   };
 
-  //TODO Get current vote from blockchain
-  const getVote = (proposal: Proposal): VoteOption | undefined => getDummyVote(proposal);
+  const getVote = async (proposal: Proposal): Promise<VoteOption | undefined> => {
+    const votes = await governor.getVotes();
+    if (!votes) return undefined;
+
+    const vote = votes.find(vote => vote.proposalId === proposal.id);
+    if (!vote) return undefined;
+
+    return vote.selection;
+  };
 
   const proposals = await governor.getProposals();
   const proposalsState = await Promise.all(
@@ -56,7 +62,7 @@ export async function getProposals(governor: Governor | undefined): Promise<Prop
           maxVotes: proposal.state.totalElectorateWeight,
         },
         result: proposal.result,
-        vote: getVote(proposal),
+        vote: await getVote(proposal),
         hasEnded: !(proposal.status === ProposalStatus.Submitted),
       };
     }),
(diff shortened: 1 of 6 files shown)


packages/sil-voting-app/src/store/extension/actions.ts
```diff
@@ -1,11 +1,15 @@
...
 import { getBnsConnection } from "../../logic/connection";
 import { ExtensionState, SetExtensionStateActionType } from "./reducer";
 
+let identities: GetIdentitiesResponse | undefined;
+
 export async function getExtensionStatus(): Promise<ExtensionState> {
-  const identities = await sendGetIdentitiesRequest();
+  if (!identities) {
+    identities = await sendGetIdentitiesRequest();
+  }
```
webmaster128: Why is this cache needed? It feels like the real bug is that getExtensionStatus is called twice

webmaster128: This is already cached in redux

abefernan: I agree, but not sure of how to access governor without using `getExtensionStatus()`.

webmaster128: After login

```ts
    const result = await getExtensionStatus();
    dispatch(setExtensionStateAction(result.connected, result.installed, result.governor));
```

it should be available in your store:

Bildschirmfoto 2019-08-08 um 14 05 34

However, I don't really know how to access this store. Maybe @albertandrejev can help

abefernan: Ok, we are on the same page then, I'll try that approach.

willclarktech: Surely you can just access the store in the component and pass the governor to the `getProposals` function?

abefernan: I found the simplest solution to be `export const store`, but that seemingly may have some consequences that I don't fully understand. So I implemented the solution with a Redux Middleware that saves the governor to a variable. However, I would need some advice on the `if (!governor)` checks in `packages/sil-voting-app/src/store/proposals/actions.ts`.

albertandrejev: Also you can use `Thunk` middleware, it will allow you to access `state` as well as `dispatch`

albertandrejev:
I think best option to provide governor to the getProposals is pass it using argument. Feel free to review https://github.com/iov-one/ponferrada/pull/520

packages/sil-voting-app/src/store/proposals/actions.ts
```diff
@@ -1,15 +1,23 @@
 import { Proposal, ProposalStatus, VoteOption } from "@iov/bns";
+import { Governor } from "@iov/bns-governance";
 
-import { getDummyElectionRules, getDummyProposals, getDummyVote } from "./dummyData";
+import { getDummyVote } from "./dummyData";
 import { AddProposalsActionType, ProposalsState } from "./reducer";
 
+let governor: Governor | undefined;
```
webmaster128: This module level storage is a suplicate of redux's job. Here you find a proposed alternative using redux: https://github.com/iov-one/ponferrada/pull/520

abefernan: The proposed alternative of #520 always results in `governor === undefined`. I have modified it a bit, storing a `governor` variable local to the `Login` component and setting it in `isExtensionConnected()`.

webmaster128: I just looked into it again. What you say makes sense. In https://github.com/iov-one/ponferrada/pull/520/files, the `gonernor` variable is set once in creation of the Login component from the current redux store value (which is undefined at Login creation time). We have to requery the store after `isExtensionConnected` is called. Anyways, your solution is a good step in the right direction and we can always improve things.

PR #526, opened 2019-08-14
Feature #522 - Replace clipboard package

### LINKED ISSUE #522: Remove react-copy-to-clipboard dependency
react-copy-to-clipboard is not well maintained and we already use the simpler, well maintained clipboard-copy in other places of the repo.

- [x] Replace usages of `import CopyToClipboard from "react-copy-to-clipboard";` with `clipboard-copy`
- [x] Remove packages `react-copy-to-clipboard` and `@types/react-copy-to-clipboard`

albertandrejev:
**Description**
This PR will replace package to work with clipboard. It will also remove not used `reselect` package to keep code consistency. Closes #522.


packages/bierzo-wallet/package.json
```diff
@@ -21,21 +21,19 @@
...
     "react-dom": "^16.9.0",
     "react-final-form-hooks": "^2.0.1",
     "react-redux": "^7.1.0",
     "react-router": "^5.0.1",
     "react-router-dom": "^5.0.0",
     "redux": "^4.0.1",
-    "reselect": "^4.0.0",
-    "ui-logic": "^0.3.0"
+    "ui-logic": "^0.3.0",
+    "clipboard-copy": "^3.1.0"
```
webmaster128: Can you sort those alphabetically?

PR #532, opened 2019-08-15
Fix #450 - Proper error messages when sending to personalized address

### LINKED ISSUE #450: Fix error messages when sending to personalized address
I registered the username

```
[ { id: 'simon*iov',
    owner: 'tiov15auuk2w3nut5s680px7tfr6ukfyn8t7ypakvx3',
    targets: [] } ]
```

i.e. a username with no connected addresses. When I try to send to this personalized name, the current error message is misleading. What we want is:

1. If username does not exist: "Recipient's personalized address was not found"
2. If no address was registered for the chain ID: "Recipient's personalized address does not contain an address for this blockchain"

albertandrejev:
**Description**
This PR will add additional toast messages for payment view. Closes #450.

### DIFF
--- a/packages/bierzo-wallet/src/logic/account.ts
+++ b/packages/bierzo-wallet/src/logic/account.ts
@@ -8,10 +8,7 @@ import { getConnectionFor, isBnsSpec } from "./connection";
  * Returns the address associated with the name, or undefined if not registered.
  * The name must include a namespace ("*iov")
  */
-export async function lookupRecipientAddressByName(
-  username: string,
-  chainId: ChainId,
-): Promise<Address | undefined> {
+export async function lookupRecipientAddressByName(username: string, chainId: ChainId): Promise<Address> {
   if (!username.endsWith("*iov")) {
     throw new Error("Username must include namespace suffix");
   }
@@ -27,12 +24,16 @@ export async function lookupRecipientAddressByName(
     const connection = (await getConnectionFor(chain.chainSpec)) as BnsConnection;
     const usernames = await connection.getUsernames({ username });
     if (usernames.length !== 1) {
-      return undefined;
+      throw new Error("Recipient's personalized address was not found");
     }
 
     const chainAddressPair = usernames[0].targets.find(addr => addr.chainId === chainId);
 
-    return chainAddressPair ? chainAddressPair.address : undefined;
+    if (chainAddressPair) {
+      return chainAddressPair.address;
+    }
+
+    throw new Error("Recipient's personalized address does not contain an address for this blockchain");
   }
 
   throw new Error("No BNS connection found");
--- a/packages/bierzo-wallet/src/logic/account.ts
+++ b/packages/bierzo-wallet/src/logic/account.ts
@@ -4,11 +4,16 @@ import { BnsConnection } from "@iov/bns";
 import { getConfig } from "../config";
 import { getConnectionFor, isBnsSpec } from "./connection";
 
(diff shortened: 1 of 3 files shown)


packages/bierzo-wallet/src/logic/account.ts
```diff
@@ -8,10 +8,7 @@ import { getConnectionFor, isBnsSpec } from "./connection";
  * Returns the address associated with the name, or undefined if not registered.
  * The name must include a namespace ("*iov")
  */
-export async function lookupRecipientAddressByName(
-  username: string,
-  chainId: ChainId,
-): Promise<Address | undefined> {
+export async function lookupRecipientAddressByName(username: string, chainId: ChainId): Promise<Address> {
```
webmaster128: Can we use return values instead of exceptions? This allows us to differentiate between expected return values and errors.

A return type could be: `Promise<Address | "name_not_found" | "no_address_for_blockchain">`

packages/bierzo-wallet/src/routes/payment/index.tsx
```diff
@@ -54,12 +54,13 @@ const Payment = (): JSX.Element => {
...
+    let recipient: Address;
     if (isIov(formValues[ADDRESS_FIELD])) {
-      recipient = await lookupRecipientAddressByName(formValues[ADDRESS_FIELD], chainId);
-
-      if (!recipient) {
-        toast.show("IOV username was not found", ToastVariant.ERROR);
+      try {
+        recipient = await lookupRecipientAddressByName(formValues[ADDRESS_FIELD], chainId);
+      } catch (err) {
+        toast.show(err.message, ToastVariant.ERROR);
```
webmaster128: Here we convert all errors into toasts, the expected ones you created and all the other stuff that can go wrong during the network request.

packages/bierzo-wallet/src/routes/payment/index.tsx
```diff
@@ -54,13 +54,18 @@ const Payment = (): JSX.Element => {
 
     const chainId = token.chainId;
 
-    let recipient: Address;
+    let recipient: LookupResultType;
```
webmaster128: This type can remain `Address`

packages/bierzo-wallet/src/routes/payment/index.tsx
```diff
@@ -54,13 +54,18 @@ const Payment = (): JSX.Element => {
...
     if (isIov(formValues[ADDRESS_FIELD])) {
-      try {
-        recipient = await lookupRecipientAddressByName(formValues[ADDRESS_FIELD], chainId);
-      } catch (err) {
-        toast.show(err.message, ToastVariant.ERROR);
-        console.error(err);
+      recipient = await lookupRecipientAddressByName(formValues[ADDRESS_FIELD], chainId);
```
webmaster128: when you store the `lookupResult` in a separate variable

packages/bierzo-wallet/src/routes/payment/index.tsx
```diff
@@ -54,13 +54,18 @@ const Payment = (): JSX.Element => {
...
+      if (recipient === "name_not_found") {
+        toast.show("Recipient's personalized address was not found", ToastVariant.ERROR);
+        return;
+      } else if (recipient === "no_address_for_blockchain") {
+        toast.show(
+          "Recipient's personalized address does not contain an address for this blockchain",
+          ToastVariant.ERROR,
+        );
         return;
       }
```
webmaster128: and assign `recipient = lookupResult;` only in the case where you know it is an address

PR #533, opened 2019-08-15
Feature #405 - Personalized address E2E test

### LINKED ISSUE #405: Add E2E tests to check BNS username on /balance route
Implement E2E tests to create and check BNS username on /balance route.
Use following logic:

1. We signup in the app using the extension;
2. We buy a human-readable address;
3. We travel to the account balance view;
4. We see the human-readable address just bought.

Based on [comment](https://github.com/iov-one/ponferrada/pull/403#discussion_r305751876).

albertandrejev:
**Description**
This PR will implement E2E test to check buy process of a personalized address.

Closes #405.

### DIFF
--- a/packages/ui-logic/src/random.ts
+++ b/packages/ui-logic/src/random.ts
@@ -2,4 +2,6 @@ import PseudoRandom, { MersenneTwister19937, string } from "random-js";
 
 const prng: PseudoRandom.Engine = MersenneTwister19937.autoSeed();
 const pool = "abcdefghijklmnopqrstuvwxyz0123456789";
-export const randomString = (len: number): string => string(pool)(prng, len);
+export function randomString(len: number): string {
+  return string(pool)(prng, len);
+}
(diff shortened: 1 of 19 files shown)


packages/ui-logic/src/random.ts
```diff
@@ -2,4 +2,6 @@ import PseudoRandom, { MersenneTwister19937, string } from "random-js";
 
 const prng: PseudoRandom.Engine = MersenneTwister19937.autoSeed();
 const pool = "abcdefghijklmnopqrstuvwxyz0123456789";
-export const randomString = (len: number): string => string(pool)(prng, len);
+export function randomString(len: number): string {
+  return string(pool)(prng, len);
+}
```
webmaster128: 👍 

Can you create a simple unittest for everything that is in `ui-logic`? It does not need to covery every edge case but at least run `randomString` to make sure the dependencies are in place

PR #535, opened 2019-08-18
Feature #516 - Personalized address term

### LINKED ISSUE #516: Use term "personalized address" in the user interface
Right now we are working with a wild mix of candidates, including:
1. Name
2. Username
3. IOV address
4. Human readable address

The new marketing name is *personalized address* and this should be used everywhere in the user interface (and if possible with low priority in code as well).

This ticket is about going through the UI and searcgh the project for places where it is not yet called personalized address.

-----------

Some places are (not exhaustive):

- `export const REGISTER_USERNAME_ROUTE = "/register-username";`

albertandrejev:
**Description**
This PR will fix `personalized address` term in UI labels.
Closes #516.

### DIFF
--- a/packages/bierzo-wallet/src/routes/balance/index.e2e.spec.ts
+++ b/packages/bierzo-wallet/src/routes/balance/index.e2e.spec.ts
@@ -57,13 +57,13 @@ withChainsDescribe("E2E > Balance route", () => {
     expect(balances).toEqual(["10 BASH", "10 CASH", "10 ETH", "5 LSK"]);
   }, 45000);
 
-  it("should contain message to get username", async () => {
+  it("should contain message to get address", async () => {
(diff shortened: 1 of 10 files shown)


packages/bierzo-wallet/src/routes/balance/index.e2e.spec.ts
```diff
@@ -63,9 +63,9 @@ withChainsDescribe("E2E > Balance route", () => {
     expect(username).toBe("Get your human readable");
   }, 45000);
 
-  it("should create username", async () => {
+  it("should create personalized username", async () => {
```
webmaster128: "personalized username" -> "personalized address" or "username". Both terms are fine for me in code, but "personalized username" does not make sense

Issue #480, opened 2019-08-06
Make it possible to change configuration without recompiling


webmaster128:
Very similar to https://github.com/iov-one/ponferrada/pull/474 but with some additional adjustments which we did afterwards.

webmaster128:
Done in https://github.com/iov-one/ponferrada/pull/536

PR #536, opened 2019-08-19
Feature #480 - Change configuration without recompiling

albertandrejev:
**Description**
This PR will make it possible to change app configuration runtime. Closes #490.

### DIFF
--- a/packages/sil-voting-app/package.json
+++ b/packages/sil-voting-app/package.json
@@ -23,7 +23,7 @@
     "react-scripts": "^3.0.1"
   },
   "scripts": {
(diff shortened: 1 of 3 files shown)


packages/sil-voting-app/package.json
```diff
@@ -23,20 +23,24 @@
     "react-scripts": "^3.0.1"
   },
   "scripts": {
+    "use-extension-config-development": "cd ../sanes-chrome-extension && yarn use-config-development && yarn override-config-development && cd ../bierzo-wallet",
```
webmaster128: copy/paste error: `bierzo-wallet`

Issue #544, opened 2019-08-20
Support governance transactions in browser extension


willclarktech:
- [x] `Vote`
- [x] `CreateProposal`

Signing and displaying. `CreateProposal` should just display its content as a JSON object for MVP.

PR #558, opened 2019-08-21
Feature #502 - Remove error handler and check previous vote

### LINKED ISSUE #502: Implement Voting Logic
Make use of `governor.buildVoteTx()` and  `governor.signAndPost()` to implement the voting mechanism.

abefernan:
**Description**

Makes some fixes regarding the Voting Panel:
- Removes the error handler from the Voting Panel now that #544 is implemented
- Adds a `previousVote` state so that if a user clicks on the same current vote it isn't cast
- Changes the ocurrencies of `creationDate`s (which were specified in the design wireframes) to `startDates` (which better represent the actual data).

Closes #502.


packages/sil-voting-app/src/routes/dashboard/components/Proposal/VotingPanel/Buttons.tsx
```diff
@@ -15,6 +15,7 @@ interface Props {
 
 const Buttons = ({ id, vote }: Props): JSX.Element => {
   const [currentVote, setCurrentVote] = useState(vote);
+  const [previousVote, setPreviousVote] = useState(currentVote);
```
webmaster128: This could also be `= useState(currentVote);` like above, right? If you do that, it makes the source of the initial state more direct and you can swap the order of those lines easily

abefernan: Do you mean `const [previousVote, setPreviousVote] = useState(vote);`? If not I'm afraid I don't understand.

webmaster128: I do

packages/sil-voting-app/src/store/proposals/actions.ts
```diff
@@ -43,12 +43,14 @@ export async function getProposals(governor: Governor): Promise<ProposalsState>
   const proposals = await governor.getProposals();
   const proposalsState = await Promise.all(
     proposals.map(async proposal => {
+      const creationDate = new Date(proposal.votingStartTime * 1000);
```
webmaster128: `creationDate` -> `startDate`

packages/sil-voting-app/src/store/proposals/actions.ts
```diff
@@ -61,6 +63,7 @@ export async function getProposals(governor: Governor): Promise<ProposalsState>
         },
         result: proposal.result,
         vote: await getVote(proposal),
+        hasStarted: creationDate < new Date(),
```
webmaster128: There is no operator overloading in JavaScript. So I think `<` between `Date` objects does not what you think it does. You can use `.getTime()` on both sides and compare milliseconds as integers

packages/sil-voting-app/src/store/proposals/dummyData.ts
```diff
@@ -217,12 +217,14 @@ export const getDummyProposalsState = (): ProposalsState => {
   const getVote = (proposal: Proposal): VoteOption | undefined => getDummyVote(proposal);
 
   const proposalsState = proposals.map(proposal => {
+    const creationDate = new Date(proposal.votingStartTime * 1000);
```
webmaster128: `startDate`

packages/sil-voting-app/src/store/proposals/dummyData.ts
```diff
@@ -235,6 +237,7 @@ export const getDummyProposalsState = (): ProposalsState => {
       },
       result: proposal.result,
       vote: getVote(proposal),
+      hasStarted: creationDate < new Date(),
```
webmaster128: as above

packages/sil-voting-app/src/store/proposals/actions.ts
willclarktech: A couple of StackOverflow answers I've seen suggest this actually works fine. (You can use `<` and `>`, but not `==`.)

https://stackoverflow.com/a/13285569

webmaster128: Alright, 👍 then

webmaster128:
Not critical but something to keep in mind: here we compare a local time with an on-chain time. This can potentially cause troube when the machine's time is out of sync and thus the user cannot vote even if the voting time did not end on chain. It would be possible to get the time from the chain (via block headers). But for MVP this is unnecessarily complicated.

packages/sil-governance/src/routes/dashboard/components/Proposal/Period.tsx
```diff
@@ -3,22 +3,22 @@ import React from "react";
...
+  const startLocaleDate = startDate.toLocaleString();
 
   return (
     <Block display="flex" alignItems="center">
       <Typography variant="body2" weight="semibold">
         {expiryLabel} {expiryLocaleDate}
       </Typography>
       <Block marginLeft={1}>
-        <Typography variant="body2">Created on {creationLocaleDate}</Typography>
+        <Typography variant="body2">Created on {startLocaleDate}</Typography>
```
webmaster128: Do I understand correctly this is the date when the voting on the proposal starts? Then the proposal can be created a long time before the voting actually starts. So I suggest the texts "Starts on" / "Started on" like we have with `expiryLabel`.

Sorry if this is out of scope for this PR. You can also create a different ticket if you want to merge as is

abefernan: Yes, the proper creation date of the proposal doesn't belong in the UI, I believe. The wireframes called the "start date" "creation date". I'll update the label no problem.

webmaster128: Creation date of the proposal is not important for now. But the governor needs to know when they can vote.

Issue #509, opened 2019-08-07
Create proposals on blockchain via governor instance


abefernan:
Make use of `governor.buildCreateProposalTx` to create a proposal and then sign it with `signAndPost()`.

Part of #400

- [ ] AddCommitteeMember
- [ ] RemoveCommitteeMember
- [ ] AmendElectionRuleThreshold
- [ ] AmendElectionRuleQuorum
- [ ] AddValidator
- [ ] RemoveValidator
- [ ] ReleaseGuaranteeFunds
- [ ] DistributeFunds
- [x] AmendProtocol

webmaster128:
What is missing here? I see create proposal functionality (AmendProtocol) in code and will wrote it is working. Can you close or update the ticket, @abefernan?

abefernan:
> 
> 
> What is missing here? I see create proposal functionality (AmendProtocol) in code and will wrote it is working. Can you close or update the ticket, @abefernan?

Because I thought maybe this should include support for the rest of the proposal types.

webmaster128:
@abefernan okay, I created a checklist for visibility of that. Can you please keep it up to date? Thanks!

webmaster128:
Let's break this down into smaller tickets, since not all proposal types are required for inital user testing.

1. AddCommitteeMember/RemoveCommitteeMember: https://github.com/iov-one/ponferrada/issues/614
2. AmendElectionRuleThreshold/AmendElectionRuleQuorum: https://github.com/iov-one/ponferrada/issues/615
3. AddValidator/RemoveValidator:  https://github.com/iov-one/ponferrada/issues/616
4. ReleaseGuaranteeFunds: https://github.com/iov-one/ponferrada/issues/617
5. DistributeFunds: https://github.com/iov-one/ponferrada/issues/618

PR #567, opened 2019-08-21
Feature #509 - Proposal creation with governor

### LINKED ISSUE #508: Layout for creating Amend Protocol Proposal
Prepare the layout for Amend Protocol Proposal (Create Text Resolution) so that proposal creation with the Governor can be tested with the UI.

abefernan:
**Description**
This PR enables the creation of Amend Protocol proposals in the Create Proposal view. It currently uses a hardcoded `electionRuleId` to that end.

Closes #508 
Part of #509


packages/sil-governance/src/routes/create-proposal/components/FormOptions/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+    <React.Fragment>
+      {proposalType === ProposalType.AmendProtocol && <AmendProtocol form={form} />}
+      {proposalType === ProposalType.AddCommitteeMember && <AddCommitteeMember form={form} />}
+      {proposalType === ProposalType.RemoveCommitteeMember && <RemoveCommitteeMember form={form} />}
+      {proposalType === ProposalType.AmendElectionRuleThreshold && <AmendCommitteeThreshold form={form} />}
+      {proposalType === ProposalType.AmendElectionRuleQuorum && <AmendCommitteeQuorum form={form} />}
+      {proposalType === ProposalType.AddValidator && <AddValidator form={form} />}
+      {proposalType === ProposalType.RemoveValidator && <RemoveValidator form={form} />}
+      {proposalType === ProposalType.ReleaseGuaranteeFunds && <ReleaseGuaranteeFunds form={form} />}
+      {proposalType === ProposalType.DistributeFunds && <DistributeFunds form={form} />}
```
willclarktech: Wouldn't a switch statement be clearer? This way I have to look down the whole list to work out that there'll only be one child component.

packages/sil-governance/src/routes/create-proposal/components/ProposalForm.tsx
```diff
@@ -1,53 +1,75 @@
...
-  amendCommitteeThreshold: "Amend Committee Threshold",
-  amendCommitteeQuorum: "Amend Committee Quorum",
-  releaseGuaranteeFunds: "Release Guarantee Funds",
-  distributeFunds: "Distribute Funds",
-  addValidator: "Add Validator",
-  removeValidator: "Remove Validator",
-  amendProtocol: "Amend Blockchain Software",
-};
+const buildProposalOptions = (type: ProposalType, values: FormValues): ProposalOptions => {
+  console.log(JSON.stringify(values));
```
willclarktech: `console.log`

packages/sil-governance/src/routes/create-proposal/components/ProposalForm.tsx
```diff
@@ -1,53 +1,75 @@
...
+
+  const [year, month, day] = values["Date"].split("-").map(el => parseInt(el));
```
willclarktech: Best practice is to supply the radix parameter when using `parseInt`: does the linter not raise this? See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Description

packages/sil-governance/src/routes/create-proposal/components/ProposalForm.tsx
```diff
@@ -1,53 +1,75 @@
...
+    description,
+    startTime,
+    electionRuleId,
+  };
+
+  const text = values["Text"];
+
+  switch (type) {
+    default:
+      return { ...commonOptions, type: ProposalType.AmendProtocol, text };
```
willclarktech: The default should probably be to throw. This `return` statement can be in a `case ProposalType.AmendProtocol`.

packages/sil-governance/src/routes/create-proposal/components/ProposalForm.tsx
```diff
@@ -1,53 +1,75 @@
...
+  }
+};
 
 const ProposalForm = (): JSX.Element => {
-  const [proposalType, setProposalType] = useState("");
+  const [proposalType, setProposalType] = useState(ProposalType.AmendProtocol);
+  const governor = ReactRedux.useSelector((state: RootState) => state.extension.governor);
+
+  const changeProposal = (proposalType: ProposalType): void => setProposalType(proposalType);
```
willclarktech: Why not just use `setProposalType` directly?

packages/sil-governance/src/routes/create-proposal/components/ProposalTypeSelect.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+const formOptionsItems = Object.keys(formOptions).map(formOption => {
+  return {
+    name: formOption,
+  };
+});
+
+interface Props {
+  form: FormApi;
+  changeProposal: (proposalType: ProposalType) => void;
```
willclarktech: Can we call this `changeProposalType` or something else that makes it clear we're switching to a different proposal type, not changing the content of a specific proposal?

packages/sil-governance/src/routes/create-proposal/components/FormOptions/index.tsx
apanizo: Of course if you want you can use a switch, but in this case, I suggest to leave the code as it is because it is a very standard way to configure conditionally views on React. And if you use a switch it is going to look strange. 

But guys, if you definitely want to do a switch, feel free do it. Just my 2 cents.

willclarktech: In general I'm totally on board with conditional views. But here I had to do some work to figure out that it's displaying *one* component dependent on the proposal type, whereas a `switch` would make this 100% clear at a glance.

willclarktech: An alternative here would be a map and no conditional at all:
```tsx
const proposalMap = {
  [ProposalType.AddValidator]: AddValidator,
  [ProposalType.RemoveValidator]: RemoveValidator,
  // ...
};
// ...
const FormComponent = proposalMap[proposalType];
// ...
<FormComponent form={form} />
```

abefernan: I agree with both points of view. I like the map solution because:
1. It has just one line of code per case.
2. The return function has JSX code, which adds familiarity. (The current switch statement has a not needed `React.Fragment` that if omitted leaves `return getCurrentProposalOption()` which seems kind weird)

willclarktech: The map option also makes it transparent that these components all take the same props.

webmaster128:
@willclarktech @abefernan is this ready for merge?

PR #572, opened 2019-08-22
Feature #459 - Close wallet session

### LINKED ISSUE #459: Add "Close session" button
It should be possible to close the wallet session and login with a different account. Closing a session should be equivalent to reload the tab, i.e.

1. clear all redux state
2. redirect to `/`
3. Disconnect all blockchain connections

albertandrejev:
**Description**
This PR will add ability to close current wallet session by click on Logout button. Closes #459.

### DIFF
--- a/packages/bierzo-wallet/src/components/Header/components/HiMenu/index.tsx
+++ b/packages/bierzo-wallet/src/components/Header/components/HiMenu/index.tsx
@@ -5,9 +5,12 @@ import ListItemIcon from "@material-ui/core/ListItemIcon";
 import ListItemText from "@material-ui/core/ListItemText";
 import { Block, Hairline, Image, ListMenu, Typography } from "medulas-react-components";
 import * as React from "react";
+import * as ReactRedux from "react-redux";
 
+import { disconnect } from "../../../../logic/connection";
 import { history } from "../../../../routes";
 import { POLICY_ROUTE, TERMS_ROUTE } from "../../../../routes/paths";
+import { resetAppAction } from "../../../../store";
 import { getBorderColor } from "../../../../theme/css";
 import chevronDown from "../../assets/chevronDown.svg";
 import chevronUp from "../../assets/chevronUp.svg";
@@ -70,13 +73,15 @@ const onPolicy = (): void => {
   history.push(POLICY_ROUTE);
 };
 
-const onLogout = (): void => {
-  history.push(POLICY_ROUTE);
-};
-
 const HiMenu = (props: {}): JSX.Element => {
   const classes = useStyles();
+  const dispatch = ReactRedux.useDispatch();
   const { ...rest } = props;
+
+  const onLogout = (): void => {
+    dispatch(resetAppAction());
+    disconnect();
+  };
 
   const starter = (open: boolean): JSX.Element => (
     <Block className={classes.root}>
--- a/packages/bierzo-wallet/src/components/Header/components/HiMenu/index.tsx
+++ b/packages/bierzo-wallet/src/components/Header/components/HiMenu/index.tsx
@@ -11,6 +11,7 @@ import { POLICY_ROUTE, TERMS_ROUTE } from "../../../../routes/paths";
 import { getBorderColor } from "../../../../theme/css";
 import chevronDown from "../../assets/chevronDown.svg";
 import chevronUp from "../../assets/chevronUp.svg";
+import logout from "../../assets/logout.svg";
 import privacy from "../../assets/privacyPolicy.svg";
 import terms from "../../assets/terms.svg";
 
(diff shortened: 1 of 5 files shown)


packages/bierzo-wallet/src/store/reducers.ts
```diff
@@ -15,14 +21,24 @@ export interface RootReducer {
...
+  balances: balancesReducer,
+  usernames: usernamesReducer,
+});
+
+const createRootReducer = (): Reducer<RootReducer> => (
+  state: RootReducer | undefined,
+  action: ActionType<typeof actions>,
+) => {
+  if (action.type === "@@app/RESET") {
+    state = undefined;
```
webmaster128: Reassigning a function argument is not good practice.

But `return allReducers(undefined, action);` would do the job, right?

albertandrejev: yes it should.

albertandrejev:
Odd I added new commits but they did not appeared in this PR. But pushed branch reflects those changes.

packages/bierzo-wallet/src/components/Header/components/HiMenu/index.tsx
```diff
@@ -71,8 +75,14 @@ const onPolicy = (): void => {
 
 const HiMenu = (props: {}): JSX.Element => {
   const classes = useStyles();
+  const dispatch = ReactRedux.useDispatch();
   const { ...rest } = props;
 
+  const onLogout = (): void => {
+    dispatch(resetAppAction());
```
webmaster128: Does this also change the location?

packages/bierzo-wallet/src/components/Header/components/HiMenu/index.tsx
```diff
@@ -71,8 +75,14 @@ const onPolicy = (): void => {
...
+    disconnect();
```
webmaster128: Can we call this first before we reset the app? Just in case it throws errors.

packages/bierzo-wallet/src/components/Header/components/HiMenu/index.tsx
albertandrejev: Yes, It will change location also https://github.com/iov-one/ponferrada/blob/a457876b8d008007a0acdcee8cfb9b235d6dda37/packages/bierzo-wallet/src/components/RequireLogin/index.tsx#L19

packages/bierzo-wallet/src/store/reducers.ts
```diff
@@ -15,14 +21,24 @@ export interface RootReducer {
...
+  balances: balancesReducer,
+  usernames: usernamesReducer,
+});
+
+const createRootReducer = (): Reducer<RootReducer> => (
+  state: RootReducer | undefined,
+  action: ActionType<typeof actions>,
+) => {
+  if (action.type === "@@app/RESET") {
+    return allReducers(undefined, action);
```
apanizo: What does it do? Force to all reducers to load their initial state?

albertandrejev: Yes

packages/bierzo-wallet/src/components/Header/components/HiMenu/index.tsx
webmaster128: Thanks for checking!

packages/bierzo-wallet/src/store/reducers.ts
apanizo: @webmaster128 @albertandrejev if the code works, is fine, I was going to add, this is the first time I see implement the logout in this way (maybe for some reason in the latest redux version is recommended), but I think is much more natural and clean to listen in all reducers for the reset action and returns directly the partial state you want there. 

Potential problems I see with this approach:
1. If you add more reducers which you do not want to be modified when the user disconnects you will have problems.
2. If you have strange errors it is not trivial to arrive here and remember the reset stuff.

Anyway, if you guys think this is a safe bet, feel free to leave as it is.

albertandrejev: But with this approach you do not need to implement reset action in every reducer (in most cases this is simply copy+paste solution and will violate DRY principle). Also you can forget to implement this action and this can cause security leaks. 
If you want you can implement specific behaviour per reducer for reset action, but you do not need to care about it in case if you simply want to return default state from the specific reducer.
You can enhance my solution if you want to leave specific part of the state simply by copying this part into returning state.

For example:
```ts
 if (action.type === "@@app/RESET") {
    const { tokens } = state!;
    const newState = allReducers(undefined, action);
    newState.tokens = { ...tokens };
    return newState;
}
```

You can use this approach and make it more flexible.

albertandrejev: From Dan Abramov's answer https://stackoverflow.com/questions/35622588/how-to-reset-the-state-of-a-redux-store/35641992#35641992

PR #573, opened 2019-08-22
Feature - Request flow improved

### LINKED ISSUE #523: Improve Chrome extension requests flow
When the website makes a request, the extension shows a little badge (so far so good). Then when you open the extension there’s no guidance about where to go next, you just have to know/remember/guess you have to click on the menu and navigate to “requests”.

Possible improvements:
- take the user to that page directly when the extension is opened and requests are pending
- highlight the path to that page a bit more clearly e.g with more badges on the menu icon/requests item
- have a popup that puts the request queue above the rest of the UI
- add a requests view to the homepage

apanizo:
**Description**
In this PR we have modified the request flow. Now we see the request view when opening the extension if they are requests in the queue.

See it in action [here](https://drive.google.com/open?id=1ZvlGEToiP89V1Svo_z4uw8VNnIxfjf89)
See it in action [after comments amended](https://drive.google.com/open?id=1g62z_F9Rp5w9oT-5EGStQFpj8c8SSguW)

It closes #523

### DIFF
--- a/packages/bierzo-wallet/src/utils/test/persona.ts
+++ b/packages/bierzo-wallet/src/utils/test/persona.ts
@@ -48,13 +48,19 @@ export async function openEnqueuedRequest(extensionPage: Page): Promise<void> {
   await extensionPage.click("ul > li > div");
 }
 
-export async function acceptEnqueuedRequest(browser: Browser): Promise<void> {
+async function clickOnFirstRequest(browser: Browser): Promise<Page> {
   const extensionPage = await createExtensionPage(browser);
   await extensionPage.bringToFront();
 
   // click on first request
   await extensionPage.click("ul > li > div");
   await sleep(500);
+
+  return extensionPage;
+}
+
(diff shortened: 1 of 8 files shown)


packages/bierzo-wallet/src/utils/test/persona.ts
```diff
@@ -47,8 +48,12 @@ export async function openEnqueuedRequest(extensionPage: Page): Promise<void> {
...
 }
 
-export async function acceptEnqueuedRequest(extensionPage: Page): Promise<void> {
-  await openEnqueuedRequest(extensionPage);
+export async function acceptEnqueuedRequest(browser: Browser): Promise<void> {
+  const extensionPage = await createExtensionPage(browser);
+  await extensionPage.bringToFront();
+
+  // click on first request
+  await extensionPage.click("ul > li > div");
```
albertandrejev: Maybe it is better to move common code from `acceptEnqueuedRequest` and `rejectEnqueuedRequest` to separate method?

PR #580, opened 2019-08-23
Feature #574 - Remove unused Sentry dependency

### LINKED ISSUE #574: Remove sentry
As long as we have no GDPR-complient UX, we should remove all traces of sentry from the codebase.

This incluses the package `@sentry/browser` as well as all usages of it. The `ErrorBoundary` component is good to keep. But whatever is cought should be logged to `console.error` instead of being sent to an external service.

albertandrejev:
**Description**
This PR will remove Sentry package as well as `ErrorBoundary` component because it will become unused. It will close #574.

**Developer note**
There is no need to left ErrorBoundary at all. Because this is default React behaviour to show error on the console.


packages/sanes-chrome-extension/src/routes/paths.unit.spec.ts
```diff
@@ -1,35 +0,0 @@
-import { ACCOUNT_STATUS_ROUTE, initialUrl, LOGIN_ROUTE, REQUEST_ROUTE, WELCOME_ROUTE } from "./paths";
-
```
webmaster128: Why this unrelated deletion of `packages/sanes-chrome-extension/src/routes/paths.unit.spec.ts`?

albertandrejev: oops!

PR #581, opened 2019-08-23
Feature #576 - Better personalized address validation

### LINKED ISSUE #576: Improve name validation locally
We want to avoid that uses need to sign username registrations that are rejected by the blockchain. Thus the validation rules should be applied locally along with helpful error messages.

Rules:

- A username must be `<name>*iov` (i.e. a namespace labels is always `iov`. We do not allow any other label)
- A `name` must be minimum 3 and maximum 64 characters,
- A `name` must be build using only the following ASCII characters `abcdefghijklmnopqrstuvwxyz0123456789-_.` (last 3 being dot, underscore and dash),

This required improve the already existing validator:

```ts
const validate = async (values: object): Promise<object> => {
  const formValues = values as FormValues;
  const errors: ValidationError = {};

  const username = formValues[REGISTER_USERNAME_FIELD];
  if (!username) {
    errors[REGISTER_USERNAME_FIELD] = "Required";
    return errors;
  }

  if (!username.endsWith("*iov")) {
    errors[REGISTER_USERNAME_FIELD] = "Personalized address must include namespace suffix";
    return errors;
  }

  const connection = await getConnectionForBns();
  const usernames = await connection.getUsernames({ username });
  if (usernames.length > 0) {
    errors[REGISTER_USERNAME_FIELD] = "Personalized address already exists";
  }
  return errors;
};
```

albertandrejev:
**Description**
This PR will implement better personalized address validation for registration component, Closes #576.

### DIFF
--- a/packages/bierzo-wallet/src/logic/account.ts
+++ b/packages/bierzo-wallet/src/logic/account.ts
@@ -46,16 +46,21 @@ export async function lookupRecipientAddressByName(
   throw new Error("No BNS connection found");
 }
 
-export function isValidIov(username: string): "valid" | "not_iov" | "too_short" | "too_long" | "wrong_chars" {
+export function isValidIov(
+  username: string,
+): "valid" | "not_iov" | "wrong_number_of_asterisks" | "too_short" | "too_long" | "wrong_chars" {
   if (!isIov(username)) return "not_iov";
 
-  const usernameLength = username.length - "*iov".length;
+  const parts = username.split("*");
+  if (parts.length !== 2) return "wrong_number_of_asterisks";
+  //TODO: add namespace variable as soon as multiple namespaces has been supported
+  const [name] = parts;
 
   //Username length must be at least 3 chars long
-  if (usernameLength < 3) return "too_short";
+  if (name.length < 3) return "too_short";
 
   //Username length must maximum 64 chars long
-  if (usernameLength > 64) return "too_long";
+  if (name.length > 64) return "too_long";
 
   //Must contain only allowed chars
   if (/^[a-z0-9_\-.]{3,64}\*iov$/.test(username)) return "valid";
--- a/packages/bierzo-wallet/src/logic/account.ts
+++ b/packages/bierzo-wallet/src/logic/account.ts
@@ -4,6 +4,10 @@ import { BnsConnection } from "@iov/bns";
 import { getConfig } from "../config";
 import { getConnectionFor, isBnsSpec } from "./connection";
 
+export function isIov(username: string): boolean {
+  return username.endsWith("*iov");
+}
+
 /**
  * Returns the address associated with the name, or undefined if not registered.
  * The name must include a namespace ("*iov")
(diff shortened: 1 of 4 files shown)


packages/bierzo-wallet/src/routes/registerName/index.tsx
```diff
@@ -31,8 +32,25 @@ const validate = async (values: object): Promise<object> => {
...
+      errors[REGISTER_USERNAME_FIELD] = "Personalized address should be maximum 64 characters";
+      break;
+    case "wrong_chars":
+      errors[REGISTER_USERNAME_FIELD] =
+        "Personalized address should contain 'abcdefghijklmnopqrstuvwxyz0123456789-_.' characters only";
+      break;
+  }
+
+  if (checkResult !== "valid") {
     return errors;
```
webmaster128: The type system assumes, an `errors` object is always returned (see `Promise<object>`). So I think the condition should be removed. In case of success, `errors` is `{}`, like before

packages/bierzo-wallet/src/logic/account.ts
```diff
@@ -42,6 +46,19 @@ export async function lookupRecipientAddressByName(
   throw new Error("No BNS connection found");
 }
 
-export function isIov(username: string): boolean {
-  return username.endsWith("*iov");
+export function isValidIov(username: string): "valid" | "not_iov" | "too_short" | "too_long" | "wrong_chars" {
+  if (!isIov(username)) return "not_iov";
+
+  const usernameLength = username.length - "*iov".length;
```
webmaster128: To make the logic easier, can you split by `*` first?

```ts
const parts = username.split("*");
if (parts.lenth !== 2) return "wrong_number_of_asterisks";
const [name, namespace] = parts;
```

This is also more future-proof since we will add different namespaces soon after launch.

packages/bierzo-wallet/src/routes/registerName/index.stories.tsx
```diff
@@ -34,9 +36,33 @@ async function onSubmit(_: object): Promise<void> {
...
+    case "too_short":
+      errors[REGISTER_USERNAME_FIELD] = "Personalized address should be at least 3 characters";
+      break;
+    case "too_long":
+      errors[REGISTER_USERNAME_FIELD] = "Personalized address should be maximum 64 characters";
+      break;
+    case "wrong_chars":
+      errors[REGISTER_USERNAME_FIELD] =
+        "Personalized address should contain 'abcdefghijklmnopqrstuvwxyz0123456789-_.' characters only";
+      break;
```
webmaster128: Can you add an empty `case "valid"` where nothing happens and throw an error in `default:`? This ensures an error is not ignored in case of a missing error case.

packages/bierzo-wallet/src/routes/registerName/index.tsx
albertandrejev: I used this condition to avoid running validation against blockchain in next lines, this will help to save resources a little instead of validating 100% incorrect personalized address in blockchain.

webmaster128: Ah, yes, perfect. I did not see the rest of the code in this function. Makes sense, my faultl

PR #599, opened 2019-08-26
Feature #278 - Extension reducer test

### LINKED ISSUE #278: Extension state methods unit tests
Create unit tests for helper methods used to check extension availability and state.

albertandrejev:
**Description**
This PR will implement extension reducer test. Closes #278.


packages/bierzo-wallet/src/store/extension/index.unit.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+    };
+
+    const extensionState = store.getState().extension;
+    expect(extensionState.connected).toBeTruthy();
+    expect(extensionState.installed).toBeTruthy();
+    expect(extensionState.identities).toStrictEqual(extensionIdentities);
+  });
+});
+
+describe("groupIdentitiesByChain", () => {
```
webmaster128: Can we keep this describe block in `actions.unit.spec.ts`, which are the tests for `actions.ts`?

The `describe("Extension reducer", () => {` can remain here of that is more for the folder than `actions.ts`

packages/bierzo-wallet/src/store/extension/index.unit.spec.ts
```diff
@@ -44,10 +43,14 @@ withChainsDescribe("Extension reducer", () => {
...
 
+    const extensionIdentities: { [chain: string]: Identity } = {
+      "ethereum-eip155-5777": identitiesResponse[0],
+    };
+
     const extensionState = store.getState().extension;
     expect(extensionState.connected).toBeTruthy();
     expect(extensionState.installed).toBeTruthy();
-    expect(extensionState.identities).toBe(identitiesResponse);
+    expect(extensionState.identities).toStrictEqual(extensionIdentities);
```
webmaster128: Interesting new matcher `toStrictEqual`. I think a simple `toEqual` would do the job as well

PR #601, opened 2019-08-27
Feature - Wallet wording refactoring

### LINKED ISSUE #537: Wording inconsistency: Account has two different meaning right now
When we are in the welcome page, we have the button: create account.  It has the meaning to create a new wallet (with a new seed). 

So we suggest to use the word Wallet instead of Account on this page.

When we are in the home screen, we have the text: Account status (and create a new account), We should keep the word Account on this page.

apanizo:
**Description**
This PR removes the term account in the welcome route, and therefore in the "Create New Wallet" and the "Recover Wallet" process.

It closes #537

Screenshot 2019-08-27 11 33 34

Screenshot 2019-08-27 11 47 07
Screenshot 2019-08-27 11 49 21

Screenshot 2019-08-27 11 33 29


webmaster128:
Looks like `submitExtensionSignupForm` in the wallet needs to be adjusted as well in order to run the e2e tests that need the extension (see CI fail).

PR #609, opened 2019-08-27
Feature #510 - Highlight user address

### LINKED ISSUE #510: Highlight current user's addresses in transactions
Create some kind of _mine_ marker for addresses that are owned by the current user (especially in the transactions list, both sender and recipient).

This should be some reuable component that can be used whenever we display addresses.

Clicking this _mine_ marker can navigate to `/addresses`.

## Suggested implementation

**Recipient**:
tiov1mh9p0y0ne8n03y5x8tjrjlxz4ku3vqluq6g26s

**Recipient**:
tiov1hx6vjcmsgj7wgu64ajdglrhpusl3fmercl8mxe [(mine)](/addresses)

**Sender**:
tiov1mh9p0y0ne8n03y5x8tjrjlxz4ku3vqluq6g26s

**Sender**:
tiov1hx6vjcmsgj7wgu64ajdglrhpusl3fmercl8mxe [(mine)](/addresses)

You sent tiov1mh9p0y0n...

You sent tiov1hx6v...8mxe [⌂](/addresses)

cc @willclarktech

albertandrejev:
**Description**
This PR will implement highlight of user address in transaction screen.
Closes #510

**After**
mine-address png

### DIFF
--- a/packages/bierzo-wallet/src/components/BlockchainAddress/index.tsx
+++ b/packages/bierzo-wallet/src/components/BlockchainAddress/index.tsx
@@ -1,0 +1,40 @@
+import { Address } from "@iov/bcp";
+import { makeStyles } from "@material-ui/core";
+import { Badge, Typography } from "medulas-react-components";
+import React from "react";
+
+const useStyles = makeStyles({
+  sectionName: {
+    overflowWrap: "break-word",
+  },
+});
+
+interface Props {
+  readonly addresses: Address[];
+  readonly address: Address;
+}
+
+const BlockchainAddress = ({ addresses, address }: Props): JSX.Element => {
+  const classes = useStyles();
+
+  const addressNode = (
+    <Typography variant="subtitle2" weight="regular" color="textSecondary" className={classes.sectionName}>
+      {address}
+    </Typography>
+  );
+
+  console.log(addresses);
+  console.log(address);
+
+  if (addresses.indexOf(address) > -1) {
+    return (
+      <Badge variant="text" badgeContent="mine">
+        {addressNode}
+      </Badge>
+    );
+  } else {
+    return addressNode;
+  }
+};
+
+export default BlockchainAddress;
--- a/packages/bierzo-wallet/src/components/BlockchainAddress/index.tsx
+++ b/packages/bierzo-wallet/src/components/BlockchainAddress/index.tsx
@@ -10,11 +10,11 @@ const useStyles = makeStyles({
 });
 
 interface Props {
-  readonly addresses: Address[];
+  readonly userAddresses: Address[];
   readonly address: Address;
 }
 
-const BlockchainAddress = ({ addresses, address }: Props): JSX.Element => {
+const BlockchainAddress = ({ userAddresses, address }: Props): JSX.Element => {
   const classes = useStyles();
 
   const addressNode = (
@@ -23,9 +23,9 @@ const BlockchainAddress = ({ addresses, address }: Props): JSX.Element => {
     </Typography>
   );
 
-  if (addresses.indexOf(address) > -1) {
+  if (userAddresses.includes(address)) {
     return (
-      <Badge variant="text" badgeContent="mine">
+      <Badge variant="text" badgeContent="me">
         {addressNode}
       </Badge>
     );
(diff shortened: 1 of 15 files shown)


webmaster128:
@willclarktech what do you think?

packages/bierzo-wallet/src/components/BlockchainAddress/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+import React from "react";
+
+const useStyles = makeStyles({
+  sectionName: {
+    overflowWrap: "break-word",
+  },
+});
+
+interface Props {
+  readonly addresses: Address[];
```
webmaster128: `addresses` -> `userAddresses` to make this easier to follow

Type `Address[];` -> `readonly Address[];`

packages/bierzo-wallet/src/components/BlockchainAddress/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+const BlockchainAddress = ({ addresses, address }: Props): JSX.Element => {
+  const classes = useStyles();
+
+  const addressNode = (
+    <Typography variant="subtitle2" weight="regular" color="textSecondary" className={classes.sectionName}>
+      {address}
+    </Typography>
+  );
+
+  if (addresses.indexOf(address) > -1) {
```
webmaster128: `addresses.includes(address)` should do the same, right?

packages/bierzo-wallet/src/routes/transactions/index.stories.tsx
```diff
@@ -25,6 +34,153 @@ const lsk: Pick<Token, "tokenTicker" | "fractionalDigits"> = {
...
+function createIdentities(): { [chain: string]: Identity } {
+  const identities: { [chain: string]: Identity } = {};
+
+  // create ETH pub key
+  const ethChain = "ethereum-eip155-5777";
+  const ethIdentity: Identity = {
+    chainId: ethChain as ChainId,
+    pubkey: {
+      algo: Algorithm.Secp256k1,
+      data: new Uint8Array([
```
webmaster128: where did you get this pubkey from? Can you store it as hex for readability?

```ts
data: Encoding.fromHex("aabbcc") as PubkeyBytes,
```

packages/bierzo-wallet/src/routes/transactions/index.tsx
```diff
@@ -7,18 +8,32 @@ import PageMenu from "../../components/PageMenu";
...
   const [rows, setRows] = React.useState(5);
   const [page, setPage] = React.useState(0);
   const [orderBy, setOrderBy] = React.useState(TX_DATE_COLUMN);
   const [order, setOrder] = React.useState(ORDER_DESC);
   const parsedTxs = useSelector((state: RootState) => state.notifications.transactions);
+  const identities = useSelector((state: RootState) => state.extension.identities);
+
+  React.useEffect(() => {
+    async function processIdentities(identities: { [chain: string]: Identity }): Promise<void> {
+      setAddresses((await getChainAddressPair(identities)).map(pair => pair.address));
```
webmaster128: Oh, this function should be called `getChainAddressPairs` (plural) since it returns many pairs

willclarktech:
Looks good to me. The only thing is that it could be a bit confusing given the ambiguity of "mine" between "belonging to me" and "place where mining occurs". I'd suggest just "me", which also sounds slightly more natural anyway.

webmaster128:
>  I'd suggest just "me", which also sounds slightly more natural anyway.

👍 you are the native speaker here. And shorter is better in this case

packages/bierzo-wallet/src/components/BlockchainAddress/index.tsx
albertandrejev: What is the difference between `readonly addresses: Address[];` and `readonly addresses: readonly Address[];`?

webmaster128: In interfaces, `readonly` means you cannot reassign the value. However, a single array is still mutable, i.e. you can do stuff like `.push` or `.unshift`, causing the array reference to change in all places. This can have unintended effects. `readonly Address[]` is the new native syntax for `ReadonlyArray<Address>`, which is like `Array<Address>` but does not have methods that can change the elements.

PR #623, opened 2019-08-29
Feature – Deactivate wallet creation in staging/production

### LINKED ISSUE #608: Deactivate wallet creation for staging/production config
We need to deal with the fact that the browser extension and the wallet do not work without a running network. This means this tooling cannot be used to generate a mainnet address. In order to prepare users as well as governors for the mainnet scenario, we should deactivate wallet creation and create tutorial on how to use external address generators instead.

webmaster128:
Closes #608

### DIFF
--- a/packages/sanes-browser-extension/config/staging.json
+++ b/packages/sanes-browser-extension/config/staging.json
@@ -30,5 +30,6 @@
       }
     }
   ],
+  "walletCreationDisabled": true,
   "accountCreationDisabled": true
 }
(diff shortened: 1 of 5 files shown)


packages/sanes-browser-extension/config/staging.json
```diff
@@ -30,5 +30,6 @@
       }
     }
   ],
+  "walletCreationDisabled": true,
```
apanizo: Should be false?

webmaster128: No, in staging (i.e. testnets), the new address generator should be used as well (see https://github.com/iov-one/address-generator). This is to make testers familiar with the tooling.

Issue #551, opened 2019-08-21
Complete display of governance transactions


willclarktech:
Follow-up to #544 

https://github.com/iov-one/ponferrada/pull/549 adds the logic for `VoteTx` and `CreateProposalTx` but provides only skeleton views on the processed transactions list and request pages. They should be pretty. 💅

webmaster128:
Ticket was split in #725 (vote transaction) and  #726 (create proposal transaction)

PR #626, opened 2019-08-29
Show formatted JSON for governance transactions

webmaster128:
This is a baby step towards #551, making the JSON content from the transaction a bit better readable.

**Before**

Bildschirmfoto 2019-08-29 um 17 30 58
Bildschirmfoto 2019-08-29 um 17 31 06

**After**

create-proposal
vote1
vote2

### DIFF
--- a/packages/bierzo-wallet/src/utils/test/persona.ts
+++ b/packages/bierzo-wallet/src/utils/test/persona.ts
@@ -16,8 +16,8 @@ export const submitExtensionSignupForm = async (page: Page, password: string): P
   await clickCreatePersona(page);
 
   // Fill the form
-  await page.type(`input[name="${PASSWORD_FIELD}`, password);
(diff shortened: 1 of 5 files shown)


webmaster128:
This is a race condition. The page is not (always) rendered when trying to type in the information. I'm adding a short sleep here.

willclarktech:
Looks like there's a missing closing quote on some of these selectors:
```
  // Fill the form
  await page.type(`input[name="${PASSWORD_FIELD}`, password); // <-- here
  await page.type(`input[name="${PASSWORD_CONFIRM_FIELD}`, password); // <-- here
  await page.click(`input[name="${TERMS_ACCEPT_FIELD}"]`);
  await page.click('button[type="submit"]');
```

webmaster128:
True story, let's start with a fix for that

PR #634, opened 2019-09-02
Feature #633 - Transaction Confirmation Page

### LINKED ISSUE #633: Create transaction confirmation page
See the flow in the wallet

This should work 1:1 the same way with the same component/route for create proposal and vote transactions

abefernan:
Closes #633.

### DIFF
--- a/packages/sil-governance/src/index.tsx
+++ b/packages/sil-governance/src/index.tsx
@@ -7,7 +7,7 @@ import Routes from "./routes";
 import { configureStore } from "./store";
 import { globalStyles } from "./theme/globalStyles";
 
-export const store = configureStore();
+const store = configureStore();
 const rootEl = document.getElementById("root");
 
 const render = (Component: React.ComponentType): void => {
--- a/packages/sil-governance/src/index.tsx
+++ b/packages/sil-governance/src/index.tsx
@@ -7,7 +7,7 @@ import Routes from "./routes";
 import { configureStore } from "./store";
 import { globalStyles } from "./theme/globalStyles";
 
-const store = configureStore();
+export const store = configureStore();
 const rootEl = document.getElementById("root");
 
 const render = (Component: React.ComponentType): void => {
--- a/packages/sil-governance/src/routes/create-proposal/components/ProposalForm.tsx
+++ b/packages/sil-governance/src/routes/create-proposal/components/ProposalForm.tsx
@@ -142,9 +142,7 @@ const ProposalForm = (): JSX.Element => {
     const createProposalTx = await governor.buildCreateProposalTx(proposalOptions);
 
     const transactionId = await sendSignAndPostRequest(connection, createProposalTx);
-    if (transactionId) {
-      dispatch(setTransactionsStateAction(transactionId));
-    }
+    dispatch(setTransactionsStateAction(transactionId));
   };
 
   const { form, handleSubmit, invalid, pristine, submitting } = useForm({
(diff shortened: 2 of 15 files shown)


packages/sil-governance/src/routes/create-proposal/components/ProposalForm.tsx
```diff
@@ -138,7 +141,10 @@ const ProposalForm = (): JSX.Element => {
     const proposalOptions = buildProposalOptions(values);
     const createProposalTx = await governor.buildCreateProposalTx(proposalOptions);
 
-    await sendSignAndPostRequest(connection, createProposalTx);
+    const transactionId = await sendSignAndPostRequest(connection, createProposalTx);
+    if (transactionId) {
+      dispatch(setTransactionsStateAction(transactionId));
```
webmaster128: This should also be called for `transactionId === null` because then we know that `lastSignAndPostResult` was a rejection

packages/sil-governance/src/routes/dashboard/components/Proposal/VotingPanel/Buttons.tsx
```diff
@@ -33,8 +36,11 @@ const Buttons = ({ id, vote }: Props): JSX.Element => {
       const connection = await getBnsConnection();
       const voteTx = await governor.buildVoteTx(id, currentVote);
 
-      await sendSignAndPostRequest(connection, voteTx);
-      setPreviousVote(currentVote);
+      const transactionId = await sendSignAndPostRequest(connection, voteTx);
+      if (transactionId) {
+        dispatch(setTransactionsStateAction(transactionId));
```
webmaster128: This should also be called for `transactionId === null` because then we know that `lastSignAndPostResult` was a rejection

packages/sil-governance/src/index.tsx
```diff
@@ -7,7 +7,7 @@ import Routes from "./routes";
 import { configureStore } from "./store";
 import { globalStyles } from "./theme/globalStyles";
 
-const store = configureStore();
+export const store = configureStore();
```
webmaster128: Better to not export this. There are other ways to get the current store

packages/sil-governance/src/routes/create-proposal/index.tsx
```diff
@@ -1,19 +1,37 @@
...
 import Header from "../../components/Header";
+import { RootState } from "../../store/reducers";
+import { setTransactionsStateAction } from "../../store/transactions";
+import { DASHBOARD_ROUTE } from "../paths";
 import ProposalForm from "./components/ProposalForm";
 
+const onReturnToDashboard = (): void => {
+  store.dispatch(setTransactionsStateAction());
+  history.push(DASHBOARD_ROUTE);
+};
```
webmaster128: You can move this function into the component and then use the default way to get the dispatch: `const dispatch = ReactRedux.useDispatch();`

packages/sil-governance/src/routes/dashboard/index.tsx
```diff
@@ -1,19 +1,37 @@
...
 import Header from "../../components/Header";
+import { RootState } from "../../store/reducers";
+import { setTransactionsStateAction } from "../../store/transactions";
+import { DASHBOARD_ROUTE } from "../paths";
 import ProposalsList from "./components/ProposalsList";
 
+const onReturnToDashboard = (): void => {
+  store.dispatch(setTransactionsStateAction());
+  history.push(DASHBOARD_ROUTE);
+};
```
webmaster128: You can move this function into the component and then use the default way to get the dispatch: `const dispatch = ReactRedux.useDispatch();`

PR #638, opened 2019-09-02
Feature #636 - Changes Proposals' Add to Replace

### LINKED ISSUE #636: Change addProposalsAction -> replaceProposalsAction
`addProposalsAction` adds proposals to the list. However, since we only use this on empty lists, we can change the name and functionality to `replaceProposalsAction`. `proposalsReducer` must be udated to do a replace logic instead of add.

The update code

```ts
  const loadProposals = async (): Promise<void> => {
    const governor = store.getState().extension.governor;
    if (!governor) throw new Error("Governor not set in store. This is a bug.");
    const chainProposals = await getProposals(governor);
    dispatch(replaceProposalsAction(chainProposals));
  };
```

can then be called multiple times in order to reload the list of proposals

abefernan:
Closes #636.

### DIFF
--- a/packages/sil-governance/src/routes/create-proposal/components/ProposalForm.tsx
+++ b/packages/sil-governance/src/routes/create-proposal/components/ProposalForm.tsx
@@ -8,6 +8,7 @@ import * as ReactRedux from "react-redux";
 
 import { sendSignAndPostRequest } from "../../../communication/signandpost";
 import { getBnsConnection } from "../../../logic/connection";
+import { getProposals, replaceProposalsAction } from "../../../store/proposals";
 import { RootState } from "../../../store/reducers";
 import { setTransactionsStateAction } from "../../../store/transactions";
 import CommitteeRulesSelect from "./CommitteeRulesSelect";
@@ -142,6 +143,16 @@ const ProposalForm = (): JSX.Element => {
     const createProposalTx = await governor.buildCreateProposalTx(proposalOptions);
 
     const transactionId = await sendSignAndPostRequest(connection, createProposalTx);
+
+    const updateChainProposals = async (): Promise<void> => {
+      const chainProposals = await getProposals(governor);
+      dispatch(replaceProposalsAction(chainProposals));
+    };
+
+    setTimeout(() => {
+      updateChainProposals();
+    }, 5000);
+
     dispatch(setTransactionsStateAction(transactionId));
   };
 
--- a/packages/sil-governance/src/routes/create-proposal/components/ProposalForm.tsx
+++ b/packages/sil-governance/src/routes/create-proposal/components/ProposalForm.tsx
@@ -149,9 +149,7 @@ const ProposalForm = (): JSX.Element => {
       dispatch(replaceProposalsAction(chainProposals));
     };
 
-    setTimeout(() => {
-      updateChainProposals();
-    }, 5000);
+    setTimeout(() => updateChainProposals(), 5000);
 
     dispatch(setTransactionsStateAction(transactionId));
   };
--- a/packages/sil-governance/src/routes/create-proposal/index.tsx
+++ b/packages/sil-governance/src/routes/create-proposal/index.tsx
@@ -6,6 +6,7 @@ import { history } from "..";
 import AsideFilter from "../../components/AsideFilter";
 import ConfirmTransaction from "../../components/ConfirmTransaction";
 import Header from "../../components/Header";
+import { replaceProposalsAction } from "../../store/proposals";
 import { RootState } from "../../store/reducers";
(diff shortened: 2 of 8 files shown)


packages/sil-governance/src/routes/create-proposal/index.tsx
```diff
@@ -14,10 +15,12 @@ import ProposalForm from "./components/ProposalForm";
 const CreateProposal = (): JSX.Element => {
   const dispatch = useDispatch();
   const lastSignAndPostResult = useSelector((state: RootState) => state.transactions.lastSignAndPostResult);
+  const proposals = useSelector((state: RootState) => state.proposals);
 
   const onReturnToDashboard = (): void => {
     dispatch(setTransactionsStateAction());
     history.push(DASHBOARD_ROUTE);
+    dispatch(replaceProposalsAction(proposals));
```
webmaster128: Why this change? Here you take data from the store and write it to the store.

To reload from chain, we need something like

```ts
  const loadProposals = async (): Promise<void> => {
    const governor = store.getState().extension.governor;
    if (!governor) throw new Error("Governor not set in store. This is a bug.");
    const chainProposals = await getProposals(governor);
    dispatch(replaceProposalsAction(chainProposals));
  };
```

packages/sil-governance/src/routes/dashboard/index.tsx
```diff
@@ -14,10 +15,12 @@ import ProposalsList from "./components/ProposalsList";
 const Dashboard = (): JSX.Element => {
   const dispatch = useDispatch();
   const lastSignAndPostResult = useSelector((state: RootState) => state.transactions.lastSignAndPostResult);
+  const proposals = useSelector((state: RootState) => state.proposals);
 
   const onReturnToDashboard = (): void => {
     dispatch(setTransactionsStateAction());
     history.push(DASHBOARD_ROUTE);
+    dispatch(replaceProposalsAction(proposals));
```
webmaster128: as above

packages/sil-governance/src/routes/create-proposal/index.tsx
abefernan: Sure, I messed up there because I began to follow the suggested types... The issue I'm having is `governor.getProposals()` returns `Proposal[]` and `replaceProposalsAction` expects `ProposalProps[]`.

webmaster128: When you get the transaction ID, it was just posted into the network. I.e. it is not yet included in a block. This is the reason I suggetsed a 5 second delay between getting the result and triggering a reload

abefernan: The delay should be before showing the confirmation page, or before redirecting to Dashboard after confirming the return?

webmaster128: The best thing would be to have the reload independent of the screen navigation, like 5 seconds after you received the transaction ID (triggered by `setTimeout`)

packages/sil-governance/src/routes/create-proposal/components/ProposalForm.tsx
```diff
@@ -142,6 +143,16 @@ const ProposalForm = (): JSX.Element => {
...
     const transactionId = await sendSignAndPostRequest(connection, createProposalTx);
+
+    const updateChainProposals = async (): Promise<void> => {
+      const chainProposals = await getProposals(governor);
+      dispatch(replaceProposalsAction(chainProposals));
+    };
+
+    setTimeout(() => {
+      updateChainProposals();
+    }, 5000);
```
webmaster128: Suggestion: this can be written as a on-liner:

```ts
setTimeout(() => updateChainProposals(), 5000);
```

packages/sil-governance/src/routes/dashboard/components/Proposal/VotingPanel/Buttons.tsx
```diff
@@ -37,6 +38,16 @@ const Buttons = ({ id, vote }: Props): JSX.Element => {
...
       const transactionId = await sendSignAndPostRequest(connection, voteTx);
+
+      const updateChainProposals = async (): Promise<void> => {
+        const chainProposals = await getProposals(governor);
+        dispatch(replaceProposalsAction(chainProposals));
+      };
+
+      setTimeout(() => {
+        updateChainProposals();
+      }, 5000);
```
webmaster128: same here

packages/sil-governance/src/routes/create-proposal/components/ProposalForm.tsx
abefernan: Sure! There were more lines before

PR #642, opened 2019-09-03
Storybook version upgrade

albertandrejev:


webmaster128:
Shouldn't this change the output format of the snapshot?

albertandrejev:
Not sure. I was ran the snapshot tests and them was passed successfully.

webmaster128:
okay sorry, I misunderstoof the diff in the storybook snapshot. Looks like it is not caused by the upgrade. Great.

PR #643, opened 2019-09-03
Feature #603 - CSV builder

### LINKED ISSUE #603: Create interface for CSV representation if different transaction types
Right now it is very hard to keep different implementations of `csvRepresentation()` consistent. In `BwUnkownParser` for example this is currently implemented as an empty string but transaction ID and fees could be available in the CSV row instead.

For simplicity, all fields can be `string`, like we have right now. Some fields can be optional. E.g.

```ts
interface CsvRepresentation {
  readonly transactionId: string;
  readonly blockTime: string;
  readonly feeQuantity?: string;
  readonly feeFractionalDigits?: string;
  readonly feeTokenTicker?: string;
  // ...
}
```

The same interface should be used for the header to ensure consistency:

```ts
const csvHeader: CsvRepresentation = {
  // titles here
}
```

In `onDownloadCSV`, all the undefiend fields get converted to empty strings to ensure the position of columns is consistent even if not all transaction types provide all fields.

albertandrejev:
**Description**
This PR will create CSV representation builder. Closes #603.

### DIFF
--- a/packages/bierzo-wallet/src/logic/csvBuilder/index.ts
+++ b/packages/bierzo-wallet/src/logic/csvBuilder/index.ts
@@ -14,28 +14,36 @@ export interface CsvRow {
 }
 
 class CsvRepresentation {
-  private readonly fields: readonly string[];
-  private readonly header: string;
-  private readonly rows: CsvRow[] = [];
+  private readonly _fields: readonly string[];
+  private readonly _header: string;
+  private readonly _rows: CsvRow[] = [];
 
   public constructor(header: CsvRow) {
-    this.fields = Object.keys(header);
-    this.header = this.parse(header);
+    this._fields = Object.keys(header);
+    this._header = this.parse(header);
   }
 
   public addRow(row: CsvRow): void {
-    this.rows.push(row);
+    this._rows.push(row);
+  }
+
+  public get csvRows(): CsvRow[] {
+    return this._rows;
+  }
+
+  public get csvHeader(): string {
+    return this._header;
   }
 
   public blob(): Blob {
-    const csvBody = this.rows.map((row: CsvRow) => this.parse(row));
+    const csvBody = this._rows.map((row: CsvRow) => this.parse(row));
 
-    return new Blob([`${this.header}\n${csvBody.join("\n")}`], { type: "text/plain;charset=utf-8" });
+    return new Blob([`${this._header}\n${csvBody.join("\n")}`], { type: "text/plain;charset=utf-8" });
   }
 
   private parse(row: CsvRow): string {
     const parsedRow: string[] = [];
-    this.fields.forEach(field => parsedRow.push(`"${row[field]}"`));
+    this._fields.forEach(field => parsedRow.push(`"${row[field]}"`));
 
     return parsedRow.join(";");
   }
--- a/packages/bierzo-wallet/src/logic/csvBuilder/index.ts
+++ b/packages/bierzo-wallet/src/logic/csvBuilder/index.ts
@@ -1,3 +1,5 @@
+import FileSaver from "file-saver";
+
 export interface CsvRow {
   readonly id: string;
   readonly recepient: string;
@@ -11,3 +13,33 @@ export interface CsvRow {
   readonly time: string;
   readonly note: string;
 }
+
+class CsvRepresentation {
+  private readonly fields: readonly string[];
+  private readonly header: string;
+  private readonly rows: CsvRow[] = [];
+
+  public constructor(header: CsvRow) {
+    this.fields = Object.keys(header);
+    this.header = this.parse(header);
+  }
+
+  public addRow(row: CsvRow): void {
+    this.rows.push(row);
+  }
+
+  public get(): Blob {
+    const csvBody = this.rows.map((row: CsvRow) => this.parse(row));
+
+    return new Blob([`${this.header}\n${csvBody.join("\n")}`], { type: "text/plain;charset=utf-8" });
+  }
+
+  private parse(row: CsvRow): string {
+    const parsedRow: string[] = [];
+    this.fields.forEach(field => parsedRow.push(`"${row[field]}"`));
+
+    return parsedRow.join(";");
+  }
+}
+
+export default CsvRepresentation;
--- a/packages/bierzo-wallet/src/logic/csvBuilder/index.ts
+++ b/packages/bierzo-wallet/src/logic/csvBuilder/index.ts
@@ -1,5 +1,3 @@
-import FileSaver from "file-saver";
-
 export interface CsvRow {
   readonly id: string;
   readonly recepient: string;
@@ -12,6 +10,7 @@ export interface CsvRow {
   readonly feeTokenTicker: string;
   readonly time: string;
   readonly note: string;
+  [key: string]: string;
 }
 
 class CsvRepresentation {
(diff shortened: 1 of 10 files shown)


packages/bierzo-wallet/src/logic/csvBuilder/index.ts
```diff
@@ -0,0 +1,52 @@
...
+  readonly sender: string;
+  readonly quantity: string;
+  readonly fractionalDigits: string;
+  readonly tokenTicker: string;
+  readonly feeQuantity: string;
+  readonly feeFractionalDigits: string;
+  readonly feeTokenTicker: string;
+  readonly time: string;
+  readonly note: string;
+  [key: string]: string;
```
webmaster128: Can you remove this signature, since it is not very type safe?

packages/bierzo-wallet/src/logic/csvBuilder/index.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+class CsvRepresentation {
+  private readonly _fields: readonly string[];
```
webmaster128: Let's drop this field

packages/bierzo-wallet/src/logic/csvBuilder/index.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  public blob(): Blob {
+    const csvBody = this._rows.map((row: CsvRow) => this.parse(row));
+
+    return new Blob([`${this._header}\n${csvBody.join("\n")}`], { type: "text/plain;charset=utf-8" });
+  }
+
+  private parse(row: CsvRow): string {
+    const parsedRow: string[] = [];
+    this._fields.forEach(field => parsedRow.push(`"${row[field]}"`));
```
webmaster128: and access the field explicitely here

packages/bierzo-wallet/src/logic/csvBuilder/index.ts
```diff
@@ -0,0 +1,52 @@
...
+  readonly feeTokenTicker: string;
+  readonly time: string;
+  readonly note: string;
+  [key: string]: string;
+}
+
+class CsvRepresentation {
+  private readonly _fields: readonly string[];
+  private readonly _header: string;
+  private readonly _rows: CsvRow[] = [];
```
webmaster128: Why the `_ ` prefixes? We don't use them in the rest of the project as far as I know

packages/bierzo-wallet/src/logic/csvBuilder/index.ts
```diff
@@ -0,0 +1,52 @@
...
+    return this._header;
+  }
+
+  public blob(): Blob {
+    const csvBody = this._rows.map((row: CsvRow) => this.parse(row));
+
+    return new Blob([`${this._header}\n${csvBody.join("\n")}`], { type: "text/plain;charset=utf-8" });
+  }
+
+  private parse(row: CsvRow): string {
```
webmaster128: This can be `private static`, since it does not access instance members

packages/bierzo-wallet/src/logic/csvBuilder/index.ts
```diff
@@ -0,0 +1,52 @@
...
+  public addRow(row: CsvRow): void {
+    this._rows.push(row);
+  }
+
+  public get csvRows(): CsvRow[] {
+    return this._rows;
+  }
+
+  public get csvHeader(): string {
+    return this._header;
```
webmaster128: Why not store/return the header as `CsvRow` as well for symmetry with the rows?

packages/bierzo-wallet/src/logic/csvBuilder/index.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  public blob(): Blob {
+    const csvBody = this._rows.map((row: CsvRow) => this.parse(row));
+
+    return new Blob([`${this._header}\n${csvBody.join("\n")}`], { type: "text/plain;charset=utf-8" });
```
webmaster128: type: "text/csv;charset=utf-8"

packages/bierzo-wallet/src/routes/transactions/index.tsx
```diff
@@ -73,13 +73,30 @@ const Transactions = (): JSX.Element => {
...
+
     // TODO UPDATE HEADER WHEN OTHER TX TYPES ARE ADDED
-    const csvHeader =
+    /*const csvHeader =
       '"ID";"Recipient";"Sender";"Quantity";"Fractional Digits";"Token Ticker";"Fee Quantity";"Fee Fractional Digits";"Fee Token Ticker";"Time";"Note"';
     const csvBody = orderedTxs.map((tx: ProcessedTx) => BwParserFactory.getCsvRepresentation(tx));
 
-    const blob = new Blob([`${csvHeader}\n${csvBody.join("\n")}`], { type: "text/plain;charset=utf-8" });
-    FileSaver.saveAs(blob, "transactions.csv");
+    const blob = new Blob([`${csvHeader}\n${csvBody.join("\n")}`], { type: "text/plain;charset=utf-8" });*/
```
webmaster128: All those old comments can be removed now

packages/bierzo-wallet/src/logic/csvBuilder/index.ts
albertandrejev: If I will remove it, I will get error from TS because in this case it can not access property by name.

packages/bierzo-wallet/src/logic/csvBuilder/index.ts
albertandrejev: We can not be 100% sure about order of the properties, so I want to used order of properties from Header row.

packages/bierzo-wallet/src/logic/csvBuilder/index.ts
albertandrejev: I want to preserve order of the header fields.And use it for row values.

packages/bierzo-wallet/src/logic/csvBuilder/index.ts
webmaster128: But do you need this index signature when you do the following instead?

```ts
const parsedRow: string[] = [
  row.id,
  row.recepient,
  row.sender,
  row.quantity,
  row.fractionalDigits,
  row.tokenTicker,
  row.feeQuantity,
  row.feeFractionalDigits,
  row.feeTokenTicker,
  row.time,
  row.note,
].map(content => `"${content}"`);
```

albertandrejev: In this case if we will update CsvRow interface then we will need to update this map values too. Easily to miss.

packages/bierzo-wallet/src/logic/csvBuilder/index.ts
albertandrejev: And we can manipulate order of the fields using this variable

packages/bierzo-wallet/src/logic/csvBuilder/index.ts
webmaster128: Right, but only 2 places need to be in sync and it makes everything else much easier.

With `this._fields = Object.keys(header);` you have no guarantee that `header` contains only `CsvRow` fields. It can have more, unsupported fields.

So we get
1. Guaranteed content
2. Explicit order
3. Removal of usafe index signature
4. No `_fields` necessary
5. `parse` can become static

PR #645, opened 2019-09-03
Feature #281 - Side Filter Implementation

### LINKED ISSUE #281: Create main menu (left column) with the 3 filter options
As I user I can select one of three options

* All elections [ default ]
* Active elections
* Closed elections 

please see : 
https://www.figma.com/file/ffb3NsuSizFxKzwApMxewRgN/Voting?node-id=0%3A1

abefernan:
Closes #281.


webmaster128:
> right now when a filter button is clicked from the Create Proposal View it redirects to Dashboard but with the All Elections default filter applied. Not sure about how to share this state between to routes.

I think a good approach is to have different routes for the different main menu items. Depending on the route you can render the same component with different filters (similar to https://github.com/iov-one/address-generator/blob/master/src/index.tsx#L16-L22)

Issue #646, opened 2019-09-03
Make wallet usable with Ledger as an alternative to Neuma


webmaster128:
Because of https://github.com/iov-one/ponferrada/issues/527#issuecomment-527463496, we now follow a different approach to Ledger integration. Instead of adding Ledger support to the extension, we add it to the wallet directly via WebUSB.

The [address generator](https://github.com/iov-one/address-generator) is a proof of concept showing that WebUSB works with Chromium-based browsers.

It is relatively simple to communicate to a Ledger device the same way as we do with the extension right now. The interface has basically two calls:

1. get identitites
2. sign (and post)

For 2. the wallet uses the ledger for signing only and posts the transaction to the chain directly.

webmaster128:
Done in #646

PR #647, opened 2019-09-03
Feature – Improve identitities source flexibility

webmaster128:
In order to prepare [the new Ledger stategy](https://github.com/iov-one/ponferrada/issues/646), this cleans up identities management to allow for easier Ledger integration.

Main changes:
1. Remove `installed` and `connected` from redux, which are one-time values rather than app-global data
2. Simplify extension state -> identitites state
3. Use `Map<ChainId, Identity>` to store identitites fot better type safety
4. ~Change target to es6, which allows some more modern loop syntax~


webmaster128:
I don't know what the problem is here. Need to break down into smaller steps. @albertandrejev can you review https://github.com/iov-one/ponferrada/pull/651 instead? Those are the simple parts of this PR

packages/bierzo-wallet/src/communication/extension.ts
```diff
@@ -0,0 +1,45 @@
...
+ * Groups the identites by blockchain. Returns the first identity of each chain.
+ *
+ * The browser extension is supposed to only send one identity per chain.
+ * @see https://github.com/iov-one/ponferrada/issues/447
+ */
+export function groupIdentitiesByChain(identities: readonly Identity[]): ReadonlyMap<ChainId, Identity> {
+  const out = new Map<ChainId, Identity>();
+  for (const identity of identities) {
+    if (!out.has(identity.chainId)) {
+      out.set(identity.chainId, identity);
```
albertandrejev: You can `set` new value without checking for it's existence.

webmaster128: Right, but this overrides existing entries. In this function we always want to use the first entry for any given chain ID. Because of that, if the chain ID exists already we don't override.

albertandrejev: thanks

webmaster128:
After having massive issues with asynchonous component changes in DOM tests, we now store `chainName` and `Address` together with the identity in the redux store. This allows for easy, synchonous access in many many use cases, including the DOM tests we have (see last 2 commits).

PR #653, opened 2019-09-04
Feature #578 - Configure expected chain

### LINKED ISSUE #578: Configure expected chain IDs
In

```ts
  const expectedBnsChainId: ChainId | undefined = undefined;
  const expectedLiskChainId: ChainId | undefined = undefined;
  const expectedEthereumChainId: ChainId | undefined = undefined;
```

the expected chain IDs should be read from a config file

albertandrejev:
**Description**
This PR will provide expected chain id to the `create<Chain Name>Connector` method. 

Closes #578.

### DIFF
--- a/packages/sanes-browser-extension/src/extension/background/model/persona/config/codec.ts
+++ b/packages/sanes-browser-extension/src/extension/background/model/persona/config/codec.ts
@@ -5,7 +5,7 @@ import { createEthereumConnector } from "@iov/ethereum";
 import { HdPaths } from "@iov/keycontrol";
 import { createLiskConnector } from "@iov/lisk";
 
-import { CodecString } from "./configurationfile";
+import { CodecString, getConfigurationFile } from "./configurationfile";
 
 export enum CodecType {
   Bns,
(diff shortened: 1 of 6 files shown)


packages/sanes-browser-extension/src/extension/background/model/persona/config/codec.ts
```diff
@@ -54,22 +54,19 @@ export function pathBuilderForCodec(codecType: CodecType): (derivation: number)
   return pathBuilder;
 }
 
-export function chainConnector(
+export async function chainConnector(
```
webmaster128: Why async?

packages/sanes-browser-extension/src/extension/background/model/persona/index.ts
```diff
@@ -152,7 +152,7 @@ export class Persona {
     const out: SoftwareAccountManagerChainConfig[] = [];
     for (const chainSpec of config.chains.map(chain => chain.chainSpec)) {
       const codecType = codecTypeFromString(chainSpec.codecType);
-      const connector = chainConnector(codecType, chainSpec.node, chainSpec.scraper);
+      const connector = await chainConnector(codecType, chainSpec.chainId, chainSpec.node, chainSpec.scraper);
```
webmaster128: If we can make `chainConnector` synchonous again, we can drop this `await`

packages/sanes-browser-extension/src/extension/background/model/persona/config/codec.ts
albertandrejev: You are right, I had used another way but forgot to remove this `async`

PR #661, opened 2019-09-04
Feature – Add initial Ledger support to Wallet

### LINKED ISSUE #527: Ledger integration

webmaster128:
Closes #527


packages/bierzo-wallet/src/communication/ledgerRpcEndpoint.ts
```diff
@@ -0,0 +1,108 @@
...
+      throw new Error("Unsupported request format");
+    }
+
+    const transaction = TransactionEncoder.fromJson(request.params.transaction);
+
+    const bnsConnection = await getConnectionForBns();
+    const nonce = await bnsConnection.getNonce({ pubkey: transaction.creator.pubkey });
+    const { bytes } = bnsCodec.bytesToSign(transaction, nonce);
+
+    const transport = await TransportWebUSB.create(5000);
```
albertandrejev: There is same code inside `try ... catch` https://github.com/iov-one/ponferrada/blob/a2454c4eb2e16a9b4c67de20af176b311ab1e645/packages/bierzo-wallet/src/communication/ledgerRpcEndpoint.ts#L28, why there is difference? Also it would be great to move this code and following code with different validations for `IovLedgerApp` to the separate entity.

packages/bierzo-wallet/src/store/rpcendpoint/actions.ts
```diff
@@ -0,0 +1,16 @@
+import { Action } from "redux";
+import { ActionType } from "typesafe-actions";
+
+import { RpcEndpoint } from "../../communication/rpcEndpoint";
+
+export interface SetRpcEndpointActionType extends Action {
+  type: "@@rpcendpoint/SET_RPC_ENDPOINT";
```
albertandrejev: This type name is little redundant because you already have `@@rpcendpoint` in type name, so better name will be `@@rpcendpoint/SET`.

packages/bierzo-wallet/src/communication/ledgerRpcEndpoint.ts
webmaster128: The reason is that (for historic reasons), the getIdentities call has a specific return type for "not installed" (`undefined`). But I agree this should be consistent. I'll cleanup the interface.

packages/bierzo-wallet/src/store/rpcendpoint/actions.ts
webmaster128: Done in https://github.com/iov-one/ponferrada/pull/661/commits/6142d8970489eb77e2da0180987d16dff3b63b68

packages/bierzo-wallet/src/communication/ledgerRpcEndpoint.ts
webmaster128: Done in https://github.com/iov-one/ponferrada/pull/661/commits/b01a520050e8dfb9a846bc655a372c1151025faf

PR #682, opened 2019-09-08
Feature #622 - Rename "New wallet" to "Create wallet"

### LINKED ISSUE #622: Change button "New Wallet" -> "Create Wallet"
Buttons should always contain actions

albertandrejev:
**Description**
This PR will rename "New wallet" button to "Create wallet"  button.

Closes #622.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,6 @@
     "@types/storybook__react": "^4.0.2",
     "@typescript-eslint/eslint-plugin": "^1.13.0",
     "@typescript-eslint/parser": "^1.13.0",
-    "core-js": "^2.4.0",
     "eslint-config-prettier": "^4.1.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -64,6 +64,7 @@
     "@types/storybook__react": "^4.0.2",
     "@typescript-eslint/eslint-plugin": "^1.13.0",
     "@typescript-eslint/parser": "^1.13.0",
+    "core-js": "^2.4.0",
```
albertandrejev: This package is required for storybook, otherwise it will throw an error and will not start.

webmaster128: Thanks for pointing this out. In https://github.com/iov-one/ponferrada/pull/684 there is a better solution for the problem (using less dependencies)

PR #721, opened 2019-09-13
Feature #701 - Show unlock when wallet exists

### LINKED ISSUE #701: Only show Unlock button when a locked wallet (database) exists
This is how the screen looks for newly installed extensions:

Bildschirmfoto 2019-09-11 um 10 37 57

The "Unlock" button is misleading, since this is not an option when the database exists. It should only be displayed when there is a database (see `hasStoredPersona()` helper)

albertandrejev:
**Description**
Extension's Welcome screen should show Unlock button only in case if there is database created for the person in other case Unlock button should not be shown. Closes #701.

### DIFF
--- a/packages/bierzo-wallet/src/utils/test/persona.ts
+++ b/packages/bierzo-wallet/src/utils/test/persona.ts
@@ -9,7 +9,7 @@ const PASSWORD_CONFIRM_FIELD = "passwordConfirmInputField";
 const TERMS_ACCEPT_FIELD = "termsAcceptCheckboxField";
 
 async function clickCreatePersona(page: Page): Promise<void> {
-  await page.click("button:nth-of-type(1)");
+  await page.click("#welcome-create-wallet");
 }
 
 export const submitExtensionCreateWalletForm = async (page: Page, password: string): Promise<void> => {
--- a/packages/bierzo-wallet/src/utils/test/persona.ts
+++ b/packages/bierzo-wallet/src/utils/test/persona.ts
@@ -9,7 +9,7 @@ const PASSWORD_CONFIRM_FIELD = "passwordConfirmInputField";
 const TERMS_ACCEPT_FIELD = "termsAcceptCheckboxField";
 
 async function clickCreatePersona(page: Page): Promise<void> {
-  await page.click("button:nth-of-type(2)");
+  await page.click("button:nth-of-type(1)");
 }
 
 export const submitExtensionCreateWalletForm = async (page: Page, password: string): Promise<void> => {
--- a/packages/sanes-browser-extension/src/context/PersonaProvider.tsx
+++ b/packages/sanes-browser-extension/src/context/PersonaProvider.tsx
@@ -14,38 +14,37 @@ export interface PersonaContextUpdateData {
   readonly accounts?: readonly PersonaAcccount[];
   readonly mnemonic?: string;
   readonly txs?: readonly ProcessedTx[];
+  readonly hasStoredPersona?: boolean;
 }
 
 export interface PersonaContextInterface {
   readonly accounts: readonly PersonaAcccount[];
   readonly txs: readonly ProcessedTx[];
   readonly mnemonic: string;
-  readonly hasPersona: boolean;
+  readonly hasStoredPersona: boolean;
   readonly update: (newData: PersonaContextUpdateData) => void;
-  readonly clearPersona: () => void;
 }
 
 export const PersonaContext = React.createContext<PersonaContextInterface>({
   accounts: [],
   mnemonic: "",
   txs: [],
-  hasPersona: false,
+  hasStoredPersona: false,
   update: (): void => {},
-  clearPersona: (): void => {},
 });
 
 interface Props {
   readonly children: React.ReactNode;
   readonly persona: GetPersonaResponse;
-  readonly hasPersona: boolean;
+  readonly hasStoredPersona: boolean;
 }
 
 type Accounts = readonly PersonaAcccount[];
 
-export const PersonaProvider = ({ children, persona, hasPersona }: Props): JSX.Element => {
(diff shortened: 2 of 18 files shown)


packages/bierzo-wallet/src/utils/test/persona.ts
```diff
@@ -9,7 +9,7 @@ const PASSWORD_CONFIRM_FIELD = "passwordConfirmInputField";
 const TERMS_ACCEPT_FIELD = "termsAcceptCheckboxField";
 
 async function clickCreatePersona(page: Page): Promise<void> {
-  await page.click("button:nth-of-type(2)");
+  await page.click("button:nth-of-type(1)");
```
webmaster128: Can we get this button by ID or something similar, which is more stable? Especially in this case where the number of buttons are changing depending on the state of the extension

packages/sanes-browser-extension/src/index.tsx
```diff
@@ -33,20 +34,19 @@ const render = (
...
+type PersonaContexType = [GetPersonaResponse, boolean];
 
-  hasStoredPersona().then(hasPersona => {
-    const hasRequests = requests.length > 0;
-    const url = initialUrl(!!persona, hasPersona, hasRequests);
-    history.push(url);
+Promise.all([getPersonaData(), hasStoredPersona()]).then(([persona, hasPersona]: PersonaContexType) => {
+  const requests = getQueuedRequests();
+  const hasRequests = requests.length > 0;
+  const url = initialUrl(!!persona, hasPersona, hasRequests);
```
webmaster128: I think the variable name `hasPersona` is not optimal because `persona` can be `null` as well, so it is hard to understand here the difference is. Can you do the following renamings to make clear this is about the database?

1. `hasStoredPersona()` -> `getHasStoredPersona()`
2. `hasPersona` -> `hasStoredPersona`

packages/sanes-browser-extension/src/context/PersonaProvider.tsx
```diff
@@ -20,26 +20,32 @@ export interface PersonaContextInterface {
   readonly accounts: readonly PersonaAcccount[];
   readonly txs: readonly ProcessedTx[];
   readonly mnemonic: string;
+  readonly hasPersona: boolean;
   readonly update: (newData: PersonaContextUpdateData) => void;
+  readonly clearPersona: () => void;
```
webmaster128: Can't we use `personaProvider.update` for that?

packages/sanes-browser-extension/src/routes/welcome/index.dom.spec.ts
```diff
@@ -21,6 +21,12 @@ describe("DOM > Feature > Welcome", () => {
     expect(buttons.length).toBe(3);
   }, 60000);
 
+  it("should contain two buttons in case if no persona exists", async () => {
+    const welcomeDom = await travelToWelcome(true);
+    const buttons = TestUtils.scryRenderedDOMComponentsWithTag(welcomeDom, "button");
+    expect(buttons.length).toBe(3);
```
webmaster128: Shouldn't this be 2, like the description says?

packages/sanes-browser-extension/src/utils/test/navigation.ts
```diff
@@ -46,8 +47,8 @@ export const travelTo = async (
   return dom;
 };
 
-export const travelToWelcome = async (): Promise<React.Component> => {
-  return travelTo(WELCOME_ROUTE);
+export const travelToWelcome = async (hasPersona: boolean): Promise<React.Component> => {
+  return travelTo(WELCOME_ROUTE, [], null, hasPersona);
```
webmaster128: you can set the 2 arguments to `undefined`, which is the same at not setting them. Then the caller does not need to bother what good default values are.

packages/sanes-browser-extension/src/routes/create-wallet/test/operateCreateWallet.ts
```diff
@@ -139,7 +139,7 @@ export const submitSecurityHint = async (securityHintDom: React.Component, hint:
 };
 
 export const travelToCreateWalletNewWalletStep = async (page: Page): Promise<void> => {
-  await page.click("button:nth-of-type(2)");
+  await page.click("button:nth-of-type(1)");
```
webmaster128: `await page.click("#welcome-create-wallet");` as above?

packages/sanes-browser-extension/src/routes/restore-wallet/test/operateRestoreWallet.ts
```diff
@@ -71,9 +71,6 @@ export const submitRecoveryPhraseE2E = async (
 };
 
 export const travelToRestoreWalletStep = async (page: Page): Promise<void> => {
-  await page.click("button:nth-of-type(1)");
-  await findRenderedE2EComponentWithId(page, UNLOCK_ROUTE);
-
-  await page.click("a:nth-of-type(1)");
+  await page.click("button:nth-of-type(2)");
```
webmaster128: Can we use an ID here too because this button has a different number depending on the `hasStoredPersona` state

packages/sanes-browser-extension/src/routes/welcome/index.stories.tsx
```diff
@@ -2,14 +2,35 @@ import { storiesOf } from "@storybook/react";
...
+      <PersonaProvider persona={persona} hasStoredPersona={true}>
+        <Storybook>
+          <Layout />
+        </Storybook>
+      </PersonaProvider>
+    ),
+  )
+  .add(
+    "Without persona",
+    (): JSX.Element => (
```
webmaster128: Return types of unnames function not needed `() => (`, which makes storybook and test code much shorter

PR #724, opened 2019-09-16
Feature #723 – Use blockchain clock for hasStarted

### LINKED ISSUE #723: Add blockchain clock
In order to make transparent to the user what the current blockchain time is (which is when voting really starts and ends), a small UI element should show the current time

webmaster128:
Closes #723

**After**

Bildschirmfoto 2019-09-16 um 11 30 09

Bildschirmvideo-aufnehmen-2019-09-16-um-11 30 57

### DIFF
--- a/packages/medulas-react-components/src/components/forms/TextFieldForm/index.tsx
+++ b/packages/medulas-react-components/src/components/forms/TextFieldForm/index.tsx
@@ -22,6 +22,7 @@ const TextFieldForm = ({ name, form, validate, ...restProps }: Props): JSX.Eleme
     <MuiTextField
       error={error}
       name={input.name}
+      value={input.value}
       helperText={error ? meta.error : undefined}
       onChange={input.onChange}
       margin="normal"
--- a/packages/sil-governance/src/components/BlockchainTime/index.tsx
+++ b/packages/sil-governance/src/components/BlockchainTime/index.tsx
@@ -1,0 +1,30 @@
+import { makeStyles } from "@material-ui/styles";
+import { Block, Typography } from "medulas-react-components";
+import React from "react";
+import { ReadonlyDate } from "readonly-date";
+
+interface Props {
+  readonly lastBlockHeight: number;
+  readonly lastBlockTime: ReadonlyDate;
+}
+
+const useStyles = makeStyles({
+  container: {
+    textAlign: "center",
+  },
+});
+
+const BlockchainTime = ({ lastBlockHeight, lastBlockTime }: Props): JSX.Element => {
+  const classes = useStyles();
+  const datetime = lastBlockTime.toLocaleString();
+  return (
+    <Block className={classes.container} padding={1}>
+      <Typography variant="body1">Blockchain time</Typography>
+      <Typography variant="body2">
+        {datetime} (Height&nbsp;{lastBlockHeight})
+      </Typography>
+    </Block>
+  );
+};
+
+export default BlockchainTime;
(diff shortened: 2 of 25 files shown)


webmaster128:
DOM tests updated. https://github.com/iov-one/ponferrada/pull/724/commits/732eb8b7c656a89b30d26f2d79f626f39fecc7d6 fixes the input bug I had @albertandrejev

packages/medulas-react-components/src/components/forms/TextFieldForm/index.tsx
```diff
@@ -22,6 +22,7 @@ const TextFieldForm = ({ name, form, validate, ...restProps }: Props): JSX.Eleme
     <MuiTextField
       error={error}
       name={input.name}
+      value={input.value}
```
albertandrejev: If you want to set initial value for the form you can use `initalValues` as well during form initialization. https://github.com/final-form/react-final-form-hooks/issues/2#issuecomment-438323339

packages/sil-governance/src/components/BlockchainTime/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { ReadonlyDate } from "readonly-date";
+
+interface Props {
+  readonly lastBlockHeight: number;
+  readonly lastBlockTime: ReadonlyDate;
+}
+
+const useStyles = makeStyles({
+  container: {
+    textAlign: "center",
```
albertandrejev: Have you tried `align` prop of `Typography` component? https://material-ui.com/ru/api/typography/

packages/sil-governance/src/logic/boot.ts
```diff
@@ -0,0 +1,53 @@
...
+  // Subscriptions
+
+  const current = Stream.fromPromise(connection.height().then(height => connection.getBlockHeader(height)));
+  const updates = connection.watchBlockHeaders();
+  const blockHeadersSubscription = Stream.merge(current, updates).subscribe({
+    next: header => {
+      dispatch(
+        setBlockchainAction({
+          connection: connection,
+          subscriptions: [blockHeadersSubscription],
```
albertandrejev: I think there is no need to update `connection` and `subscriptions` every time on new header. So you can create new action to update current block header data.

packages/sil-governance/src/store/blockchain/actions.ts
```diff
@@ -0,0 +1,6 @@
+import { BlockchainState, SetBlockchainActionType } from "./reducer";
+
+export const setBlockchainAction = (blockchain: BlockchainState): SetBlockchainActionType => ({
+  type: "@@blockchain/SET",
+  payload: blockchain,
+});
```
albertandrejev: You can define additional action like `setBlockchainHeaderAction` here, with type: `@@blockchain/SET_HEADER`

packages/medulas-react-components/src/components/forms/TextFieldForm/index.tsx
webmaster128: I tried this but it was not working in our setup. I don't understand why. However, an initial values is not needed at the moment as long as the user input is preserved.

packages/sil-governance/src/components/BlockchainTime/index.tsx
webmaster128: No, thanks. That would work as well I guess. My idea is to set the textAlign for the whole container to include all `<Typography>`s

packages/sil-governance/src/logic/boot.ts
webmaster128: Thanks. I'll take it into account next time I work on this code (there is some more refactorings upcoming). For now the updates do no harm since `connection` is the exact same reference all the time and `subscriptions` is unused.

PR #728, opened 2019-09-16
Fix #708 – Stabilize submitExtensionCreateWalletForm

### LINKED ISSUE #708: Fix instable test
```
sil-governance: FAIL src/routes/login/index.e2e.spec.ts (34.957s)
sil-governance:   E2E > Login route
sil-governance:     ✓ should redirect when enqueued login request is accepted (13209ms)
sil-governance:     ✕ should stay in login view if enqueued login request is rejected (8332ms)
sil-governance:     ✓ shows login to IOV extension if not persona detected (5210ms)
sil-governance:     ✓ shows install IOV extension message (4672ms)
sil-governance:   ● E2E > Login route › should stay in login view if enqueued login request is rejected
sil-governance:     Node is either not visible or not an HTMLElement
sil-governance:       23 |   await sleep(1000);
sil-governance:       24 |   const buttons = await page.$$("button");
sil-governance:     > 25 |   await buttons[1].click();
sil-governance:          |   ^
sil-governance:       26 |   await page.click('button[type="submit"]');
sil-governance:       27 |   await whenOnNavigatedToE2eRoute(page, "/wallet");
sil-governance:       28 | };
sil-governance:       at ElementHandle._clickablePoint (../../node_modules/puppeteer/lib/JSHandle.js:199:13)
sil-governance:           at runMicrotasks (<anonymous>)
sil-governance:       at ElementHandle.click (../../node_modules/puppeteer/lib/JSHandle.js:265:20)
sil-governance:       at submitExtensionSignupForm (src/utils/test/persona.ts:25:3)
sil-governance:       at Object.<anonymous> (src/routes/login/index.e2e.spec.ts:85:5)
```

from https://travis-ci.com/iov-one/ponferrada/builds/127180067

webmaster128:
Closes #708

### DIFF
--- a/packages/bierzo-wallet/src/utils/test/persona.ts
+++ b/packages/bierzo-wallet/src/utils/test/persona.ts
@@ -4,6 +4,9 @@ import { sleep } from "ui-logic";
 import { createExtensionPage } from "./e2e";
 import { whenOnNavigatedToE2eRoute } from "./navigation";
 
+const CREATE_WALLET_ID_STEP_1 = "create-wallet-step1";
+const CREATE_WALLET_ID_STEP_2 = "create-wallet-step2";
+const CREATE_WALLET_ID_STEP_3 = "create-wallet-step3";
 const PASSWORD_FIELD = "passwordInputField";
 const PASSWORD_CONFIRM_FIELD = "passwordConfirmInputField";
 const TERMS_ACCEPT_FIELD = "termsAcceptCheckboxField";
@@ -15,15 +18,20 @@ async function clickCreatePersona(page: Page): Promise<void> {
 export const submitExtensionCreateWalletForm = async (page: Page, password: string): Promise<void> => {
   await clickCreatePersona(page);
 
-  await sleep(20); // wait until redirect and render is done
   // Fill the form
+  await page.waitForSelector(`#${CREATE_WALLET_ID_STEP_1}`);
   await page.type(`input[name="${PASSWORD_FIELD}"]`, password);
   await page.type(`input[name="${PASSWORD_CONFIRM_FIELD}"]`, password);
   await page.click(`input[name="${TERMS_ACCEPT_FIELD}"]`);
   await page.click('button[type="submit"]');
-  await sleep(1000);
+
+  // Show recovery words page
+  await page.waitForSelector(`#${CREATE_WALLET_ID_STEP_2}`);
   const buttons = await page.$$("button");
   await buttons[1].click();
+
+  // Security hint page
+  await page.waitForSelector(`#${CREATE_WALLET_ID_STEP_3}`);
   await page.click('button[type="submit"]');
   await whenOnNavigatedToE2eRoute(page, "/wallet");
 };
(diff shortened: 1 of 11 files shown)


packages/bierzo-wallet/src/utils/test/persona.ts
```diff
@@ -21,9 +21,13 @@ export const submitExtensionCreateWalletForm = async (page: Page, password: stri
   await page.type(`input[name="${PASSWORD_CONFIRM_FIELD}"]`, password);
   await page.click(`input[name="${TERMS_ACCEPT_FIELD}"]`);
   await page.click('button[type="submit"]');
-  await sleep(1000);
+  await sleep(1500); // TODO: find direct way to detect if page was change to next step
```
albertandrejev: We can use `waitFor` puppeteer api method (https://github.com/GoogleChrome/puppeteer/blob/v1.20.0/docs/api.md#pagewaitforselectororfunctionortimeout-options-args). We can wait for next step id in this case this is `SECOND_STEP_CREATE_WALLET_ROUTE` from packages/sanes-browser-extension/src/routes/create-wallet/components/ShowPhraseForm.tsx

packages/sil-governance/src/utils/test/persona.ts
```diff
@@ -20,9 +20,13 @@ export const submitExtensionCreateWalletForm = async (page: Page, password: stri
   await page.type(`input[name="${PASSWORD_CONFIRM_FIELD}"]`, password);
   await page.click(`input[name="${TERMS_ACCEPT_FIELD}"]`);
   await page.click('button[type="submit"]');
-  await sleep(1000);
+  await sleep(1500); // TODO: find direct way to detect if page was change to next step
```
albertandrejev: `waitFor` api call.

packages/bierzo-wallet/src/utils/test/persona.ts
webmaster128: Good idea.

So we would do

```ts
await page.waitForSelector(`#${SECOND_STEP_CREATE_WALLET_ROUTE}`);
```

? I was already thinking in the direction but only with routes (https://github.com/iov-one/ponferrada/issues/729). But it would be nice if the routes are not necessary to fix this.

albertandrejev: Yes, you're right, this is exactly what I had in mind. Routes is not required in this case, also there is no route changes between wallet creation steps.

webmaster128: Great, much better than the sleeps as it will save much testing time on average and is more stable in slow situations

albertandrejev: And in this case #729 is not required any more.

webmaster128: I implemented the suggested approach now. Much better! Main commit for the fix is https://github.com/iov-one/ponferrada/pull/728/commits/7975d6971841c0027869007aa7312607a1b1ea5b

PR #741, opened 2019-09-18
Fix #739 – Election rule selection

### LINKED ISSUE #739: Governing Board member cannot change quora or thresholds of other committees
As a Governing Board member I need to be able to change both the quora and thresholds of all the committees.  In the current version I only see the Governing Board as an option.

image

Note only Governing Board members may change threshold or quora of the committees. Members who are NOT Governing Board members cannot change the threshold or quora.

webmaster128:
Closes #739

Calculate selectable rules for amanding threshold/quorum from the currently selected rule. All writeable rules can now be selected.


packages/sil-governance/src/routes/create-proposal/components/FormOptions/AmendCommitteeQuorum.tsx
```diff
@@ -1,52 +1,72 @@
...
-      const ruleItems = electionRules.map(rule => {
+      const allRules = await getAllElectionRules(governor);
+      const writeableRules = allRules.filter(rule => rule.admin === electionRuleAddress);
+      const ruleItems = writeableRules.map(rule => {
         return {
           name: `${rule.id}: ${rule.title}`,
         };
       });
 
       setRuleItems(ruleItems);
```
albertandrejev: This can cause snapshot/DOM tests warnings if you will not implement unmount/cleanup code in `useEffect`

packages/sil-governance/src/routes/create-proposal/components/FormOptions/AmendCommitteeThreshold.tsx
```diff
@@ -9,47 +11,68 @@ import {
...
-      const ruleItems = electionRules.map(rule => {
+      const allRules = await getAllElectionRules(governor);
+      const writeableRules = allRules.filter(rule => rule.admin === electionRuleAddress);
+      const ruleItems = writeableRules.map(rule => {
         return {
           name: `${rule.id}: ${rule.title}`,
         };
       });
 
       setRuleItems(ruleItems);
```
albertandrejev: This too

packages/sil-governance/src/routes/create-proposal/components/ProposalForm.tsx
```diff
@@ -215,7 +229,7 @@ const ProposalForm = (): JSX.Element => {
     updateElectionRules();
```
albertandrejev: line 226: same possible issue.

packages/sil-governance/src/routes/create-proposal/components/FormOptions/AmendCommitteeQuorum.tsx
webmaster128: Thanks for feedback. There are indeed multiple issues with this kind of async data loading, one is that the `setRuleItems(ruleItems)` can can cause an unhandles promise rejection. With the cleanup approach I was also having massive issues in the wallet (don't remember what exactly). The long term solution is probably to avoid async data loading in the UI components entirely and work with redux instead. However, this requires a big refactoring, which is out of scope here.

Issue #656, opened 2019-09-04
Periodically update proposals and votes in the background


webmaster128:
There should be some kind of background process that can update proposals and votes automatically every x seconds or on demand (triggered by user action).

It is not optimal that we have this stuff in the UI component logic.

Any ideas how to implement this @albertandrejev @abefernan?

PR #749, opened 2019-09-24
Feature #656 - Update Proposals Periodically

abefernan:
This PR adds an `updateRequired` field to `ProposalsState` and gets the proposals from the network when it's set to `true` or when 20 seconds have passed since the last update.

### DIFF
--- a/packages/sil-governance/src/routes/dashboard/components/Proposal/VotingPanel/Buttons.tsx
+++ b/packages/sil-governance/src/routes/dashboard/components/Proposal/VotingPanel/Buttons.tsx
@@ -8,7 +8,6 @@ import { Weave } from "ui-logic";
 import { communicationTexts } from "../../../../../communication";
 import { sendSignAndPostRequest } from "../../../../../communication/signandpost";
 import { getBnsConnection } from "../../../../../logic/connection";
-import { getProposals, replaceProposalsAction } from "../../../../../store/proposals";
 import { RootState } from "../../../../../store/reducers";
 import { setTransactionsStateAction } from "../../../../../store/transactions";
 
@@ -48,13 +47,6 @@ const Buttons = ({ id, vote }: Props): JSX.Element => {
         toast.show(communicationTexts.notReadyMessage, ToastVariant.ERROR);
       } else {
         dispatch(setTransactionsStateAction(transactionId));
-
-        setTimeout(() => {
-          getProposals(governor).then(
-            chainProposals => dispatch(replaceProposalsAction(chainProposals)),
-            error => console.error(error),
-          );
-        }, 5000);
       }
     } catch (error) {
       console.error(error);
--- a/packages/sil-governance/src/routes/dashboard/components/Proposal/VotingPanel/Buttons.tsx
+++ b/packages/sil-governance/src/routes/dashboard/components/Proposal/VotingPanel/Buttons.tsx
@@ -8,6 +8,7 @@ import { Weave } from "ui-logic";
 import { communicationTexts } from "../../../../../communication";
 import { sendSignAndPostRequest } from "../../../../../communication/signandpost";
 import { getBnsConnection } from "../../../../../logic/connection";
+import { requireUpdateProposalsAction } from "../../../../../store/proposals";
 import { RootState } from "../../../../../store/reducers";
 import { setTransactionsStateAction } from "../../../../../store/transactions";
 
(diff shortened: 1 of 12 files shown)


packages/sil-governance/src/store/proposals/actions.ts
```diff
@@ -74,3 +79,7 @@ export const replaceProposalsAction = (proposals: ProposalsState): ReplacePropos
   type: "@@proposals/REPLACE",
   payload: proposals,
 });
+
+export const refreshProposalsAction = (): RefreshProposalsActionType => ({
+  type: "@@proposals/REFRESH",
+});
```
webmaster128: I don't think this is an "action" is the sense of redux (https://redux.js.org/basics/actions), as it does not manipulate the store. It is an instruction to potentially (if network works) update the store at some point in the future.

The better way to implement is:
1. have some kind of `updateRequired` property in the store (e.g. at the proposals level)
2. Have an action that sets this property to true if new data should be loaded at next chance
3. In `bootApplication` create an interval that checks if something needs to be updated, reloads data, updates store

packages/sil-governance/src/routes/dashboard/index.tsx
```diff
@@ -28,6 +28,10 @@ const Dashboard = ({ filter }: Props): JSX.Element => {
   const address = governor ? governor.address : ("" as Address);
   const [electorates, setElectorates] = useState<Readonly<Electorate[]>>([]);
 
+  useEffect(() => {
+    dispatch(refreshProposalsAction());
+  }, [dispatch]);
```
webmaster128: Great to see more UI component code becoming synchonous

packages/sil-governance/src/routes/dashboard/components/Proposal/VotingPanel/Buttons.tsx
```diff
@@ -48,13 +47,6 @@ const Buttons = ({ id, vote }: Props): JSX.Element => {
         toast.show(communicationTexts.notReadyMessage, ToastVariant.ERROR);
       } else {
         dispatch(setTransactionsStateAction(transactionId));
-
-        setTimeout(() => {
-          getProposals(governor).then(
-            chainProposals => dispatch(replaceProposalsAction(chainProposals)),
-            error => console.error(error),
-          );
-        }, 5000);
```
webmaster128: this is a good place to set the `needsUpdate` flag (described below) to true as we are expecing changes

packages/sil-governance/src/store/index.ts
```diff
@@ -1,12 +1,59 @@
...
+const refreshProposalsPeriodicallyMw: Middleware<{}, RootState, Dispatch<AnyAction>> = ({
+  dispatch,
+  getState,
+}: MiddlewareAPI): ((next: Dispatch<AnyAction>) => (action: AnyAction) => any) => next => {
+  const refreshProposals = async (): Promise<void> => {
+    await replaceProposals(dispatch, getState());
+    setTimeout(refreshProposals, 3000);
+  };
+
+  refreshProposals();
```
albertandrejev: This will create timeout handler on every update. I am not sure so please correct me if I am wrong.

abefernan: I thought about that but I saw several implementations using it, so not sure.

albertandrejev: ok, but can you check this please?

packages/sil-governance/src/routes/dashboard/components/Proposal/VotingPanel/Buttons.tsx
abefernan: Should it be set with the timeout of 5 seconds or inmediately after `dispatch(setTransactionsStateAction(transactionId));`?

webmaster128: Very good question. We expect a change on the blockchain shortly after running this code. What about setting it twice, immediately and again after 5 seconds

packages/sil-governance/src/store/proposals/actions.ts
abefernan: I'm not sure how (3.) would get along with the periodic update of the proposals. I think it's also needed to get changes from other users.

Should I have a `setInterval` each X seconds that updates the proposals if `updateRequired`, and another `setInterval` each Y seconds that updates the proposals regardless?

webmaster128: What about this: you have an interval that runs very often but is usually cheap, like every 500 ms. Also an additional time stores when the last network request was triggered. 

- if user requested updated, trigger network request
- if last network request is older than 20 seconds, trigger a new one
- otherwise do nothing since the data we have is good enough

abefernan:
> LGTM, nice job. Did you test that a little bit (observing the network monitor) so make sure nother crazy happens by accident?

I tested with `console.log`, I'll check out the network monitor a bit more and merge it if everything is ok.

PR #754, opened 2019-09-24
Feature #752 - Add chain id to the config file

### LINKED ISSUE #752: Add chainId to ChainSpec type
The `interface ChainSpec` should have the same `chainId` property as in extension and governance.

This requires updating the type and the 3 config files development/staging/production

albertandrejev:
Closes #752.

### DIFF
--- a/packages/bierzo-wallet/src/logic/connection.ts
+++ b/packages/bierzo-wallet/src/logic/connection.ts
@@ -11,6 +11,7 @@ let liskConnection: LiskConnection | undefined;
 
 async function getEthereumConnection(url: string, scraperApiUrl?: string): Promise<EthereumConnection> {
   if (!ethereumConnection) {
+    // eslint-disable-next-line require-atomic-updates
     ethereumConnection = await EthereumConnection.establish(url, { scraperApiUrl });
   }
   return ethereumConnection;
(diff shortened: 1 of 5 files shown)


packages/bierzo-wallet/src/logic/connection.ts
```diff
@@ -11,20 +11,23 @@ let liskConnection: LiskConnection | undefined;
...
   if (!bnsConnection) {
+    // eslint-disable-next-line require-atomic-updates
     bnsConnection = await BnsConnection.establish(url);
   }
   return bnsConnection;
 }
 
 async function getLiskConnection(url: string): Promise<LiskConnection> {
   if (!liskConnection) {
+    // eslint-disable-next-line require-atomic-updates
```
webmaster128: 👍 I tried to understand those warnings but gave up after some time. I don't see how this can be problematic in our use case

packages/bierzo-wallet/src/logic/connection.ts
```diff
@@ -52,8 +55,9 @@ export async function getConnectionFor(spec: ChainSpec): Promise<BlockchainConne
 
 export async function getConnectionForChainId(chainId: ChainId): Promise<BlockchainConnection> {
   for (const chain of (await getConfig()).chains) {
-    const connection = await getConnectionFor(chain.chainSpec);
-    if (connection && connection.chainId() === chainId) return connection;
+    if (chain.chainSpec.chainId === chainId) {
+      return await getConnectionFor(chain.chainSpec);
+    }
```
webmaster128: 👍

packages/bierzo-wallet/src/logic/connection.ts
albertandrejev: Me too

PR #755, opened 2019-09-24
Feature #753 - Use expected chainId in chain connectors

### LINKED ISSUE #753: Use expected chainId in chain connectors
After #752, we know the expected chain ID for connections.

In `getConnectionFor(spec: ChainSpec)` we should use `createBnsConnector`, `createLiskConnector`, `createEthereumConnector` with the expected chain ID parameter (see extension)

albertandrejev:
Closes #753.

### DIFF
--- a/packages/bierzo-wallet/src/logic/connection.ts
+++ b/packages/bierzo-wallet/src/logic/connection.ts
@@ -1,42 +1,41 @@
 import { BlockchainConnection, ChainId } from "@iov/bcp";
-import { BnsConnection, createBnsConnector } from "@iov/bns";
-import { EthereumConnection } from "@iov/ethereum";
+import { createBnsConnector } from "@iov/bns";
 import { createEthereumConnector, EthereumConnectionOptions } from "@iov/ethereum";
-import { createLiskConnector, LiskConnection } from "@iov/lisk";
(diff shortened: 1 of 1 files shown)


packages/bierzo-wallet/src/logic/connection.ts
```diff
@@ -1,34 +1,42 @@
...
   return bnsConnection;
 }
 
-async function getLiskConnection(url: string): Promise<LiskConnection> {
+async function getLiskConnection(url: string, chainId: ChainId): Promise<LiskConnection> {
   if (!liskConnection) {
+    const connector = createLiskConnector(url, chainId);
     // eslint-disable-next-line require-atomic-updates
-    liskConnection = await LiskConnection.establish(url);
+    liskConnection = (await connector.establishConnection()) as LiskConnection;
```
webmaster128: I think we can change the return types of all `getLiskConnection`, `getBnsConnection`, `getEthereumConnection` as well as the 3 variables in this file to `BlockchainConnection`. This avoids the needs for all the casting

webmaster128: Yeah, the real issue is that we don't get the specific types from IOV-Core as we should. I created a ticket for that. In the case of this file, we don't need the specific types.

PR #758, opened 2019-09-25
Feature – Use new Decimal type from IOV-Core

webmaster128:
Using JavaScript `number`s to express money amounts is potentially harmful since a `number` is a IEEE 754 float, which cannot express all values we need precisely (e.g. 0.1). To avoid any kind of issues, a new Decimal type was created.

* `parseFigures` and `Figures` removed because they are not needed anymore
* `stringToAmount` and `amountToString` now without any `number` representation in the process
* Support for comma as decimal separator dropped. If we want that as a feature, we should implement it at the input component level in the UI. Right now we don't need it.


webmaster128:
> Nice to have it now!

Let me know if you need additional features on that type. It is very minimal at the moment.

PR #764, opened 2019-09-25
Feature/#737 show ledger address prompt

### LINKED ISSUE #737: Prompt users ledger is not usable for signing registration transaction
I log on to the Ponferrada https://wallet-staging.iov.one/ with ledger but as @webmaster128 told: you cannot register username using **Ledger Nano S**. This information and registration of username should be introduced to users.

albertandrejev:
**Description**
This PR will separate personalized name creation messages on balance page based on rpc endpoint type (extension or ledger). Closes #737.

**Before**
N/A

**After**
username-can-not-reg

### DIFF
--- a/packages/bierzo-wallet/src/communication/rpcEndpoint.ts
+++ b/packages/bierzo-wallet/src/communication/rpcEndpoint.ts
@@ -15,11 +15,14 @@ export type GetIdentitiesResponse = readonly Identity[];
  */
 export type SignAndPostResponse = TransactionId | null;
 
+export type RpcEndpointType = "ledger" | "extension";
+
 export interface RpcEndpoint {
   readonly authorizeGetIdentitiesMessage: string;
   readonly authorizeSignAndPostMessage: string;
   readonly notAvailableMessage: string;
   readonly noMatchingIdentityMessage: string;
+  readonly rpcEndpointType: RpcEndpointType;
 
   /**
    * @returns a response or `undefined` if the endpoint was not available
--- a/packages/bierzo-wallet/src/communication/rpcEndpoint.ts
+++ b/packages/bierzo-wallet/src/communication/rpcEndpoint.ts
@@ -22,7 +22,7 @@ export interface RpcEndpoint {
   readonly authorizeSignAndPostMessage: string;
   readonly notAvailableMessage: string;
   readonly noMatchingIdentityMessage: string;
-  readonly rpcEndpointType: RpcEndpointType;
+  readonly type: RpcEndpointType;
 
   /**
    * @returns a response or `undefined` if the endpoint was not available
--- a/packages/bierzo-wallet/src/routes/balance/components/index.tsx
+++ b/packages/bierzo-wallet/src/routes/balance/components/index.tsx
@@ -5,6 +5,7 @@ import { Block, Hairline, Image, Typography } from "medulas-react-components";
 import React from "react";
 import { amountToString } from "ui-logic";
 
+import { RpcEndpointType } from "../../../communication/rpcEndpoint";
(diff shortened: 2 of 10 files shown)


packages/bierzo-wallet/src/communication/rpcEndpoint.ts
```diff
@@ -15,11 +15,14 @@ export type GetIdentitiesResponse = readonly Identity[];
...
 export type SignAndPostResponse = TransactionId | null;
 
+export type RpcEndpointType = "ledger" | "extension";
+
 export interface RpcEndpoint {
   readonly authorizeGetIdentitiesMessage: string;
   readonly authorizeSignAndPostMessage: string;
   readonly notAvailableMessage: string;
   readonly noMatchingIdentityMessage: string;
+  readonly rpcEndpointType: RpcEndpointType;
```
webmaster128: 👍

packages/bierzo-wallet/src/routes/balance/components/index.tsx
```diff
@@ -15,6 +16,7 @@ interface Props {
   readonly onSendPayment: () => void;
   readonly onReceivePayment: () => void;
   readonly onRegisterUsername: () => void;
+  readonly rpcEndpointType: RpcEndpointType | undefined;
```
webmaster128: Why do we allow undefined? If we want to allow undefined, let's make it optional such that you can create the component without setting the value

packages/bierzo-wallet/src/store/rpcendpoint/selectors.ts
```diff
@@ -0,0 +1,7 @@
+import { RpcEndpointType } from "../../communication/rpcEndpoint";
+import { RootState } from "../reducers";
+
+export const getRpcEndpointType = (state: RootState): RpcEndpointType | undefined => {
+  const rpcEndpointType = state.rpcEndpoint;
+  return rpcEndpointType ? rpcEndpointType.rpcEndpointType : undefined;
```
webmaster128: What about throwing an exception if `state.rpcEndpoint` is unset and return type `RpcEndpointType`?

packages/bierzo-wallet/src/store/rpcendpoint/selectors.ts
```diff
@@ -0,0 +1,7 @@
+import { RpcEndpointType } from "../../communication/rpcEndpoint";
+import { RootState } from "../reducers";
+
+export const getRpcEndpointType = (state: RootState): RpcEndpointType | undefined => {
+  const rpcEndpointType = state.rpcEndpoint;
```
webmaster128: `rpcEndpointType` -> `rpcEndpoint`

packages/bierzo-wallet/src/communication/rpcEndpoint.ts
webmaster128: Can we rename this to just `type`, since it is a property of the RPC endpoint anyways. Then we use it like `rpcEndpoint.type` instead of `rpcEndpoint.rpcEndpointType`.

packages/bierzo-wallet/src/store/rpcendpoint/selectors.ts
albertandrejev: Hmm.. yes we can. Because we are 100% sure that this state should be set before we reach the balance route.

albertandrejev: And in this case we will not need to make `readonly rpcEndpointType: RpcEndpointType | undefined;` optional

webmaster128: > Because we are 100% sure that this state should be set before we reach the balance route.

This is at least the case for using the app. Not sure about tests

PR #765, opened 2019-09-25
Feature #731 - Limits Participation Percentage

### LINKED ISSUE #731: Limit precision of Participation
Bildschirmfoto 2019-09-16 um 18 40 09

abefernan:
Limits the decimals of the Participation percentage to 2, following [this approach](https://stackoverflow.com/a/11832950).

Closes #731.

### DIFF
--- a/packages/sil-governance/src/routes/dashboard/components/Proposal/VotingPanel/ParticipationData.tsx
+++ b/packages/sil-governance/src/routes/dashboard/components/Proposal/VotingPanel/ParticipationData.tsx
@@ -10,7 +10,7 @@ interface Props {
 }
 
 const ParticipationData = ({ quorum, threshold, tally }: Props): JSX.Element => {
-  const participation = (tally.totalVotes / tally.maxVotes) * 100;
+  const participation = Math.round((tally.totalVotes / tally.maxVotes) * 100 * 100) / 100;
 
   return (
     <Block>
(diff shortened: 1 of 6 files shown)


packages/sil-governance/src/routes/dashboard/components/Proposal/VotingPanel/ParticipationData.tsx
```diff
@@ -10,7 +10,7 @@ interface Props {
 }
 
 const ParticipationData = ({ quorum, threshold, tally }: Props): JSX.Element => {
-  const participation = (tally.totalVotes / tally.maxVotes) * 100;
+  const participation = Math.round((tally.totalVotes / tally.maxVotes) * 100 * 100) / 100;
```
webmaster128: What about `participation.toFixed(2)` instead?

abefernan: That works too but doesn't avoid trailing 0s. `78.4` would be converted to `78.40` and `78` to `78.00`

webmaster128: 👍 thanks for checking that.

In this case, can we have a `round(value: number, digits: number): number` function in ui-logic that implements the same logic you use here? This is such a common use case and makes code easier to read.

packages/ui-logic/src/number.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+    expect(round(12.3456, 3)).toBe(12.346);
+    expect(round(12.3456, 4)).toBe(12.3456);
+    expect(round(12.3456, 5)).toBe(12.3456);
+  });
+
+  it("rounds value without trailing zeroes", async () => {
+    expect(round(12, 2)).toBe(12);
+    expect(round(12.3, 2)).toBe(12.3);
+  });
+});
```
webmaster128: nice tests

PR #766, opened 2019-09-25
Feature #282 - Sort Proposals Menu

### LINKED ISSUE #282: List of proposals: implement sorting based on menu selection
After https://github.com/iov-one/ponferrada/issues/281 is implemented, the list of proposals should be filtered depending on the selected main menu item.

As a user I can sort the column on:

* Expiration date
* Creation date
* Vote [ yes/no/abstain ]

https://www.figma.com/file/ffb3NsuSizFxKzwApMxewRgN/Voting?node-id=0%3A1

abefernan:
Closes #282.


packages/sil-governance/src/routes/dashboard/components/ProposalsList.tsx
```diff
@@ -34,14 +38,45 @@ const getFilter = (filterType: ElectionFilter, currentUser: Address | null): Pro
...
+
+const compareById: ProposalsComparator = (proposal1, proposal2): number => proposal1.id - proposal2.id;
+const compareByExpiryDate: ProposalsComparator = (proposal1, proposal2): number => {
+  return proposal1.expiryDate.getTime() - proposal2.expiryDate.getTime();
+};
+const compareByStartDate: ProposalsComparator = (proposal1, proposal2): number => {
+  return proposal1.startDate.getTime() - proposal2.startDate.getTime();
+};
+const compareByVote: ProposalsComparator = (proposal1, proposal2): number => {
+  if (proposal1.vote === undefined) return 1;
```
webmaster128: This gives you `proposal1 > proposal2` and `proposal2 > proposal1` at the same time if `.vote` is undefined in both of them.

packages/sil-governance/src/routes/dashboard/components/ProposalsList.tsx
```diff
@@ -34,14 +38,45 @@ const getFilter = (filterType: ElectionFilter, currentUser: Address | null): Pro
...
+
+  const changeComparator = (selectedItem: SelectFieldItem): void => {
+    switch (selectedItem.name) {
+      case "Expiry Date":
+        setComparator(compareByExpiryDate);
+        break;
+      case "Start Date":
+        setComparator(compareByStartDate);
+        break;
+      case "Vote":
```
webmaster128: Can we have those strings in constants? E.g. using an enum with string values?

packages/sil-governance/src/routes/dashboard/components/ProposalsList.tsx
```diff
@@ -34,14 +48,47 @@ const getFilter = (filterType: ElectionFilter, currentUser: Address | null): Pro
   }
 };
 
-function compareById<E extends { readonly id: number }>(element1: E, element2: E): number {
-  return element1.id - element2.id;
-}
+type ProposalsComparator = (proposal1: ProposalProps, proposal2: ProposalProps) => number;
+
+const compareByIdDescending: ProposalsComparator = (proposal1, proposal2): number => {
+  return proposal1 > proposal2 ? 1 : -1;
```
webmaster128: What is the `>` operator between two `ProposalProps` supposed to do?

packages/sil-governance/src/routes/dashboard/components/ProposalsList.tsx
```diff
@@ -34,14 +48,47 @@ const getFilter = (filterType: ElectionFilter, currentUser: Address | null): Pro
...
+};
+const compareByExpiryDate: ProposalsComparator = (proposal1, proposal2): number => {
+  return proposal1.expiryDate.getTime() - proposal2.expiryDate.getTime();
+};
+const compareByStartDate: ProposalsComparator = (proposal1, proposal2): number => {
+  return proposal1.startDate.getTime() - proposal2.startDate.getTime();
+};
+const compareByVote: ProposalsComparator = (proposal1, proposal2): number => {
+  if (proposal2.vote === undefined) return -1;
+  if (proposal1.vote === undefined) return 1;
```
webmaster128: Same feedback as before: you need to handle the case `proposal1.vote === undefined && proposal2.vote === undefined` explicitely. Otherise `compareByVote(proposal1, proposal2) === -1` and `compareByVote(proposal2, proposal1) === -1` which means `proposal1 < proposal2` and `proposal2 < proposal1` at the same time, which is incorrect and will lead to wrong sorting

packages/sil-governance/src/routes/dashboard/components/ProposalsList.tsx
```diff
@@ -34,14 +48,47 @@ const getFilter = (filterType: ElectionFilter, currentUser: Address | null): Pro
...
+        setComparator(() => compareByExpiryDate);
+        break;
+      case ComparatorLabels.StartDate:
+        setComparator(() => compareByStartDate);
+        break;
+      case ComparatorLabels.Vote:
+        setComparator(() => compareByVote);
+        break;
+      default:
+        setComparator(() => compareByIdDescending);
```
webmaster128: Can't you use simply `setComparator(compareByIdDescending);`? Or will that be interpreted as a state update callback function?

abefernan: This gave me a lot of headache. If you pass the comparator directly it gets executed immediately and therefore the app crashes because the `proposal1` and `proposal2` arguments are `undefined`.

The same happens in `useState<ProposalsComparator>(() => compareByIdDescending);`

webmaster128: I don't think it is because the proposals are undefined. Even if the list of proposals was not ready yet, you get an empty list that can be sorted.

I think what happens is the following: Let look at the initial call `const [comparator, setComparator] = useState<ProposalsComparator>(() => compareByIdDescending);`. `useState` has the following API:

```ts
    /**
     * Returns a stateful value, and a function to update it.
     *
     * @version 16.8.0
     * @see https://reactjs.org/docs/hooks-reference.html#usestate
     */
    function useState<S>(initialState: S | (() => S)): [S, Dispatch<SetStateAction<S>>];
```

So you can either pass in a value or an update function that returns this value (see https://reactjs.org/docs/hooks-reference.html#usestate). At runtime, if the value you pass in is any kind of function, it gets executed and the retuned  value is used: https://github.com/facebook/react/blob/v16.10.0/packages/react-reconciler/src/ReactFiberHooks.js#L816-L818

So generally speaking it is probably not a good idea to store the comparators directly in the state. It would be better to use some kind of sort identifier and then switch over the identifier in

```ts
    .filter(filter)
    .sort(comparator);
```

packages/sil-governance/src/routes/dashboard/components/ProposalsList.tsx
```diff
@@ -34,16 +44,78 @@ const getFilter = (filterType: ElectionFilter, currentUser: Address | null): Pro
...
+      default:
+        setComparatorId(ComparatorId.IdDescending);
+    }
+  };
+
+  const { form } = useForm({ onSubmit: () => {} });
 
   const filter = getFilter(filterType, governor ? governor.address : null);
+
+  let comparator;
```
webmaster128: Can you make this typed: `let comparator: ProposalsComparator`? Thiy may require throwing an exception in the switch default to let TypeScript understand it will always be set.

packages/sil-governance/src/routes/dashboard/components/ProposalsList.tsx
```diff
@@ -34,16 +44,80 @@ const getFilter = (filterType: ElectionFilter, currentUser: Address | null): Pro
...
+      comparator = compareByExpiryDate;
+      break;
+    case ComparatorId.StartDate:
+      comparator = compareByStartDate;
+      break;
+    case ComparatorId.Vote:
+      comparator = compareByVote;
+      break;
+    default:
+      comparator = compareByIdDescending;
```
webmaster128: Can you throw an exception here? This only happens if a new value is added but no switch case is created. In this case we want hard crashed to detect this programming error early.

PR #771, opened 2019-09-27
Fix #769 - Fix Create Proposal transaction start date output

### LINKED ISSUE #769: Proposal start date wrong in transcaction display
Screen Shot 2019-09-26 at 22 33 28

This is because the start date is in seconds, not milliseconds.

In `{new Date(tx.startTime).toLocaleString()}` we need to multiply with 1000. Would be good to cover this by a test

albertandrejev:
Closes #769.

### DIFF
--- a/packages/sanes-browser-extension/src/routes/tx-request/index.dom.spec.ts
+++ b/packages/sanes-browser-extension/src/routes/tx-request/index.dom.spec.ts
@@ -30,14 +30,14 @@ const sendRequests: readonly Request[] = [
   },
 ];
 
-describe("DOM > Feature > Transaction Request", (): void => {
+describe("DOM > Feature > Transaction Request", () => {
   let txRequestDOM: React.Component;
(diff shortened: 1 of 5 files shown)


packages/sanes-browser-extension/src/routes/tx-request/index.dom.spec.ts
```diff
@@ -94,13 +95,39 @@ describe("DOM > Feature > Username Registration Request", (): void => {
...
+    },
+  ];
+
+  let txRequestDOM: React.Component;
+
+  beforeEach(async () => {
+    txRequestDOM = await travelToTXRequest(requests);
+  }, 60000);
+
+  fit("should show proper propsal start date", async (): Promise<void> => {
```
webmaster128: - `fit` -> `it`
- Do we need async here?
- Please skip return types for unnamed test callbacks

PR #778, opened 2019-09-30
Feature #717 - Extract CSV Parser

### LINKED ISSUE #717: Extract CSV parsing into separate file; add unittests
The CSV text to recipients (from `string` to `readonly Recipient[]`) code

```ts
const recipients = reader.result
        // split per new line
        .split(/\r\n|\r|\n/)
        // remove empty lines
        .filter(str => str.length > 0)
        // generate Recipient[]
        .map(recipient => {
          const [address, weight] = recipient.split(",");
          return { address: address as Address, weight: parseInt(weight, 10) };
        });
```

should go into a separate file where it can easily be unittested.

----------

- [ ] Extract CSV parsing into separate file
- [ ] add unittests
- [ ] Ensure address column contains address
- [ ] Ensure weight column contains an integer

abefernan:
Closes #717.

### DIFF
--- a/packages/ui-logic/src/csv.spec.ts
+++ b/packages/ui-logic/src/csv.spec.ts
@@ -26,9 +26,9 @@ describe("parseRecipients", () => {
 
     expect(recipients.length).toBe(2);
     expect(recipients[0].address).toBe("tiov15nuhg3l8ma2mdmcdvgy7hme20v3xy5mkxcezea");
-    expect(recipients[0].weight).toBe("4");
+    expect(recipients[0].weight).toBe(4);
     expect(recipients[1].address).toBe("tiov12shyht3pvvacvyee36w5844jkfh5s0mf4gszp9");
-    expect(recipients[1].weight).toBe("5");
+    expect(recipients[1].weight).toBe(5);
   });
 
   it("ignores empty lines", async () => {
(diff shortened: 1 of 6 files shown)


packages/ui-logic/src/csv.spec.ts
```diff
@@ -0,0 +1,71 @@
+import { parseRecipients } from "./csv";
+
+const validText = `
+tiov15nuhg3l8ma2mdmcdvgy7hme20v3xy5mkxcezea,4
```
webmaster128: The initial line break should not be valid. You need to start with 

```
`tiov15nuhg3l8ma2mdmcdvgy7hme20v3xy5mkxcezea,4
```

in the first line

packages/ui-logic/src/csv.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+`;
+
+const invalidAddressText = "invalidAddress,4";
+
+const noNumberWeightText = "tiov15nuhg3l8ma2mdmcdvgy7hme20v3xy5mkxcezea,invalidWeight";
+const noIntegerWeightText = "tiov15nuhg3l8ma2mdmcdvgy7hme20v3xy5mkxcezea,4.2";
+const zeroWeightText = "tiov15nuhg3l8ma2mdmcdvgy7hme20v3xy5mkxcezea,0";
+
+describe("parseRecipients", () => {
+  it("outputs a Recipient array if input valid", async () => {
```
webmaster128: All those test do not need to be `async`

packages/ui-logic/src/csv.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    expect(recipients.length).toBe(2);
+    expect(recipients[0].address).toBe("tiov15nuhg3l8ma2mdmcdvgy7hme20v3xy5mkxcezea");
+    expect(recipients[0].weight).toBe(4);
+    expect(recipients[1].address).toBe("tiov12shyht3pvvacvyee36w5844jkfh5s0mf4gszp9");
+    expect(recipients[1].weight).toBe(5);
+  });
+
+  it("throws error if address not valid", async () => {
+    try {
+      parseRecipients(invalidAddressText);
```
webmaster128: The code you have right now will pass if `parseRecipients` does not throw, even if an exception is expected.

You typically do:

```ts
expect(() => parseRecipients(invalidAddressText)).toThrowError(/Address not valid: invalidAddress/)
```

Same for the error tests below

packages/ui-logic/src/csv.spec.ts
webmaster128: Forget about this one. Our parser allows empty lines. Not strictly CSV but should do harm

PR #780, opened 2019-09-30
Add ReleaseGuaranteeFunds story 

webmaster128:
This adds an initial story for the ReleaseGuaranteeFunds component


packages/sil-governance/src/routes/create-proposal/components/FormOptions/ReleaseGuaranteeFunds.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import DecoratedStorybook, { governanceRoot } from "../../../../utils/storybook";
+import ReleaseGuaranteeFunds from "./ReleaseGuaranteeFunds";
+
+const initialTickers = ["BASH" as TokenTicker, "CASH" as TokenTicker];
+
+storiesOf(governanceRoot, module)
+  .addParameters({ viewport: { defaultViewport: "responsive" } })
+  .add("ReleaseGuaranteeFunds", () => {
+    const { form } = useForm({ onSubmit: () => 0 });
```
abefernan: What does `onSubmit: () => 0` do? I've used `onSubmit: () => {}`. Is there any difference?

webmaster128: No, it is exactly the same in this case. Your version returns `void` due to a missing return value. My version returns `0`. However, the return value is ignored anyways, so it does not matter.

PR #786, opened 2019-09-30
Fix# 774 - Fixes File Still Required when Correct

### LINKED ISSUE #774: Distribute funds: file is still required after CSV was processed
grafik

abefernan:
Closes #774.

### DIFF
--- a/packages/sil-governance/src/routes/create-proposal/components/FormOptions/DistributeFunds/index.tsx
+++ b/packages/sil-governance/src/routes/create-proposal/components/FormOptions/DistributeFunds/index.tsx
@@ -13,7 +13,8 @@ import { parseRecipients } from "ui-logic";
 
 import RecipientsTable from "./RecipientsTable";
 
-const IMPORT_FIELD = "Import recipients";
+const IMPORT_FIELD_NAME = "Import recipients";
+const IMPORT_FIELD_ID = "recipients-file";
 
 export interface Recipient {
   address: Address;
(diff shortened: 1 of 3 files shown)


packages/sil-governance/src/routes/create-proposal/components/FormOptions/DistributeFunds/index.tsx
```diff
@@ -59,6 +60,10 @@ const DistributeFunds = ({ form, recipientsChanged }: Props): JSX.Element => {
         setRecipients(recipients);
         recipientsChanged(recipients);
       } catch (error) {
+        // TODO the value of an input file can only be set to empty string, not working
+        (document.getElementById(IMPORT_FIELD_ID) as HTMLInputElement).value = "";
```
webmaster128: The issue with this approach is that it only changes the rendered html component, not the state that remains unchanged in the React component. Let's better not access DOM directly like this

packages/sil-governance/src/routes/create-proposal/components/FormOptions/DistributeFunds/index.tsx
```diff
@@ -73,16 +76,17 @@ const DistributeFunds = ({ form, recipientsChanged }: Props): JSX.Element => {
   return (
     <React.Fragment>
       <Block marginTop={2} display="flex" alignItems="center">
-        <Typography>{IMPORT_FIELD}</Typography>
+        <Typography>{IMPORT_FIELD_NAME}</Typography>
         <Block marginLeft={2}>
           <TextField
-            name={IMPORT_FIELD}
+            name={IMPORT_FIELD_NAME}
+            id={IMPORT_FIELD_ID}
```
webmaster128: Can we remove this ID as well?

PR #798, opened 2019-10-01
Feature #773 - Electorate Members Page

### LINKED ISSUE #773: Create page for electorates members
A simple page for each electorate with a table that lists the addresses and weights of the electorate members.

I think we can link to the pages from the "Member of ..." section (one link per committee)

abefernan:
Closes #773.

**Screenshot**
imagen

### DIFF
--- a/packages/sil-governance/src/components/Header/index.tsx
+++ b/packages/sil-governance/src/components/Header/index.tsx
@@ -1,12 +1,21 @@
 import { Address } from "@iov/bcp";
 import { Electorate } from "@iov/bns";
 import { Block, CircleImage, Image, Typography } from "medulas-react-components";
-import React from "react";
+import React, { useMemo } from "react";
 
 import iovLogo from "../../assets/iov-logo-title.svg";
 import userIcon from "../../assets/user.svg";
+import { history } from "../../routes";
+import { SHOW_ELECTORATE_ROUTE } from "../../routes/paths";
 
 export const HEADER_HTML_ID = "header";
+
+const navigateToElectorate = (id: number): void => {
+  const target = `${SHOW_ELECTORATE_ROUTE}/${id}`;
+  if (history.location.pathname !== target) {
+    history.push(target);
+  }
+};
 
 interface Props {
   readonly address: Address;
@@ -14,6 +23,26 @@ interface Props {
 }
 
 const Header = ({ address, electorates }: Props): JSX.Element => {
+  const ElectorateLinks = useMemo(
+    () => (): JSX.Element => (
+      <React.Fragment>
+        {electorates.map((e, index) => (
+          <Block display="inline" key={e.id}>
+            <Typography inline link variant="body2" onClick={() => navigateToElectorate(e.id)}>
+              {e.title}
+            </Typography>
+            {index !== 1 && (
+              <Typography inline variant="body2">
+                ,{" "}
+              </Typography>
+            )}
+          </Block>
+        ))}
+      </React.Fragment>
+    ),
+    [electorates],
+  );
+
   return (
     <Block id={HEADER_HTML_ID} width="100%" minHeight="78px" display="flex" alignItems="center">
       <Block minWidth="205px" display="flex" alignItems="center" justifyContent="center">
@@ -24,7 +53,10 @@ const Header = ({ address, electorates }: Props): JSX.Element => {
         <Block marginRight={4} display="flex" alignItems="center">
           <Block textAlign="right">
             <Typography variant="body2">{address}</Typography>
-            <Typography variant="body2">Member of {electorates.map(e => e.title).join(", ")}</Typography>
+            <Typography inline variant="body2">
+              Member of{" "}
+            </Typography>
+            <ElectorateLinks />
           </Block>
           <Block marginLeft={1}>
             <CircleImage alt="Logo" icon={userIcon} dia="32px" circleColor="#fff" />
--- a/packages/sil-governance/src/components/Header/index.tsx
+++ b/packages/sil-governance/src/components/Header/index.tsx
@@ -1,21 +1,27 @@
 import { Address } from "@iov/bcp";
 import { Electorate } from "@iov/bns";
-import { Block, CircleImage, Image, Typography } from "medulas-react-components";
+import { Block, CircleImage, Image, makeStyles, Typography } from "medulas-react-components";
 import React, { useMemo } from "react";
+import { Link } from "react-router-dom";
 
 import iovLogo from "../../assets/iov-logo-title.svg";
 import userIcon from "../../assets/user.svg";
-import { history } from "../../routes";
 import { SHOW_ELECTORATE_ROUTE } from "../../routes/paths";
 
 export const HEADER_HTML_ID = "header";
 
-const navigateToElectorate = (id: number): void => {
-  const target = `${SHOW_ELECTORATE_ROUTE}/${id}`;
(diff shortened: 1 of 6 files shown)


packages/sil-governance/src/components/Header/index.tsx
```diff
@@ -1,19 +1,48 @@
...
   readonly electorates: readonly Electorate[];
 }
 
 const Header = ({ address, electorates }: Props): JSX.Element => {
+  const ElectorateLinks = useMemo(
+    () => (): JSX.Element => (
+      <React.Fragment>
+        {electorates.map((e, index) => (
+          <Block display="inline" key={e.id}>
+            <Typography inline link variant="body2" onClick={() => navigateToElectorate(e.id)}>
```
webmaster128: By default, a Link should be created using the `<Link` component, which can be combinded with Typography like

```tsx
    <Link to="https://iov.one">
      <Typography weight="light" link color="primary">
        Go to IOV
      </Typography>
    </Link>
```

or

```tsx
          <Typography link={!!blockExplorerUrl} inline>
            {blockExplorerUrl ? (
              <Link to={blockExplorerUrl}>{`${recipientShort}`}</Link>
            ) : (
              `${recipientShort}`
            )}
          </Typography>
```

this avoids all the manual history manipulation, which should be covered by the router

packages/sil-governance/src/components/Header/index.tsx
```diff
@@ -1,19 +1,48 @@
...
+              {e.title}
+            </Typography>
+            {index !== 1 && (
+              <Typography inline variant="body2">
+                ,{" "}
+              </Typography>
+            )}
```
webmaster128: This is placed behind link 0, 2, 3, 4, 5, and so on. It only leads to the correct result for 2 links. What you want is to prepand this before every link other than 0.

packages/sil-governance/src/components/Header/index.tsx
```diff
@@ -1,19 +1,48 @@
...
   readonly address: Address;
   readonly electorates: readonly Electorate[];
 }
 
 const Header = ({ address, electorates }: Props): JSX.Element => {
+  const ElectorateLinks = useMemo(
+    () => (): JSX.Element => (
+      <React.Fragment>
+        {electorates.map((e, index) => (
+          <Block display="inline" key={e.id}>
```
webmaster128: You can archive this with CSS as well (see the very first example in the question of https://stackoverflow.com/questions/27836769/comma-separated-spans-with-pure-css). With CSS, you only need one JSX element per link, allowing you to also drop the container `<Block`

packages/sil-governance/src/routes/show-electorate/components/ElectorateMembers.tsx
```diff
@@ -0,0 +1,45 @@
...
+            </td>
+          </tr>
+        ))}
+      </tbody>
+    </table>
+  );
+
+  return (
+    <Block flexGrow={1} margin={2}>
+      <Typography>{electorate.title} members:</Typography>
```
webmaster128: Get this get a little bit of headline style?

packages/sil-governance/src/components/Header/index.tsx
abefernan: I'm getting `Uncaught Error: Invariant failed: You should not use <Link> outside a <Router>` when using a partial url as the `to` parameter. So I can set `to` to `https://www.google.com` but I can't set it to `/create-proposal` or whatever.

However in Neuma i can see the following code in `UnlockControls.tsx` without issue:
```typescript
<Link to={RESTORE_WALLET}>
    <Typography variant="subtitle2" color="primary" link inline>
        Restore wallet
    </Typography>
</Link>
```

And in `paths.ts`:
```typescript
export const RESTORE_WALLET = "/restore-wallet";
```

webmaster128: Where exactly do you get that exception? In the application or the storybook? The sil application has a `<Router` in the `<Routes />` component. However, the storybook misses a `<Router`, which might be a problem. We might need a solution like `SanesStorybook` for the governance app as well.

abefernan: The app crashed as soon as it tried to render a `<Link>`. I've solved this for now by using `react-router-dom`'s `<Link>` instead of medulas'.

packages/sil-governance/src/components/Header/index.tsx
```diff
@@ -24,7 +60,14 @@ const Header = ({ address, electorates }: Props): JSX.Element => {
...
             <Typography variant="body2">{address}</Typography>
-            <Typography variant="body2">Member of {electorates.map(e => e.title).join(", ")}</Typography>
+            {hasElectorates && (
+              <React.Fragment>
+                <Typography inline variant="body2">
+                  Member of{" "}
+                </Typography>
+                <ElectorateLinks />
+              </React.Fragment>
+            )}
```
webmaster128: May we have a text for the other case, like "Not a member of any electorate"?

packages/sil-governance/src/components/Header/index.tsx
webmaster128: Hard to tell without the stack tract. Maybe https://github.com/iov-one/ponferrada/pull/809 helps. Anyways, there is nothing wrong with using `<Link>` from `react-router-dom` directly

packages/sil-governance/src/routes/show-electorate/components/ElectorateMembers.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { Block, Typography } from "medulas-react-components";
+import React from "react";
+
+interface Props {
+  readonly electorate: Electorate;
+}
+
+const ElectorateMembers = ({ electorate }: Props): JSX.Element => {
+  const MembersTable = (): JSX.Element => (
+    <table>
```
albertandrejev: Please use MUI [table component](https://material-ui.com/components/tables/) instead.

packages/sil-governance/src/routes/show-electorate/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+        setElectorates(electorates);
+      }
+    };
+    updateElectorates();
+  }, [governor]);
+
+  useEffect(() => {
+    const electorate = electorates.find(e => e.id === parseInt(match.params.electorateId, 10));
+    setElectorate(electorate);
+  }, [electorates, match.params.electorateId]);
```
albertandrejev: You can use `useMemo` instead of `useEffect` this will save you at least one component redraw cycle.

albertandrejev:
Am I right you want to move changes I talked about to another [issue](https://github.com/iov-one/ponferrada/issues/811)? 
1. Use MUI Table
2. Replace `useEffect` with `useMemo`?
If yes, then I will approve this PR.

packages/sil-governance/src/components/Header/index.tsx
```diff
@@ -1,19 +1,55 @@
...
+      return (
+        <Block display="inline" className={classes.commaSeparator}>
+          {electorates.map(e => (
+            <Typography inline variant="body2" key={e.id}>
+              <Link to={`${SHOW_ELECTORATE_ROUTE}/${e.id}`} className={classes.electorateLink}>
+                {e.title}
+              </Link>
+            </Typography>
+          ))}
+        </Block>
```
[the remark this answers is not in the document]
webmaster128: Nice we got this working with CSS

PR #799, opened 2019-10-01
Feature #742 - Connect to requested networks only

### LINKED ISSUE #742: Do not connect to Lisk and Ethereum when using Ledger
When using the wallet with a Ledger, we can only use BNS anyways

### LINKED ISSUE #744: Remove other currencies from Payment route in case of Ledger usage.
Only native BNS currency should be left. Based on #742.

albertandrejev:
Closes #742.
Closes #744.

### DIFF
--- a/packages/bierzo-wallet/src/logic/connection.ts
+++ b/packages/bierzo-wallet/src/logic/connection.ts
@@ -7,7 +7,7 @@ import { ChainSpec, getConfig } from "../config";
 
 const connections = new Map<ChainId, BlockchainConnection>();
 
-export async function setEthereumConnection(
+async function establishEthereumConnection(
   url: string,
   chainId: ChainId,
   options: EthereumConnectionOptions,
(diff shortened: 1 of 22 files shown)


packages/bierzo-wallet/src/logic/connection.ts
```diff
@@ -5,39 +5,38 @@ import { createLiskConnector } from "@iov/lisk";
...
 import { ChainSpec, getConfig } from "../config";
 
-let ethereumConnection: BlockchainConnection | undefined;
-let bnsConnection: BlockchainConnection | undefined;
-let liskConnection: BlockchainConnection | undefined;
+export interface ChainConnections {
+  [chainId: string]: BlockchainConnection;
+}
+
+const connections: ChainConnections = {};
```
webmaster128: Can we make this a `Map<string, BlockchainConnection>`? This gives us much better type safety for the getters

PR #804, opened 2019-10-03
Fix #785 - Not stable E2E test

### LINKED ISSUE #785: Test "should create personalized address" not stable
```
bierzo-wallet: FAIL src/routes/balance/index.e2e.spec.ts (49.227s)
bierzo-wallet:   ● E2E > Balance route › should create personalized address
bierzo-wallet:     expect(received).toBe(expected) // Object.is equality
bierzo-wallet:     Expected: "gocsymzwib*iov"
bierzo-wallet:     Received: "Get your human readable"
bierzo-wallet:       78 | 
bierzo-wallet:       79 |     const personaAddress = await getUsernameE2E(await page.$$("h5"));
bierzo-wallet:     > 80 |     expect(personaAddress).toBe(username);
bierzo-wallet:          |                            ^
bierzo-wallet:       81 |   }, 45000);
bierzo-wallet:       82 | });
bierzo-wallet:       83 | 
bierzo-wallet:       at Object.toBe (src/routes/balance/index.e2e.spec.ts:80:28)
```

See 
- https://travis-ci.com/iov-one/ponferrada/builds/129692510
- https://travis-ci.com/iov-one/ponferrada/builds/129695707

albertandrejev:
Unfortunately I can not replicate this error locally, but hope this fix will improve situation.

Closes #785.

### DIFF
--- a/packages/bierzo-wallet/src/routes/balance/index.e2e.spec.ts
+++ b/packages/bierzo-wallet/src/routes/balance/index.e2e.spec.ts
@@ -8,7 +8,12 @@ import { acceptEnqueuedRequest } from "../../utils/test/persona";
 import { withChainsDescribe } from "../../utils/test/testExecutor";
 import { REGISTER_PERSONALIZED_ADDRESS_ROUTE } from "../paths";
 import { REGISTER_USERNAME_FIELD } from "../registerName/components";
-import { getBalanceTextAtIndex, getUsernameE2E, waitForAllBalances } from "./test/operateBalances";
+import {
+  getBalanceTextAtIndex,
+  getUsernameE2E,
+  waitForAllBalances,
+  waitForUsername,
+} from "./test/operateBalances";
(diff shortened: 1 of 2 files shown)


packages/bierzo-wallet/src/routes/balance/index.e2e.spec.ts
```diff
@@ -76,6 +76,7 @@ withChainsDescribe("E2E > Balance route", () => {
     const buttons = await page.$$("button");
     await buttons[2].click();
 
+    await sleep(1500);
```
webmaster128: I'm not a big fan of adding sleeps that are not needed most of the time, as this will increase the test runtime more and more. Instead, we should try to use `whenTrue` if possible, that will resolve very quickly if the test condition is met.

In this case we can build something like

```ts
await whenTrue(async () => (await getUsernameE2E(await page.$$("h5"))).matches(/\*iov$/), 5000);
```

which we can also wrap in a helper similar to `waitForAllBalances`

PR #808, opened 2019-10-07
Feature - Material UI package upgrade

albertandrejev:
**Description**
This PR will upgrade Material UI package to 4.5.0 and move packages into common location.


packages/medulas-react-components/package.json
```diff
@@ -19,9 +19,6 @@
     "not op_mini all"
   ],
   "dependencies": {
-    "@material-ui/core": "^4.3.1",
-    "@material-ui/icons": "^4.2.1",
-    "@material-ui/styles": "^4.3.0",
```
webmaster128: There are cases where CRA needs runtime libraries as `dependencies` (I don't remember exactly which ones that are). This is why all dependencies that the user runs should remain in `dependencies`. `devDependencies` is for stuff that is solely for developers like testing and and deployment tools.

Since we cannot place `dependencies` in the root package.json, we unfortunately cannot deduplicate those libraries reliably. What I usually do to quickly update them is project-wide search in VS Code (cmd+Shirft+f)

PR #824, opened 2019-10-14
Feature #812 - Generic Error Parser

### LINKED ISSUE #812: Implement Generic Error Parser
Remove the non-JSON error support from `ui-logic/Weave`'s `tryParser`.

Implement this non-JSON error support in a Generic Error Parser that also wraps Weave Error Parser, as suggested [here](https://github.com/iov-one/ponferrada/pull/805#discussion_r332325342).

abefernan:
Closes #812.


packages/ui-logic/src/error.ts
```diff
@@ -0,0 +1,30 @@
+import { Weave } from "./weave";
```
albertandrejev: you have circular dependency here.

packages/ui-logic/src/error.ts
```diff
@@ -0,0 +1,30 @@
...
+    return normalizedErrorMessage;
+  }
+  /**
+   * Tries to parse error message as a Weave error.
+   * On success, returns an error string that can be shown to the user.
+   * On failure, returns the error if it's a non empty string, otherwise retuns null.
+   *
+   * @param error pass in the variable from the catch block. Error and string are supported for parsing.
+   */
+  public static tryParseError(error: unknown): string | null {
```
albertandrejev: This method is not used anywhere across the project. Is it really necessary to have this method?

abefernan: Yes, I forgot to commit the usage.

packages/ui-logic/src/error.ts
abefernan: I'm not sure if I should include `weave.ts`'s code into `error.ts` so that it's not circular.

abefernan: The thing is `Weave`'s class may do other things besides parsing errors in the future.

albertandrejev: But you can move `ErrorParser.tryParseError` to the `wave.ts`. With different name of course.

abefernan: `ErrorParser.tryParseError` is a decorator of `Weave.tryParseError`, it parses other kind of errors besides Weave's ones. That's why I thought it may not fit in `weave.ts`.

albertandrejev: we need to avoid circular dependency somehow I think.

abefernan: I agree. How about having an `ErrorParser` class with `tryParseWeaveError` and `tryParseError`, and removing `weave.ts` since right now it does not do anything besides parsing weave errors?

albertandrejev: I think it is ok.

packages/ui-logic/src/weave.ts
```diff
@@ -6,12 +8,10 @@ export class Weave {
    * @param error pass in the variable from the catch block. Error and string are supported for parsing.
    */
   public static tryParseError(error: unknown): string | null {
-    const errorMessage = error instanceof Error ? error.message : typeof error === "string" ? error : "";
-
-    const normalizedErrorMessage = errorMessage.trim().replace(/^Error:\s*/, "");
+    const errorMessage = ErrorParser.normalizeErrorMessage(error);
```
albertandrejev: Now you do not need `Wave.tryParseError` method any more because now there is not references to this method except from `ErrorParser.tryParseError` method. So maybe you will implement all code from `ErrorParser.tryParseError`  into `Wave.tryParseError`?

abefernan: The purpose of this PR originates in [this suggestion](https://github.com/iov-one/ponferrada/pull/805#discussion_r332325342).

albertandrejev: Ok. But do you think we will use this method anywhere?

abefernan: Yes because I made a mistake and the Buttons component should only use Weave's. So i think [this approach](https://github.com/iov-one/ponferrada/pull/824#discussion_r334378250) would be good.

Issue #790, opened 2019-10-01
Authorize screen design modifications


annatorok:
Instead a single sentence in the modal, we should provide a more visual way for the user to know that they need to authorize the request in the extension.

Changes:

- [x] Use visual as seen in the design (browser bar design + arrow)
- [x] Under visual, new text: "Please authorize request in Neuma Browser Extension to continue." 
- [x] Under new text, new description should be: "Use the Neuma Browser Extension located in the Browser menu in order to authorize your request."

Questions:
- [x] Andrei - Direction of arrow is a bit off - should be tested if the design is appropriate here

<img width="1440" alt="Wallet - authorize (1)" src="https://user-images.githubusercontent.com/7419692/66117206-64e3f800-e5d4-11e9-9fd5-aff4be9062ec.png">

<img width="1440" alt="Wallet - authorize using Ledger" src="https://user-images.githubusercontent.com/3502260/66565315-d985dc00-eb6a-11e9-8126-5279b96f79d9.png">

annatorok:
Answer:
- [ ] New design is available, if having the modal moved to this upper position doesn't make sense, just leave it in the middle.

PR #827, opened 2019-10-14
Feature #790 - Neuma request billboard

albertandrejev:
**Description**
This PR will add new component to the billboard to ask for the action to be performed on the extension. Part of the #790.

neuma-billboard


packages/bierzo-wallet/src/components/BillboardMessage/index.stories.tsx
```diff
@@ -0,0 +1,16 @@
+import { storiesOf } from "@storybook/react";
+import { Block } from "medulas-react-components";
+import * as React from "react";
+
+import DecoratedStorybook, { bierzoRoot } from "../../utils/storybook";
+import NeumaBillboardMessage from "./NeumaBillboardMessage";
+
+storiesOf(`${bierzoRoot}/BillboardMessage`, module)
+  .addParameters({ viewport: { defaultViewport: "responsive" } })
+  .add("Nauma", () => (
```
abefernan: Neuma!

albertandrejev:
I see only one main reason to use arrow function in components, because of access to `this`. So I think it is better to use `function` in components definitions. There is lots of resources about this discussion: `=>` vs `function`.
* https://www.freecodecamp.org/news/when-and-why-you-should-use-es6-arrow-functions-and-when-you-shouldnt-3d851d7f0b26/
* https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1

PR #842, opened 2019-10-23
Several DOM Tests Fixes

### LINKED ISSUE #818: Create Proposal route: DOM Tests

### LINKED ISSUE #833: Dashboard route: Fixes for DOM Tests
The following issues were found when implementing DOM tests for the `Dashboard` route (#813) and for the `Create Proposal` route (#818):

- The `Add New Proposal` button cannot be tested. The `click()` method found in `src/utils/test/dom.tsx` causes the following error:
```typescript
var evt = document.createEvent('Event');
TypeError: Cannot read property 'createEvent' of null
```
A number of solutions have been tested, like changing the `SyntheticEventData` of the click event, with no success.
- The dropdown for sorting proposals cannot be tested. The dropdown menu with the sorting options is made with Material UI's Popper, which relies in the 3rd party library Popper.js.
  - The first solution that was tried was to click on the button and then on the desired option, but Popper.js seemingly uses some browser APIs that makes Popper not to work as expected in DOM tests and its options are unclickable.
  - The second solution was to change the value of the button (a MUI `InputBase` component) simulating its `onInput` event (`onChange` is not an option since it's overriden by `final-form`). But the value of the input field cannot be changed programmatically, it instantly changes back to its old value.

Both cases may be covered by E2E tests.

abefernan:
Closes #833.
Closes #818.

The `"An update inside a test was not wrapped in act(...)."` warning still persists after improving `useEffect` with cleanup functions.

### DIFF
--- a/packages/sil-governance/src/routes/create-proposal/index.tsx
+++ b/packages/sil-governance/src/routes/create-proposal/index.tsx
@@ -25,12 +25,13 @@ const CreateProposal = (): JSX.Element => {
 
   useEffect(() => {
     let isSubscribed = true;
-
     const updateElectorates = async (): Promise<void> => {
       // in DOM tests, governor is not set
-      if (governor && isSubscribed) {
+      if (governor) {
         const electorates = await governor.getElectorates();
-        setElectorates(electorates);
+        if (isSubscribed) {
+          setElectorates(electorates);
+        }
       }
     };
     updateElectorates();
--- a/packages/sil-governance/src/routes/create-proposal/index.tsx
+++ b/packages/sil-governance/src/routes/create-proposal/index.tsx
@@ -24,14 +24,20 @@ const CreateProposal = (): JSX.Element => {
   const [electorates, setElectorates] = useState<readonly Electorate[]>([]);
 
   useEffect(() => {
+    let isSubscribed = true;
+
     const updateElectorates = async (): Promise<void> => {
       // in DOM tests, governor is not set
-      if (governor) {
+      if (governor && isSubscribed) {
         const electorates = await governor.getElectorates();
         setElectorates(electorates);
       }
     };
     updateElectorates();
+
+    return () => {
+      isSubscribed = false;
+    };
   }, [governor]);
 
   const onReturnToDashboard = (): void => {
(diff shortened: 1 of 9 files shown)


packages/sil-governance/src/routes/create-proposal/index.tsx
```diff
@@ -24,14 +24,20 @@ const CreateProposal = (): JSX.Element => {
...
 
   useEffect(() => {
+    let isSubscribed = true;
+
     const updateElectorates = async (): Promise<void> => {
       // in DOM tests, governor is not set
-      if (governor) {
+      if (governor && isSubscribed) {
         const electorates = await governor.getElectorates();
         setElectorates(electorates);
```
albertandrejev: No, you should wrap `setElectorates(electorates)` inside `isSubscribed` condition, because you have `async` operation inside current block:
```ts
if (governor) {
  const electorates = await governor.getElectorates();
  if (isSubscribed) {
    setElectorates(electorates);
  }
}
```

packages/sil-governance/src/routes/dashboard/index.tsx
```diff
@@ -33,14 +33,20 @@ const Dashboard = ({ filter }: Props): JSX.Element => {
...
   useEffect(() => {
+    let isSubscribed = true;
+
     const updateElectorates = async (): Promise<void> => {
       // in DOM tests, governor is not set
-      if (governor) {
+      if (governor && isSubscribed) {
         const electorates = await governor.getElectorates();
         setElectorates(electorates);
       }
```
albertandrejev: Same here

packages/sil-governance/src/routes/show-electorate/index.tsx
```diff
@@ -30,14 +30,20 @@ const ShowElectorate = ({ match }: RouteComponentProps<Params>): JSX.Element =>
...
   useEffect(() => {
+    let isSubscribed = true;
+
     const updateElectorates = async (): Promise<void> => {
       // in DOM tests, governor is not set
-      if (governor) {
+      if (governor && isSubscribed) {
         const electorates = await governor.getElectorates();
         setElectorates(electorates);
       }
```
albertandrejev: And here. The idea is to avoid calling state change after all async operations. Because those async operations can be finished after component destruction.

packages/sil-governance/src/routes/create-proposal/index.tsx
abefernan: I already had tried this but both approaches give the same warning. I have taken a look at the show-electorate route in order to isolate it, which has very few lines of code and produces the warning and I don't know what may be wrong.

PR #854, opened 2019-10-29
Feature #853 - Manifest update

### LINKED ISSUE #853: Redirect wallet.iov.one to wallet.neuma.io
**Goal:** we will use the Neuma brand as the name/face for the wallet/extension from now on and we want to use it as THE brand.

We want to therefore use **wallet.neuma.io** instead of wallet.iov.one.

**What to change?**
https://github.com/iov-one/ponferrada/blob/master/packages/sanes-browser-extension/config/manifest-staging.json's `externally_connectable.matches` needs to replace `https://*.iov.one` with `https://*.neuma.io`

albertandrejev:
Closes #853.

### DIFF
--- a/packages/sanes-browser-extension/config/manifest-production.json
+++ b/packages/sanes-browser-extension/config/manifest-production.json
@@ -19,7 +19,7 @@
   },
   "content_scripts": [],
   "externally_connectable": {
-    "matches": ["http://localhost/*", "https://*.iov.one/*"]
+    "matches": ["http://localhost/*", "https://*.neuma.io/*"]
   },
   "background": {
     "scripts": ["js/backgroundScript.js"],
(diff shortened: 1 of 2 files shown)


packages/sanes-browser-extension/config/manifest-production.json
```diff
@@ -19,7 +19,7 @@
   },
   "content_scripts": [],
   "externally_connectable": {
-    "matches": ["http://localhost/*", "https://*.iov.one/*"]
+    "matches": ["http://localhost/*", "https://*.neuma.io/*"]
```
webmaster128: Removing `https://*.iov.one/*` from this whitelist disallows `governance.iov.one` to send messages to the browser extension, breaking the governance app

PR #887, opened 2019-11-26
Fix #872 - Login Billboard blinking

### LINKED ISSUE #872: Fix login billboard message in case if absent extensions
Right now billboard message appears and disappears very quickly (blink) in case if extension is not reachable.

albertandrejev:
Closes #872.


webmaster128:
Nice solution! I suggest making the delay argument optional and default to 0, but this is nice to have for later

PR #903, opened 2019-12-04
Fix #871 - Do not show fee for incoming TX

### LINKED ISSUE #871: The transaction fee should only appear in the transaction for the person paying
Currently, we show the transaction fee detail for every transaction, be it receiving tokens or sending tokens.

When receiving tokens, it just doesn't make sense to display the transaction fee as I only got tokens, therefore knowing cost doesn't have any value for me.

**Goal:**
To hide the transaction fee field in case the user receives tokens

(The goal of the transaction detail is that if I add up all the numbers + and - in the right column, I get my balance.)

Transactions Page - Row expanded

albertandrejev:
Closes #871.

### DIFF
--- a/packages/bierzo-wallet/src/logic/transactions/types/BwSendTransaction/ui/SendTxRow/Details.tsx
+++ b/packages/bierzo-wallet/src/logic/transactions/types/BwSendTransaction/ui/SendTxRow/Details.tsx
@@ -36,7 +36,7 @@ const TxDetails = ({ userAddresses, tx }: Props): JSX.Element => {
             {formatTime(tx.time)}
           </Typography>
         </Block>
-        {userAddresses.includes(tx.original.sender) && (
+        {tx.outgoing && (
           <Block width="20%" paddingRight={3}>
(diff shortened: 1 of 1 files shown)


packages/bierzo-wallet/src/logic/transactions/types/BwSendTransaction/ui/SendTxRow/Details.tsx
```diff
@@ -36,14 +36,16 @@ const TxDetails = ({ userAddresses, tx }: Props): JSX.Element => {
...
         </Block>
-        <Block width="20%" paddingRight={3}>
-          <Typography variant="subtitle2" weight="regular" align="right" gutterBottom>
-            Transaction fee:
-          </Typography>
-          <Typography variant="subtitle2" weight="regular" color="textSecondary" align="right">
-            {txFee}
-          </Typography>
-        </Block>
+        {userAddresses.includes(tx.original.sender) && (
```
webmaster128: You can use `tx.outgoing`, which is pre-calculated for you at some other place

PR #915, opened 2019-12-09
Feature #910 - Upgrade Typescript to 3.7

### LINKED ISSUE #910: Upgrade to TypeScript 3.7
IOV Core 2.0 generates TypeScript 3.7 type definitions and some parts of those can only be understood by TypeScript 3.7+, so we should prepare ponferrada for that.

- Update here https://github.com/iov-one/ponferrada/blob/v1.1.1-alpha.2/package.json#L78
- Run `yarn install`

abefernan:
Closes #910.


packages/bierzo-wallet/src/config/index.ts
```diff
@@ -64,7 +64,7 @@ interface WindowWithConfig extends Window {
 
 const loadConfigurationFile = async (): Promise<Config> => {
   if (process.env.NODE_ENV === "test") {
-    return (window as WindowWithConfig).developmentConfig;
+    return ((window as Window) as WindowWithConfig).developmentConfig;
```
albertandrejev: What if we will wrap this into something like `getWindowWithConfig(window)` helper where we will implement all of this type castings?

packages/sanes-browser-extension/src/index.tsx
```diff
@@ -34,20 +34,18 @@ const render = (
   );
 };
 
-type PersonaContexType = [GetPersonaResponse, boolean];
+// REVIEW getHasStoredPersona() has a Promise<boolean> signature but
+// hasStoredPersona gets type boolean | null
+Promise.all([getPersonaData(), getHasStoredPersona()]).then(([persona, hasStoredPersona]) => {
```
webmaster128: It all gets much simpler by avoiding `Promise.all`. There is no need to run those in parallel:

```ts
(async () => {
  const persona = await getPersonaData();
  const hasStoredPersona = await getHasStoredPersona();

  // use unchanged old code
})();
```

packages/sanes-browser-extension/src/index.tsx
```diff
@@ -34,20 +34,18 @@ const render = (
   );
 };
 
-type PersonaContexType = [GetPersonaResponse, boolean];
+// REVIEW getHasStoredPersona() has a Promise<boolean> signature but
+// hasStoredPersona gets type boolean | null
```
albertandrejev: Odd, I got pure boolean type.

packages/sanes-browser-extension/src/index.tsx
albertandrejev: Why not to run it in parallel?

webmaster128: Aha! There is a much simpler fix: Replace 

```
Promise.all([getPersonaData(), getHasStoredPersona()]).then(
```

with

```
Promise.all([getPersonaData(), getHasStoredPersona()] as const).then(
```

The `as const` ensure the array literal is interpreted as a 2-component array. For fixed length arrays, the types can be different in each component.

packages/sanes-browser-extension/src/index.tsx
webmaster128: For variable length arrays, `Promise.all` expects the same type for each component.

packages/bierzo-wallet/src/config/index.ts
webmaster128: We can also avoid the argument and use the global object `window` from inside `function getWindowWithConfig(): WindowWithConfig`

packages/sanes-browser-extension/src/index.tsx
albertandrejev: Really? Hmm, I got proper typing, maybe this is VC bug I am not sure.

webmaster128: Did you restart VS Code after installing the latest TS version? Otherwise you still run the old TS 3.5 type server with the VS Code.

abefernan: VSCode won't highlight this issue unless you install the [nightly extension](https://marketplace.visualstudio.com/items?itemName=ms-vscode.vscode-typescript-next).

webmaster128: You can choose if you want to use the TS version shipped with VS Code or the one installed locally in your project. See

Bildschirmfoto 2019-12-09 um 14 27 37

albertandrejev: Yes, you are right. Your solution with `as const` solves this issue.

packages/bierzo-wallet/src/components/Header/index.stories.tsx
```diff
@@ -57,8 +57,11 @@ const txs: readonly (ProcessedSendTransaction | ProcessedTx)[] = [
 ];
 
 const txStore: DeepPartial<RootState> = {
+  // REVIEW Type '{ transactions: readonly (ProcessedSendTransaction | ProcessedTx<LightTransaction>)[]; }'
+  // is not assignable to type 'DeepPartial<NotificationState>'.
+  // Type 'TransactionId' is not assignable to type 'undefined'
   notifications: {
-    transactions: txs,
+    transactions: txs as any,
```
abefernan: Should I leave this as is? @webmaster128 @albertandrejev

webmaster128: I don't mind. Next time we work with this code, we can look into.

packages/bierzo-wallet/src/config/index.ts
```diff
@@ -62,9 +62,13 @@ interface WindowWithConfig extends Window {
   readonly developmentConfig: Config;
 }
 
+function getWindowWithConfig(): WindowWithConfig {
```
albertandrejev: Please move it into `ui-logic` package and use it everywhere.

packages/sanes-browser-extension/src/extension/backgroundscript.ts
```diff
@@ -3,7 +3,7 @@ import Backgroundscript, { IovWindowExtension } from "./background/model/backgro
 import { updateExtensionBadge } from "./background/updaters/extensionBadgeUpdater";
 
 const backgroundScript = new Backgroundscript();
-backgroundScript.registerActionsInBackground(window as IovWindowExtension);
+backgroundScript.registerActionsInBackground((window as Window) as IovWindowExtension);
```
albertandrejev: for instance here.

packages/sanes-browser-extension/src/routes/unlock/index.e2e.spec.ts
```diff
@@ -50,7 +50,7 @@ withChainsDescribe("DOM > Unlock route", (): void => {
     await findRenderedE2EComponentWithId(page, WALLET_STATUS_ROUTE);
 
     await bgPage.evaluate((): void => {
-      (window as IovWindowExtension).clearPersona();
+      ((window as Window) as IovWindowExtension).clearPersona();
```
albertandrejev: and here

packages/sil-governance/src/config/index.ts
```diff
@@ -25,9 +25,13 @@ interface WindowWithConfig extends Window {
   readonly developmentConfig: Config;
 }
 
+function getWindowWithConfig(): WindowWithConfig {
```
albertandrejev: if you will implement this function in `ui-logic` package you will not need this implementation.

packages/sanes-browser-extension/src/extension/backgroundscript.ts
abefernan: This does not have Config

Issue #895, opened 2019-12-02
Wallet - add registration functionality


annatorok:
Changes:

- [x] "Your starnames" should be renamed to "Your iovnames"
- [x] In case of new starname registration, user can add blockchain addresses that he wants to connect to that starname. See design.
- [x] In case of new starname registration, user can also add or remove these addresses and then register the starname by clicking on "Register for 5 IOV".
- [x] "Back" button should be renamed to "Cancel"

Stages:
- [x] Component and titles renaming (PR #920)
- [x] Stacked form fields components (PR #924)
- [x] iovname registration (PR #925)
- [x] iovname update (PR #940)
- [x] DOM and E2E tests (PR #954)

<img width="1440" alt="Neuma Wallet - Your Starnames" src="https://user-images.githubusercontent.com/7419692/69974195-0b8d3c00-1525-11ea-9abe-b48f2bd1e500.png">

<img width="1440" alt="Neuma Wallet - Register Starname" src="https://user-images.githubusercontent.com/7419692/69974199-0d56ff80-1525-11ea-913d-3615d5cb56d3.png">

PR #925, opened 2019-12-12
Feature #895 - Register new iovname

albertandrejev:
Part of the #895.

screenshot-localhost_9009-2019 12 13-13_05_12

### DIFF
--- a/packages/bierzo-wallet/src/routes/registerName/index.tsx
+++ b/packages/bierzo-wallet/src/routes/registerName/index.tsx
@@ -109,6 +109,10 @@ const RegisterUsername = (): JSX.Element => {
   const rpcEndpoint = ReactRedux.useSelector((state: RootState) => state.rpcEndpoint);
   const identities = ReactRedux.useSelector((state: RootState) => state.identities);
   const addresses = getChainAddressPairWithNames(identities);
+  const addressesSorted = Array.from(addresses).sort(
+    (a: ChainAddressPairWithName, b: ChainAddressPairWithName) =>
+      a.chainName.localeCompare(b.chainName, undefined, { sensitivity: "base" }),
+  );
 
   const bnsIdentity = getBnsIdentity(identities);
 
@@ -185,7 +189,7 @@ const RegisterUsername = (): JSX.Element => {
           onSubmit={onSubmit}
           validate={validate}
           onCancel={onReturnToBalance}
-          chainAddresses={addresses}
+          chainAddresses={addressesSorted}
           transactionFee={transactionFee}
         />
       )}
--- a/packages/bierzo-wallet/src/routes/registerName/index.tsx
+++ b/packages/bierzo-wallet/src/routes/registerName/index.tsx
@@ -1,4 +1,4 @@
-import { ChainId, Fee, Identity, TransactionId } from "@iov/bcp";
+import { Address, ChainId, Fee, Identity, TransactionId } from "@iov/bcp";
 import { BnsConnection } from "@iov/bns";
 import {
   BillboardContext,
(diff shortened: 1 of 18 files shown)


packages/bierzo-wallet/src/routes/registerName/index.tsx
```diff
@@ -109,6 +156,10 @@ const RegisterUsername = (): JSX.Element => {
   const rpcEndpoint = ReactRedux.useSelector((state: RootState) => state.rpcEndpoint);
   const identities = ReactRedux.useSelector((state: RootState) => state.identities);
   const addresses = getChainAddressPairWithNames(identities);
+  const addressesSorted = Array.from(addresses).sort(
+    (a: ChainAddressPairWithName, b: ChainAddressPairWithName) =>
+      a.chainName.localeCompare(b.chainName, undefined, { sensitivity: "base" }),
```
webmaster128: Can you deduplicate this comparator function?

packages/bierzo-wallet/src/routes/registerName/index.tsx
```diff
@@ -99,6 +100,52 @@ const validate = async (values: object): Promise<object> => {
...
+      pair = {};
+    }
+
+    const type = key.substr(fieldValueIdxLength + 1);
+    switch (type) {
+      case addressValueField:
+        pair = { ...pair, address: values[key] as Address };
+        break;
+      case blockchainValueField:
+        // eslint-disable-next-line no-case-declarations
```
webmaster128: Instead of this exception, please wrap code in `{ .... }` as in the example from https://eslint.org/docs/rules/no-case-declarations. Usually this was auto-fixed ... Did it not work?

```
switch (foo) {
    // The following case clauses are wrapped into blocks using brackets
    case 1: {
        let x = 1;
        break;
    }
    case 2: {
        const y = 2;
        break;
    }
    case 3: {
        function f() {}
        break;
    }
    case 4:
        // Declarations using var without brackets are valid due to function-scope hoisting
        var z = 4;
        break;
    default: {
        class C {}
    }
}
```

albertandrejev: No, it is not

packages/bierzo-wallet/src/routes/registerName/index.tsx
```diff
@@ -117,16 +117,17 @@ function getChainAddressPairsFromValues(
 
     const type = key.substr(fieldValueIdxLength + 1);
     switch (type) {
-      case addressValueField:
+      case addressValueField: {
         pair = { ...pair, address: values[key] as Address };
         break;
-      case blockchainValueField:
-        // eslint-disable-next-line no-case-declarations
+      }
```
webmaster128: The brackets are only required when you declare variables in the block (like `const chain = ...` below).

In `case addressValueField:` the brackets are not needed but it is also not wrong. Maybe linter complains, let's see ...

albertandrejev:
@webmaster128 Please review once again, I have made some refactorings and logic changes.

packages/bierzo-wallet/src/routes/registerName/index.tsx
```diff
@@ -186,6 +140,70 @@ const RegisterUsername = (): JSX.Element => {
...
+    const connection = await getConnectionForBns();
+    const usernames = await connection.getUsernames({ username });
+    if (usernames.length > 0) {
+      errors[REGISTER_USERNAME_FIELD] = "Personalized address already exists";
+      return errors;
+    }
+
+    const addressesToRegister = getChainAddressPairsFromValues(formValues, addressesSorted);
+    for (let index = 0; index < addressesToRegister.length; index++) {
+      const address = addressesToRegister[index];
```
webmaster128: `for (const address of addressesToRegister) {`

packages/bierzo-wallet/src/routes/registerName/index.tsx
```diff
@@ -186,6 +140,70 @@ const RegisterUsername = (): JSX.Element => {
...
+      const codec = await getCodecForChainId(address.chainId);
+      if (!codec.isValidAddress(address.address)) {
+        const addressFieldId = Object.entries(formValues).find(([_id, value]) => {
+          if (value === address.address) return true;
+          return false;
+        });
+        if (addressFieldId) {
+          errors[addressFieldId[0]] = "Not valid blockchain address";
```
webmaster128: if `addressField` contains the field ID, shouldn't it be `errors[addressFieldId]`?

albertandrejev: No. `addressFieldId` contains array of `key` and `value` (see `Object.entries(formValues)`), and I need `key` because `key` === input field `name`

albertandrejev: I will change naming.

webmaster128: Thanks for clarification

PR #926, opened 2019-12-12
Feature #911 - Adds starname.network domain

### LINKED ISSUE #911: Whitelist starname.network
*.starname.network should be whitelisted in the extension to prepare a connection between a to-be-built web application on that domain and the browser extension

abefernan:
Closes #911.

### DIFF
--- a/packages/sanes-browser-extension/config/manifest-production.json
+++ b/packages/sanes-browser-extension/config/manifest-production.json
@@ -20,7 +20,12 @@
   },
   "content_scripts": [],
   "externally_connectable": {
-    "matches": ["http://localhost/*", "https://*.iov.one/*", "https://*.neuma.io/*"]
+    "matches": [
+      "http://localhost/*",
+      "https://*.iov.one/*",
+      "https://*.neuma.io/*",
+      "https://*.starname.network/*"
+    ]
   },
   "background": {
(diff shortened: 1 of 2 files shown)


packages/sanes-browser-extension/config/manifest-production.json
```diff
@@ -20,7 +20,12 @@
   },
   "content_scripts": [],
   "externally_connectable": {
-    "matches": ["http://localhost/*", "https://*.iov.one/*", "https://*.neuma.io/*"]
+    "matches": [
+      "http://localhost/*",
+      "https://*.iov.one/*",
+      "https://*.neuma.io/*",
+      "https://*.starname.network/*"
```
webmaster128: Can you verify if this also matches https://starname.network (i.e. no subsomain)? This is probably documented in the Google extension docs.

We don't need non-subdomains for the other domains but we might want to use it here. I'm surprised we redirect to www.starname.network at the moment. We better prepare for using pure starname.network as well.

abefernan: It should work?
image

webmaster128: Looks good, thanks for checking. If it does not work it is a Google bug ;)

PR #929, opened 2019-12-12
Fix #703 - Adds check for lastError

### LINKED ISSUE #703: Check runtime.lastError when connecting to extension
Right now, when the extension is not installed, we get this warning by the browser:

Bildschirmfoto 2019-09-11 um 11 34 51

So `runtime.lastError` should be checked in

```
chrome.runtime.sendMessage(config.extensionId, request, response => {
   // ...
```

according to the `sendMessage` docs:

> @param responseCallback Optional
> Parameter response: The JSON response object sent by the handler of the message. If an error occurs while connecting to the extension, the callback will be called with no arguments and runtime.lastError will be set to the error message.

This is not critical, since we handle `respnse === undefined`.

abefernan:
Closes #703.

### DIFF
--- a/packages/bierzo-wallet/src/communication/extensionRpcEndpoint.ts
+++ b/packages/bierzo-wallet/src/communication/extensionRpcEndpoint.ts
@@ -48,7 +48,7 @@ export const extensionRpcEndpoint: RpcEndpoint = {
 
     return new Promise(resolve => {
       chrome.runtime.sendMessage(config.extensionId, request, response => {
-        if (chrome.runtime.lastError || response === undefined) {
+        if (chrome.runtime.lastError) {
           resolve(undefined);
           return;
         }
@@ -70,7 +70,7 @@ export const extensionRpcEndpoint: RpcEndpoint = {
 
     return new Promise((resolve, reject) => {
       chrome.runtime.sendMessage(config.extensionId, request, response => {
-        if (chrome.runtime.lastError || response === undefined) {
+        if (chrome.runtime.lastError) {
           resolve(undefined);
           return;
         }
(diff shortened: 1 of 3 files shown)


packages/bierzo-wallet/src/communication/extensionRpcEndpoint.ts
```diff
@@ -70,6 +75,11 @@ export const extensionRpcEndpoint: RpcEndpoint = {
 
     return new Promise((resolve, reject) => {
       chrome.runtime.sendMessage(config.extensionId, request, response => {
+        if (chrome.runtime.lastError) {
+          resolve(undefined);
+          return;
+        }
+
         if (response === undefined) {
```
albertandrejev: Please put both blocks under common condition. `if (chrome.runtime.lastError || response === undefined) { ...`

abefernan: I tested that, I put it separately because legibility because:

- Removes the error:
`if (chrome.runtime.lastError || response === undefined) { ...`
- Does not remove the error:
`if (response === undefined || chrome.runtime.lastError) { ...`

But if it's more idiomatic that way all change it no problem.

webmaster128: Looking at the [documentation of chrome.runtime.sendMessage](https://developer.chrome.com/apps/runtime#method-sendMessage):

> The JSON response object sent by the handler of the message. If an error occurs while connecting to the extension, the callback will be called with no arguments and runtime.lastError will be set to the error message. 

you see that `response` is either a JSON object (i.e. not `undefined`) or an error occurred. This means we can as well remove the `response === undefined`, since it happens exactly when `chrome.runtime.lastError` is set.

PR #932, opened 2019-12-16
Upgrade firebase-tools to ^7.7.0

webmaster128:
Because of

Bildschirmfoto 2019-12-16 um 09 42 59


webmaster128:
It would be good to make a staging deployment once this is merged.

PR #941, opened 2019-12-20
Feature #922 - Improve amount column

### LINKED ISSUE #922: Inconsistent Amount column
For name registrations, the Amount column shows the transaction fee

Bildschirmfoto 2019-12-11 um 15 37 29

For send transactions however, only the amount of transferred tokens are shown (without the fee). In this example I send 10 IOV + 0.5 IOV fee.

Bildschirmfoto 2019-12-11 um 15 37 08

If the fee is supposed to be part of the "Amount" value, this should be consistent for all transaction types.

abefernan:
Closes #922.

### DIFF
--- a/packages/bierzo-wallet/src/logic/transactions/types/BwSendTransaction/ui/SendTxRow/index.tsx
+++ b/packages/bierzo-wallet/src/logic/transactions/types/BwSendTransaction/ui/SendTxRow/index.tsx
@@ -3,7 +3,7 @@ import { makeStyles, Theme } from "@material-ui/core";
 import { useTheme } from "@material-ui/styles";
 import { Block, CircleImage, Hairline, Image, Typography, useOpen } from "medulas-react-components";
 import * as React from "react";
-import { amountToString, ellipsifyAddress } from "ui-logic";
+import { amountToString, ellipsifyAddress, sumToString } from "ui-logic";
 
 import { getAddressPrefix } from "../../../../../../routes/transactions/components/TxTable/rowTxBuilder";
 import { ProcessedSendTransaction } from "../../../../../../store/notifications";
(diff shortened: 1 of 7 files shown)


packages/bierzo-wallet/src/logic/transactions/types/BwSendTransaction/ui/SendTxRow/index.tsx
```diff
@@ -34,11 +34,16 @@ interface Props {
...
+  const amountClass = sendTx.outgoing ? undefined : classes.amountFrom;
+
   const [isOpen, toggle] = useOpen();
 
   const address = sendTx.outgoing ? sendTx.original.recipient : sendTx.original.sender;
   const amountSign = sendTx.outgoing ? "-" : "+";
-  const amountClass = sendTx.outgoing ? undefined : classes.amountFrom;
+
+  const amount = sendTx.original.amount;
+  const fee = sendTx.outgoing && sendTx.original.fee && sendTx.original.fee.tokens;
```
webmaster128: You can even go fancies with [TypeScript 3.7 optional chaining:](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining)

```ts
const fee = sendTx.outgoing && sendTx.original.fee?.tokens;
```

abefernan: I had tried that but go this issue:
image

webmaster128: 1. Check your IDE is using TS 3.7 
Bildschirmfoto 2019-12-20 um 11 04 50
 
2. This is a prettier error. Prettier got updated to support TS 3.7 in https://github.com/iov-one/ponferrada/pull/937. Did you pull current master and base on that? Did you `yarn install`? Did you reload VS Code to restart the eslint plugin?

packages/ui-logic/src/balances.ts
```diff
@@ -40,3 +40,12 @@ export function amountToGwei(amount: Amount): string {
   const display = Number.isInteger(value) ? value : value.toFixed(2);
   return `${display} Gwei`;
 }
+
+export function sumToString(amountA: Amount, amountB: Amount): string {
```
albertandrejev: I think we should have separate function to sum two `Amounts` and then if we want, convert it to the string we should use `amountToString` function. Please update this function signature to `function sumToString(amountA: Amount, amountB: Amount): Amount` and use it to sum two `Amounts`. Then if you want to convert it to string simply use `amountToString`.

albertandrejev: Also please change function name, something like `sumAmounts`.

PR #954, opened 2020-01-14
Feature #895 - Add E2E tests

### LINKED ISSUE #895: Wallet - add registration functionality

albertandrejev:
Closes #895.

### DIFF
--- a/packages/bierzo-wallet/src/routes/registerName/components/index.tsx
+++ b/packages/bierzo-wallet/src/routes/registerName/components/index.tsx
@@ -137,7 +137,7 @@ const Layout = ({
   }, [chainAddresses, iovnameAddresses]);
 
   const initialValues = React.useMemo(() => getFormInitValues(chainAddressesItems), [chainAddressesItems]);
-  const { form, handleSubmit, invalid, pristine, submitting, validating } = useForm({
+  const { form, handleSubmit, invalid, submitting, validating } = useForm({
     onSubmit,
     validate,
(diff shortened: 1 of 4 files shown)


packages/bierzo-wallet/src/routes/registerName/components/index.tsx
```diff
@@ -156,7 +156,7 @@ const Layout = ({
         <Button
           fullWidth
           type="submit"
-          disabled={invalid || pristine || submitting || validating}
```
abefernan: Why?

albertandrejev: Unfortunately if I remove address line then internal state of the form component  not changing. So it is not possible to update the name because button is disabled.

PR #970, opened 2020-01-17
Feature #947 - Edit ability to edit IOV RPC node

### LINKED ISSUE #947: Ability to edit RPC url to Neuma extension for IOV testnet
**Goal:**
In the future, we want our dev team to easily test their new application integrations with Neuma and for that, it would be great to add a functionality for the Neuma extension, so a new custom RPC url can be added manually.

- [x] it is possible to add custom RPC url in staging extension for IOV testnet (basically changing the RPC url for the existing IOV testnet)
- [x] editing the RPC url will change the RPC url for IOV testnet
- [x] Two fields are needed for editing a new RPC url: "Network Name" which is prepopulated from the already given one (something like IOV Dancenet, or similar) and prepopulated "New RPC Url"
- [x] "Cancel" and "Save" buttons are shown after the two fields - "Save" will save all the changes, Cancel will go back to the previous screen to the Networks
- [x] below the IOV testnet RPC url, there is a "Change" text which will navigate the user to the two fields previously mentioned in a new screen

Menu - Networks - Change Network - Empty field
Menu - Networks - Change Network
Menu - Networks (1)

If you check Metamask, there is already a feature like this.

Screenshot 2020-01-07 at 12 15 37
Screenshot 2020-01-07 at 12 15 43

albertandrejev:
Part of the #947.


albertandrejev:
No, this doesn't refers to Chains it refers to Accounts only (doesn't supported right now). Correct me if I am wrong.

abefernan:
> No, this doesn't refers to Chains it refers to Accounts only (doesn't supported right now). Correct me if I am wrong.

Sorry, I meant the following rendering logic:

```typescript
const showIovAddress = persona.connectedChains.length > 0;
```

IOV address should only show when one of the connected chains is IOV/BNS, since that's where the iov address comes from.

I used the greater than 0 because iov networks where supposed to be the first network in config file.

albertandrejev:
Sorry still do not understand where this assumption is come from :( I mean that bns should be first item in the chains array.

abefernan:
> Sorry still do not understand where this assumption is come from :( I mean that bns should be first item in the chains array.

Ok let me try to be more clear:

- It was requested by Anna and by the design that the Networks view would show the IOV network before every other. So it would show "IOV Devnet", "Ganache", "Lisk Devnet", etc, with "IOV Devnet" being the first (I don't know if this requirement is necessary anymore).
- Because I didn't have a way to check that a Network was BNS (I didn't want to rely on the name of the network to put it first on the list), I assumed that the IOV network would be the first in the list of chains of the config file.
- The rendering logic of `showIovAddress` should be changed to only be true when a BNS network is inside the array `persona.connectedChains`. That is, there is a BNS network where we can get the IOV address.

packages/sanes-browser-extension/src/extension/background/model/persona/config/configurationfile.ts
```diff
@@ -80,3 +82,27 @@ export async function getChainNode(chainId: ChainId): Promise<string> {
...
+    const updatedBns = {
+      chainSpec: {
+        ...bnsChain.chainSpec,
+        name: bnsRpcConfig.networkName,
+        node: bnsRpcConfig.rpcUrl,
+      },
+    };
+
+    const withoutBns = chains.filter(chain => chain.chainSpec.codecType !== "bns");
+    return [updatedBns, ...withoutBns];
```
albertandrejev: Now bns is the first record in chain array.

PR #1027, opened 2020-02-17
Upgrade @typescript-eslint/* to 2.20.0

webmaster128:
Upgrading @typescript-eslint/* to 2.20.0 should ensure the Parsing errors should disappear, also for new and renamed files.

Manually installing eslint makes sure that version 6 is the binary installed to `node_modules/.bin/`, such that `yarn eslint --version` returns version 6. Unfortunately, `web-ext` still installs eslint 5 as well, which was used as the `eslint` binary before.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "@types/storybook__addon-storyshots": "^5.1.1",
     "@types/storybook__addon-viewport": "^4.1.1",
     "@types/storybook__react": "^4.0.2",
+    "eslint": "^6",
     "eslint-config-prettier": "^6.2.0",
     "eslint-plugin-prettier": "^3.0.1",
     "eslint-plugin-simple-import-sort": "^4.0.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -9378,7 +9378,7 @@ eslint@^3.7.1:
     text-table "~0.2.0"
     user-home "^2.0.0"
 
-eslint@^6.6.0:
+eslint@^6.6.0, eslint@^6:
   version "6.8.0"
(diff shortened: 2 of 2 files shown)


webmaster128:
@abefernan @albertandrejev 

Could you check if that solves the problem for you?

```sh
git checkout master
git pull
git clean -xdf
yarn install
yarn eslint --version # shows v6.8.0
```

Also check: eslint plugin `dbaeumer.vscode-eslint` is >= 2.0.15 and TypeScript 3.7 is used (either from VSCode or from local workspace; I alwasy use the version from the workspace).

Now close and re-open the VSCode window (or use "Reload developer windows"). 

Now when you click on ESLint in the bottom right you see something nice like

```
[Info  - 10:06:28 AM] ESLint server stopped.
[Info  - 10:06:30 AM] ESLint server running in node v12.4.0
[Info  - 10:06:30 AM] ESLint server is running.
[Info  - 10:06:34 AM] ESLint library loaded from: /projects/ponferrada/node_modules/eslint/lib/api.js
```

for OUTPUT and nothing in PROBLEMS.

albertandrejev:
> eslint plugin dbaeumer.vscode-eslint is >= 2.0.15

:+1: 
>TypeScript 3.7 is used (either from VSCode or from local workspace;

:+1: 
> for OUTPUT and nothing in PROBLEMS.

:+1:

albertandrejev:
Issues gone, thanks

webmaster128:
Ah cool, glad to hear that! Happy linting.

ps.: if linting is too slow, I suggest removing the pre-push hook. CI will lint everything for you and during development you only lint the files you work with.

PR #1062, opened 2020-03-17
Feature #1011 - Transfer account

### LINKED ISSUE #1011: Starname in Wallet: Transfer iovname/starnames/names

albertandrejev:
Closes #1011.

screenshot-localhost_9009-2020 03 17-16_37_17


packages/bierzo-wallet/src/logic/transactions/types/BwParserFactory.ts
```diff
@@ -234,6 +234,14 @@ export class BwParserFactory {
...
     } else if (isUpdateTargetsOfUsernameTx(payload)) {
       return new BwUpdateUsernameTargetParser();
+    } else if (isRegisterAccountTx(payload)) {
+      return new BwUpdateUsernameTargetParser();
+    } else if (isReplaceAccountTargetsTx(payload)) {
+      return new BwUpdateUsernameTargetParser();
+    } else if (isTransferAccountTx(payload)) {
+      return new BwUpdateUsernameTargetParser();
+    } else if (isTransferDomainTx(payload)) {
+      return new BwUpdateUsernameTargetParser();
```
abefernan: Why this changes with `BwUpdateUsernameTargetParser`?

albertandrejev: Because otherwise it will return `BwUnkownParser` which is not right. It works but I am not sure that this is right behaviour.

abefernan: The parsers are on the `types` directory, each on its corresponding `Bw<Whatever>Tx`

albertandrejev: got it

Issue #1056, opened 2020-03-09
Implement deployment of wallet without releasing chrome extension


annatorok:
Goal:
We want to release changes to Wallet separately from the Chrome extension. 

- [x] Implement changes in travis script to allow for separate deployment
- [x] Please document in Wiki and Confluence how we can do that in [the Releasing page](https://github.com/iov-one/ponferrada/wiki/Releasing).

PR #1063, opened 2020-03-17
Feature #1056 - Adds granularity for deployment in travis

abefernan:
Part of #1056.

This PR adds some logic to the Travis script that enables the deployment of ponferrada's apps in an independent fashion.

The new logic leverages git tags that make use of the following format:

`v<version_number>[-apps_to_deploy][-alpha]` Where `<>` is mandatory and where `[]` is optional.

The `apps_to_deploy` part can be any combination of the following letters: 

- e: deploy extension
- w: deploy wallet
- g: deploy governance

When no letter is present, every app is deployed except the extension.

The `alpha` part indicates that it should not be pushed to production if present.

Tag examples:
- `v1.0.0-w`: deploys the wallet app to staging and production
- `v1.0.0-ge-alpha`: deploys the governance app and the extension to staging
- `v1.0.0-alpha`: deploys the wallet and governance apps to staging

Our most common use case and the most needed as of now will be this:
- Push tag `v1.0.0-e-alpha`.
- Wait for Google's approval of the extension in staging.
- Push tag `v1.0.0-alpha` so that the versions of the wallet and governance match the newly available staging extension.
- Do all the necessary testing and bugfixing on staging.
- Push tag `v1.0.0-e`.
- Wait for Google's approval of the extension in production.
- Push tag `v1.0.0` so that the versions of the wallet and governance match the newly available production extension.

### DIFF
--- a/scripts/travis.sh
+++ b/scripts/travis.sh
@@ -134,17 +134,6 @@ fold_end
 # Deployment
 #
 
-# Detects if a version tag is a prerelease according to the semver definition (https://semver.org/#spec-item-9)
-# Example inputs: v1.0.0-alpha.1, v1.0.0
-function is_prerelease() {
-  # Bash uses the Extended Regular Expression (ERE) dialect
-  if [[ $1 =~ '-' ]]; then
-    return 0
-  else
-    return 1
-  fi
-}
-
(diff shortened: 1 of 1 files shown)


albertandrejev:
We should add this description to the wiki as well. https://github.com/iov-one/ponferrada/wiki/Releasing

PR #1107, opened 2020-04-13
Milestone - Iovname migration to account module

abefernan:
This PR aims to enable the migration of iovname functionality from BNS usernames to the account module.

https://github.com/iov-one/ponferrada/milestone/12

iov-core needs to support the latest bnsd version before moving on, which as of now will be achieved with this PR: https://github.com/iov-one/iov-core/pull/1415.

With iov-core not updated, an error occurs when registering an iovname, because the `has_superuser` field is not correctly decoded.

This PR currently points to master, in the future it may be need to be changed.


albertandrejev:
@davepuchyr looks like domain object for `iov` domain doesn't contain `hasSuper` user field. But according to the docs (https://htmlpreview.github.io/?https://github.com/iov-one/weave/blob/master/docs/proto/index.html#account.Domain) this field is not optional.

scripts/bnsd/genesis_app_state.json
```diff
@@ -17,6 +17,32 @@
...
+        "name": "test",
+        "owner": "bech32:tiov15nuhg3l8ma2mdmcdvgy7hme20v3xy5mkxcezea",
+        "valid_until": 1895827578,
+        "targets": [
+          {
+            "blockchain_id": "localiovdevnet",
+            "address": "bech32:tiov1l678408y7a64cj66s8j64fevmspyfxdmv38bcd"
+          }
+        ]
+      }
```
albertandrejev, line 43: Got it, you forgot to add `has_superuser` flag to the domain configuration.

albertandrejev:
@davepuchyr I think I found where this issue is come from.

scripts/bnsd/genesis_app_state.json
```diff
@@ -17,6 +17,32 @@
       "coins": ["4000 BASH"]
     }
   ],
+  "account": {
+    "domains": [
+      {
+        "admin": "seq:test/alice/1",
+        "domain": "iov",
+        "valid_until": 1895827578
+      }
```
albertandrejev, line 26: Got it, you forgot to add has_superuser flag to the domain configuration.

abefernan: Ok and should it be true or false? The genesis file from iov-core does not have that flag.

albertandrejev: Checking right now. Then genesis for IOV Core should be fixed too.

abefernan: I've checked out some Slack convos and I think I should set it to true.

abefernan: If I set it to false, the same problem persists. When set to true, it says only admin can register.

albertandrejev: @davepuchyr looks like a bug in bnsd.

davepuchyr: The *iov domain is special because it's the only domain with `has_superuser = false`.  Without a superuser, certain operations are restricted, like the domain admin **cannot** delete accounts within the domain, only account owners can delete their account.  

The economic design of the starnames is that casual users can inexpensively use the *iov domain whereas hardcore users/businesses create domains with `has_superuser == true`.

Long story short, it's not a bug in bnsd.

abefernan: *iov being the only domain with `has_superuser = false` was my impression and your functionality descrition makes sense, but we are talking about **registering** an iovname. I believe that operation should be able to be done by any user.

abefernan: We may have to check iov-core then, since it's an error that happens when decoding.

davepuchyr: Yes, you are right.  We need to fix https://github.com/iov-one/iov-core/blob/master/packages/iov-bns/src/decodeobjects.ts.  https://github.com/iov-one/iov-core/pull/1415 is a PR that I'm anxious to merge and can include the fix.

PR #1116, opened 2020-04-22
Mock a connection to a network

davepuchyr:


packages/sanes-browser-extension/src/routes/wallet/index.dom.spec.ts
```diff
@@ -111,7 +111,7 @@ withChainsDescribe("DOM > Feature > Wallet Status Drawer", () => {
 
     const liElements = scryRenderedDOMComponentsWithTag(walletStatusDom, "li");
     const liskNetwork = liElements[4];
-    expect(liskNetwork.textContent).toBe("IOV Devnethttp://localhost:23456/Change");
+    expect(liskNetwork.textContent).toBe("Starname Networkhttp://localhost:23456/");
```
abefernan, line 114: We used to take for granted that IOV network would always be the first one on the list. This would need to be reverted in case we want that in the future.

PR #1128, opened 2020-05-12
dialog box

antoineherzog:


antoineherzog:
@davepuchyr Yes my idea was to make it super visible in a next PR. So we can push in staging then production so we can test the flow with real user then display it after a login. let me know what do you think ..

davepuchyr:
> @davepuchyr Yes my idea was to make it super visible in a next PR. So we can push in staging then production so we can test the flow with real user then display it after a login. let me know what do you think ..

Ah, ok, I didn't know you had another PR in mind.  Sounds good.  I still think that the star1 address should not be removable, though.

antoineherzog:
i don't know how to do that .. I propose to merge this one, I can do another one about not revmoing starname address in another PR is it i good for you ? @davepuchyr

davepuchyr:
> i don't know how to do that .. I propose to merge this one, I can do another one about not revmoing starname address in another PR is it i good for you ? @davepuchyr

OK

PR #1132, opened 2020-05-18
Disallow editing/removal of star1 address

davepuchyr:
Medulas' SelectField's disable doesn't work, so what should have been a 3 line change turned into a multi-line hack.


antoineherzog:
LGTM

PR #1133, opened 2020-05-18
Message redirection

antoineherzog:


packages/bierzo-wallet/src/routes/account/register/components/SelectAddressesTable.tsx
```diff
@@ -92,6 +92,8 @@ const AddressRow = ({
     }
   };
 
+  const isStarname = addressItem.chain.chainName.toLocaleLowerCase().indexOf("starname") !== -1;
```
davepuchyr, line 95: @antoineherzog, I know it's not your fault, but having two versions of SelectAddressesTable.tsx is awful, ie not maintainable.  The `src/components/AccountEdit/SelectAddressesTable.tsx` should be the single canonical version.

antoineherzog: I agree, however I would like to keep it like that for now for this PR. I spent lots of time of testing and some test are not passing. I would like to finish it

davepuchyr: OK, we can deal with the multiple `SelectAddressesTable.tsx`s in another PR.

PR #1136, opened 2020-07-16
Proof of concept

iharob:


packages/starname-wallet/src/routes/balance/index.tsx
```diff
@@ -12,16 +14,34 @@ function onRegisterIovname(): void {
   history.push(IOVNAME_REGISTER_ROUTE);
 }
 
-const Balance = (): React.ReactElement => {
-  const tokens = ReactRedux.useSelector((state: RootState) => state.balances);
+const knownTickers: { [name: string]: string } = {
+  uvoi: "IOV",
```
davepuchyr, line 18: Values like these and the REST/RPC URL should be taken from getConfig().

iharob: I was honestly hoping to query this from somewhere. But ok, I will move these to the config files.

